PR #66, opened 2025-11-25
add uv as package manager

drumtorben:


ronaldbradford:
uv is the new data scientist way  for python packaging (think pyenv + pip + venv). 
I had just looked into it recently, and there is a learning curve to getting start first time.  I have no actual experience with it yet!
It's written in Rust so apparently it's more efficient.

PR #92, opened 2026-01-07
fix: Add AzureTransportOptionType so that curl can be used when querying due to CA certificates error

schotime:
Note: Problem with the SSL CA cert (path? access rights?)) because when specifying the extension will try to find the bundle certificate in various paths (that is not done by curl by default and might be wrong due to static linking).

https://duckdb.org/docs/stable/core_extensions/azure

### DIFF
--- a/internal/database/duckdb.go
+++ b/internal/database/duckdb.go
@@ -3,6 +3,7 @@ package database
 import (
 	"database/sql"
 	"fmt"
+	"runtime"
 	"time"
 
 	_ "github.com/duckdb/duckdb-go/v2"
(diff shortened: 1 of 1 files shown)


xe-nvdk (changes requested):
## Code Review

Thanks for this PR! The fix addresses a documented DuckDB Azure extension issue where the default HTTP transport on Linux can fail to find CA certificate bundles in certain environments (minimal containers, non-standard distros, statically linked binaries).

### What looks good:
- Fix is well-scoped (Linux only via runtime.GOOS)
- Uses SET GLOBAL consistently with other settings in the file
- Error handling follows existing patterns
- Referenced DuckDB documentation supports this approach

### Requested change:

Please add a debug log message when setting the transport option. This helps with troubleshooting and maintains consistency with how other configurations are logged in this codebase.

Note: The configureAzureAccess function does not currently have access to the logger. You could either:
1. Pass the logger as a parameter to configureAzureAccess
2. Return a boolean indicating the setting was applied, and log it in the caller (New function in duckdb.go already logs azure_enabled)

Once this is addressed, happy to approve!

schotime:
Did you want debug or info?

schotime:
Ok thats been done and rebased.

"Please add a _debug_ log message when setting"

Debug above confused me in your message.

PR #132, opened 2026-01-18
fix: Ensure UTC dates for compaction filenames

schotime:
Dates in the filenames were not using UTC when compacting. I've also removed the unused method GetCompactedFilename as this happens in CompactFiles in job.go.

### DIFF
--- a/internal/compaction/daily.go
+++ b/internal/compaction/daily.go
@@ -130,12 +130,6 @@ func (t *DailyTier) ShouldCompact(files []string, partitionTime time.Time) bool 
 	)
 }
 
-// GetCompactedFilename generates the filename for a compacted file
-func (t *DailyTier) GetCompactedFilename(measurement string, partitionTime time.Time) string {
-	timestamp := partitionTime.Format("20060102")
-	return measurement + "_" + timestamp + "_daily.parquet"
-}
-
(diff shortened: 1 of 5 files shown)


xe-nvdk:
No worries! Go's date formatting is quirky - you have to use the specific reference date `Mon Jan 2 15:04:05 MST 2006` as the format template.

For the test fixes, here's exactly what to change in `internal/compaction/daily_test.go`:

**Line 56** - change the filename:
```go
"default/metrics/2026/01/01/metrics_20260102_143000_daily.parquet",
```

**Line 58** - update the expected time to match:
```go
expected: time.Date(2026, 1, 2, 14, 30, 0, 0, time.UTC),
```

**Line 65** - change the filename:
```go
"default/metrics/2026/01/01/metrics_20260102_150000_daily.parquet",
```

**Line 67** - update the expected time:
```go
expected: time.Date(2026, 1, 2, 15, 0, 0, 0, time.UTC),
```

You can verify locally with:
```bash
go test -v ./internal/compaction/... -run TestExtractNewestFileTime
```

Re: CI - yes, that would be great! We can set that up separately. For now, just run the tests locally before pushing.

xe-nvdk:
Tests pass - merging! Thanks for the fix @schotime 🎉

Issue #131, opened 2026-01-17
S3 credentials not passed to compaction subprocess


khalid244:
## Description

Compaction subprocess fails on S3-compatible storage (Hetzner, MinIO, etc.) because credentials aren't forwarded to the subprocess.

### Error Log
```json
{"level":"warn","component":"compaction-subprocess","component":"s3-storage","error":"operation error S3: HeadBucket, get identity: get credentials: failed to refresh cached credentials, no EC2 IMDS role found, operation error ec2imds: GetMetadata, request canceled, context deadline exceeded","message":"Could not verify bucket exists (may need to create it)"}

{"level":"error","component":"compaction-subprocess","error":"failed to download files: failed to read from S3: operation error S3: GetObject, exceeded maximum number of attempts, 3, get identity: get credentials: failed to refresh cached credentials, no EC2 IMDS role found, operation error ec2imds: GetMetadata, exceeded maximum number of attempts, 3, request send failed, Get \"http://169.254.169.254/latest/meta-data/iam/security-credentials/\": dial tcp 169.254.169.254:80: i/o timeout","message":"Compaction job failed"}
```

### Root Cause
`internal/compaction/manager.go:209-214` passes Azure credentials to subprocess but not S3:

```go
var extraEnv []string
if azureBackend, ok := m.StorageBackend.(*storage.AzureBlobBackend); ok {
    // Azure credentials passed ✓
}
// S3 credentials not passed ✗
```

xe-nvdk:
**Update:** Also fixed the `use_ssl` config issue you mentioned!

The second commit (`5aad114`) now passes `use_ssl` to the subprocess, so it will use the same protocol (HTTP/HTTPS) as the main process.

**Summary of fixes in `fix/s3-subprocess-credentials` branch:**
1. ✅ S3 credentials passed via `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` env vars
2. ✅ `use_ssl` config passed to subprocess

Both issues should now be resolved. Please test when you get a chance!

PR #133, opened 2026-01-19
Fix query failure when daily compacted files don't exist

khalid244:
## Summary
- Add retry logic for queries failing with "No files found" on day-level paths
- Zero overhead on happy path (retry only on failure)


khalid244:
After further debugging, I've identified that this issue is related to [#131](https://github.com/Basekick-Labs/arc/issues/131).                           
The compactor sends requests to S3 using HTTP instead of HTTPS, even when ARC_STORAGE_S3_USE_SSL=true is set.
This causes daily compaction jobs to fail.

PR #199, opened 2026-02-11
fix(wal): prevent data duplication from periodic WAL recovery

khalid244:
## Summary

- Replace periodic WAL recovery (replay) with flush+purge to eliminate systematic data duplication — every record written through the WAL was being stored at least twice because periodic recovery replayed entries already flushed to parquet
- Add `PurgeInactive()` method to delete rotated WAL files while preserving the active one during normal operation
- Fix empty WAL files (7-byte header-only) accumulating indefinitely because `len(entries) == 0` caused them to skip the cleanup condition

## How it works

The periodic WAL maintenance goroutine now:
1. Calls `arrowBuffer.FlushAll()` to persist all buffered data to parquet
2. If flush succeeds, calls `walWriter.PurgeInactive()` to delete non-active WAL files
3. If flush fails, skips purge entirely — WAL files are preserved for retry next cycle

Startup recovery is **unchanged** — after a crash the buffer is empty and WAL files are correctly replayed.

## Test plan

- [x] All existing WAL tests pass (`go test ./internal/wal/ -v`)
- [x] Full binary builds successfully (`go build ./cmd/arc/`)

### DIFF
--- a/cmd/arc/main.go
+++ b/cmd/arc/main.go
@@ -370,42 +370,31 @@ func main() {
 		go func() {
 			ticker := time.NewTicker(recoveryInterval)
 			defer ticker.Stop()
-			walLogger := logger.Get("wal-recovery")
+			walLogger := logger.Get("wal-cleanup")
 
 			for {
 				select {
 				case <-walRecoveryCtx.Done():
 					return
 				case <-ticker.C:
-					recovery := wal.NewRecovery(cfg.WAL.Directory, walLogger)
-					// Skip the active WAL file to avoid reading while it's being written
-					activeFile := ""
-					if walWriter != nil {
-						activeFile = walWriter.CurrentFile()
+					// Flush all buffers to ensure WAL data is persisted to parquet
+					if err := arrowBuffer.FlushAll(context.Background()); err != nil {
+						walLogger.Error().Err(err).Msg("Periodic flush failed, keeping WAL files")
+						continue
 					}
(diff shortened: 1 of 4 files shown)


xe-nvdk:
Hey Khalid, thank you for this report and PR, this was fixed, what version are you using, is this for S3 backend?

xe-nvdk:
Hey Khalid, thanks for the detailed report and PR — we appreciate you digging into the WAL internals like this.

You're right that there was a duplication issue with WAL recovery, and we fixed it in PR #173 (included in v26.02.1, the version you're running). That fix added a shutdown purge: after all buffers flush to parquet, WAL files are deleted before the process exits, so on the next startup there's nothing to replay. Crash recovery (unclean shutdown) is also safe — startup replays WAL entries once via `WriteColumnarDirectNoWAL`, deletes the files, and that's it.

Could you help us narrow down what you're seeing? Specifically:

- Are you observing duplicate data **after a restart** (clean or crash)?
- Or are you seeing it **during normal runtime** without any restarts?
- How are you detecting the duplication — duplicate rows in query results, inflated row counts, storage size larger than expected?

This will help us understand if there's a scenario #173 didn't cover. The periodic recovery goroutine during runtime could still replay rotated WAL files once before deleting them, but the impact is bounded (each record at most twice, not compounding) and compaction would merge them. If that's what you're hitting, we'd like to address it properly.

Also, for future findings — would you mind opening an issue first? It helps us triage, discuss the approach together, and coordinate before jumping into implementation. That way we can align early and get fixes shipped faster.

Thanks again for the contribution!

PR #222, opened 2026-02-21
fix(ingest): replace fixed ticker with self-adjusting timer for age-b…

### LINKED ISSUE #142: Buffer age-based flush takes up to 2x configured max_buffer_age_ms
## Description

Buffers configured with `max_buffer_age_ms=5000` are being flushed at ~7 seconds instead of ~5 seconds under high throughput scenarios.

## Root Cause

The periodic flush ticker fires every `max_buffer_age_ms` (e.g., 5000ms). When buffers are created between ticker fires (which happens frequently under high throughput after size-based flushes), they must wait until the next ticker fire to be checked for age.

**Timeline:**
```
T=0ms:     Ticker fires, checks all buffers
T=1ms:     Size-based flush completes, new buffer created
T=5000ms:  Ticker fires again
           Check: age = 4999ms < 5000ms → NOT flushed
T=10000ms: Next ticker fires  
           Check: age = 9999ms >= 5000ms → FLUSHED
```

In practice, this results in buffers waiting 1.3-1.4x the configured time (observed: 6.8-7 seconds for max_buffer_age_ms=5000).

## Evidence

Real production logs showing the issue:
```
2026-01-19T16:04:33-03:00 INF Flushing aged buffer age=6951.074 buffer_key=production/net
2026-01-19T16:04:34-03:00 INF Flushing aged buffer age=6829.292 buffer_key=production/cpu
2026-01-19T16:04:34-03:00 INF Flushing aged buffer age=6749.661 buffer_key=production/mem
2026-01-19T16:04:35-03:00 INF Flushing aged buffer age=6781.154 buffer_key=production/disk
```

Expected: `age ~5000-5100ms`
Actual: `age ~6700-7000ms`

## Proposed Fix

Change the ticker to fire more frequently (every `max_buffer_age_ms / 2`) while keeping the same age threshold check:

```go
// File: internal/ingest/arrow_writer.go:950
// Current:
flushTimer: time.NewTicker(time.Duration(cfg.MaxBufferAgeMS) * time.Millisecond),

// Fixed:
flushTimer: time.NewTicker(time.Duration(cfg.MaxBufferAgeMS/2) * time.Millisecond),
```

This ensures buffers are checked twice as often, reducing worst-case flush time from 2x to ~1.5x the configured value.

## Impact

- **High throughput workloads** - Most affected due to frequent size-based flushes creating new buffers
- **Low throughput workloads** - Minimal impact (buffers already flush by size before age)
- **Performance** - Negligible CPU overhead from more frequent ticker (still only checking age, not forcing flushes)

## Reproduction

Run the high-throughput benchmark with `max_buffer_age_ms=5000` and observe `age` values in logs exceeding 6000ms.

xe-nvdk:
…ased flush

The periodic flush goroutine previously used a fixed-period ticker (firing every max_buffer_age_ms/2). Buffers created just after a tick still had to wait up to ~1.5x the configured value before being flushed.

Replace the ticker with a self-adjusting time.Timer that fires at the exact moment the oldest live buffer is due to expire:

- computeNextFlushDelay() scans all shards for the earliest expiry deadline and returns the precise duration to sleep.
- periodicFlush() resets the timer after every flush and on every newBufferCh signal.
- Both write paths send a non-blocking signal on newBufferCh whenever a new buffer is created (after a size-based flush or on first write for a key), so the timer is re-evaluated immediately and set to fire at T+maxAge rather than the previous interval boundary.

Worst-case flush delay drops from ~1.5x to ~1.0x the configured max_buffer_age_ms with negligible overhead.

Fixes #142

### DIFF
--- a/internal/ingest/arrow_writer.go
+++ b/internal/ingest/arrow_writer.go
@@ -688,6 +688,7 @@ type ArrowBuffer struct {
 
 	// Flush timeout for storage writes (prevents workers from blocking forever on S3 hangs)
 	flushTimeout time.Duration
+	maxBufferAge time.Duration // pre-calculated from cfg.MaxBufferAgeMS
 
 	// Metrics (using atomic operations to avoid lock contention)
 	totalRecordsBuffered atomic.Int64
@@ -814,6 +815,7 @@ func NewArrowBuffer(cfg *config.IngestConfig, storage storage.Backend, logger ze
 		flushQueue:      make(chan flushTask, queueSize),
 		flushWorkers:    flushWorkers,
 		flushTimeout:    flushTimeout,
+		maxBufferAge:    time.Duration(cfg.MaxBufferAgeMS) * time.Millisecond,
 		sortKeysConfig:  sortKeysConfig,
 		defaultSortKeys: defaultSortKeys,
 		logger:          logger.With().Str("component", "arrow-buffer").Logger(),
@@ -1696,7 +1698,7 @@ func (b *ArrowBuffer) periodicFlush() {
 // goroutine sleeps until a new buffer signals it via newBufferCh.
 func (b *ArrowBuffer) computeNextFlushDelay() time.Duration {
 	now := time.Now()
-maxAge := b.maxBufferAge
+	maxAge := b.maxBufferAge
 	earliest := now.Add(maxAge) // default when no buffers exist
 
 	for _, shard := range b.shards {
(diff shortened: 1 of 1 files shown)


xe-nvdk:
@claude fix the compilation errors and the busy-loop risk:

1. Add `maxBufferAge time.Duration` field to the `ArrowBuffer` struct
2. Initialize it in `NewArrowBuffer` as `time.Duration(cfg.MaxBufferAgeMS) * time.Millisecond`
3. Fix the indentation on the two lines that reference `b.maxBufferAge` (missing leading tab)
4. In `computeNextFlushDelay`, add a floor of `time.Millisecond` instead of returning `0` to prevent busy-looping when all buffers are overdue

Issue #357, opened 2026-03-27
critical(compaction): batch filename collision causes up to 84% data loss


xe-nvdk:
When partitions have >30 files, SplitCandidateIntoBatches creates sequential batches. Each batch generates output filename with second-precision timestamp. All batches within the same second produce identical filenames, causing each batch to overwrite the previous. Only the last batch survives. Affects any measurement with >30 files/hour (e.g., max_buffer_age_ms=500 produces ~120 files/hour). Fix: add nanosecond precision to compacted filenames.

PR #359, opened 2026-03-27
fix(compaction): batch filename collision causes up to 84% data loss …

xe-nvdk:
…(#357)

When partitions had >30 files, SplitCandidateIntoBatches created sequential batches of 30. Each batch generated output filenames with second-precision timestamps, causing all batches within the same second to produce identical filenames. Each batch overwrote the previous, destroying up to 84% of data.

Fixed by adding UnixNano to compacted filenames for guaranteed uniqueness. Updated extractNewestFileTime to handle both old and new filename formats.


xe-nvdk:
Fix already merged into main via PR #358. This release branch was a cherry-pick for the v26.03.2 hotfix.

Issue #398, opened 2026-04-13
Resumable transfers via HTTP Range for peer replication


xe-nvdk:
## Problem
Failed file transfers restart from zero. Acceptable for small Parquet files (~tens of MB) but wasteful for large compacted outputs.

## Proposed Solution
Add a range field to \`FetchFileRequest\` so the puller can resume from a byte offset. The fetch handler streams from the offset using backend.ReadRange or io.SectionReader.

## Context
Deferred from Phase 2. Referenced in release notes and product definition.

## Effort
Medium

xe-nvdk:
Implemented in #404 (merged). Resumable transfers via HTTP Range are now live on main — failed peer-replication pulls resume from the last committed byte on retry instead of restarting from zero.

PR #404, opened 2026-04-14
feat(cluster): resumable peer-replication transfers (#398)

xe-nvdk:
## Summary

- Failed peer-replication file transfers now resume from the last committed byte instead of restarting from zero — especially valuable for large compacted Parquet outputs on slow or flaky links
- New `ByteOffset` field on `FetchFileRequest`/`FetchFileAckHeader` (omitempty); HMAC does not bind the offset (path is already bound)
- `AppendingBackend` optional interface replaces a mandatory `AppendReader` method that S3/Azure would always reject — puller type-asserts and falls back gracefully
- `bad_offset` counter split into `bad_offset_server` (AckCodeBadOffset) and `bad_offset_backend` (ErrResumeNotSupported) for cleaner observability
- Post-review refactors: `tryResumeFromPartial` helper, `writeFileTail` extracted, pipe goroutine uses `sync.WaitGroup`, coordinator uses `ReadToAt` unconditionally

## Test plan

- [ ] `go test ./internal/cluster/filereplication/...` — all existing + new tests pass (resume happy path, AckCodeBadOffset, offset echo mismatch, partial file retry, bad-offset deletes partial)
- [ ] `go test ./internal/storage/...` — ReadToAt, StatFile, AppendReader on LocalBackend
- [ ] `go test ./internal/cluster/...` — integration test: full fetch round-trip still works
- [ ] `go build ./...` — clean build
- [ ] Manual: interrupt a large file transfer mid-way, confirm second attempt resumes from offset (not from zero)
- [ ] Manual: S3/Azure backend confirms full re-fetch fallback (no partial file left behind)

### DIFF
--- a/RELEASE_NOTES_2026.05.1.md
+++ b/RELEASE_NOTES_2026.05.1.md
@@ -20,6 +20,7 @@ Every file is SHA-256 hashed at flush time and the hash is committed into a Raft
 - `GET /api/v1/cluster/files` returns the full cluster manifest (supports `?database=` filter).
 - Replication status and catch-up progress are surfaced via `/api/v1/cluster/status`.
 - Zero overhead for OSS / standalone deployments.
+- **Resumable transfers**: interrupted pulls resume from the last committed byte instead of restarting from zero. Especially valuable for large compacted Parquet outputs on slow or flaky links. On retry the puller checks how many bytes are already on disk, hashes the prefix to continue the SHA-256 chain, and requests only the remaining tail from the peer. Full-file hash verification is still enforced. S3 and Azure Blob backends fall back to a full re-fetch on resume (append is not supported by those APIs); local-SSD nodes get complete resume support.
 
 **Configuration** (`ARC_CLUSTER_REPLICATION_*` env vars or `cluster.replication_*` in `arc.toml`):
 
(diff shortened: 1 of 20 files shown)


xe-nvdk:
@gemini-code-assist All findings from your previous review have been addressed in the latest push. Here's a summary of the fixes:

**HIGH — LocalBackend.WriteReader non-resumable (staging file deleted on error)**
`WriteReader` now writes to a deterministic `<path>.part` staging file (replacing the random `.tmp`). On transport error the staging file is left in place so the puller can discover it via `StatFile`/`ReadToAt` and resume on the next attempt. On success it is atomically renamed to the final path. Unit test updated to exercise the full failed-write → append-tail → promote flow.

**MEDIUM — Azure ReadToAt sends Range header at offset=0**
`ReadToAt` now only sets `DownloadStreamOptions{Range: ...}` when `offset > 0`. At offset=0 it passes `nil` (no Range header), matching the S3 behaviour and avoiding a spurious 400 from Azure Blob Storage.

**MEDIUM — tryResumeFromPartial skips resume for non-AppendingBackend backends**
Removed the premature `AppendingBackend` type-assertion early-return from `tryResumeFromPartial`. Non-appendable backends (S3, Azure) now correctly: detect a partial file → return a non-zero offset → hit `ErrResumeNotSupported` in `writeFileTail` → increment `bad_offset_backend` → delete the partial → retry from zero. New `TestPuller_NonAppendingBackendFallback` test verifies the full path.

**fakeBackend staging semantics**
`fakeBackend` in `puller_test.go` now mirrors `LocalBackend`'s `.part` staging key semantics (WriteReader stores partial bytes under `path+".part"`, AppendReader appends and promotes on full receipt, Delete cleans up both keys). Duplicate `Delete` method merged.

Please re-review when you get a chance. Thanks!

PR #405, opened 2026-04-16
feat(enterprise): Helm chart + Traefik compose examples

xe-nvdk:
## Summary

- **Arc Enterprise Helm chart** (`helm/arc-enterprise/`) supporting both documented deployment patterns (shared object storage with bundled MinIO, local storage with peer replication) via a single `storage.mode` value. Role-separated StatefulSets, per-pod Raft bootstrap via a sh entrypoint wrapper, fail-fast install validation for missing secrets, and production-secure defaults (non-root MinIO with console disabled, no weak default credentials, reader `Service` `ClusterIP`, seccomp `RuntimeDefault`).
- **Traefik v3.6** replaces nginx in both docker-compose examples (`deploy/docker-compose/`, `deploy/docker-compose-local/`). Routing declared via container labels through the Docker provider — writers/readers load-balance automatically as new containers come up.
- **Release notes 26.05.1** gain a Deployment section covering the chart, Traefik examples, and deployment-patterns doc.

### Chart highlights addressed from senior reviews

- HA Raft bootstrap works correctly with `writer.replicas=3` (only pod ordinal `-0` bootstraps). `writer.replicas=2` is explicitly rejected to prevent a Raft split-brain.
- Validation `fail`s catch missing license / shared secret / TLS existingSecret / MinIO credentials / external-S3 credentials / invalid `storage.mode` at `helm install` time instead of pods stuck in `CreateContainerConfigError`.
- WAL enabled by default on writers; cluster TLS wires the CA certificate for mTLS; `ARC_AUTH_BOOTSTRAP_TOKEN` plumbed through `auth.bootstrapToken`; peer-replication tuning and `ARC_STORAGE_S3_PREFIX` exposed as values.
- Bundled MinIO runs non-root, console disabled, pinned to a release tag.
- Reader `Service` defaults to `ClusterIP`; per-role `nodeSelector` / `tolerations` / `affinity` fall back to the global values.

## Test plan

- [x] `helm lint helm/arc-enterprise/` — clean (1 chart, 0 failures)
- [x] `helm template` renders both `values-shared-storage.yaml` (13 resources) and `values-local-storage.yaml` (10 resources) without errors
- [x] Validation `fail`s fire for: `writer.replicas=2`, `storage.mode=invalid`, `cluster.tls.enabled=true` without `existingSecret`, `storage.shared.external=true` without credentials
- [x] HA rendering (`writer.replicas=3`) produces the sh wrapper that bootstraps only ordinal `-0` and generates the correct seed list
- [x] `docker compose config` validates both new Traefik compose files
- [ ] Smoke test: `helm install` on a kind cluster with each preset (manual, before tagging the release)
- [ ] Smoke test: `docker compose up` in both `deploy/docker-compose/` folders (manual)


xe-nvdk:
@gemini-code-assist both findings addressed in 84b32ef:

1. **values.schema.json** (new) enforces types, enums, ranges, and shape — `storage.mode`, `writer.replicas >= 1`, service type enums, image pullPolicy, replication tunable ranges, PVC size non-empty, etc. JSON Schema errors surface before template rendering with clear `at '/path': ...` messages.

2. **templates/_validation.tpl** (new) owns the cross-field rules that JSON Schema can't express cleanly — "X required when Y" with `lookup()` checks for upgrade safety (license, sharedSecret, TLS, MinIO creds, external-S3 creds), plus the `writer.replicas=2` split-brain message which stays as a `fail()` so it keeps its human explanation.

3. **templates/_helpers.tpl** lost its inline `validate` block; it now just documents where validation lives.

All edge cases still covered by `helm template` smoke tests (typo → schema error, tls.enabled w/o secret → cross-field fail, external S3 w/o creds → cross-field fail, replicas=2 → explanatory message, replicas=3 → HA renders).

PR #407, opened 2026-04-20
feat(cluster): writer-only retention with Raft manifest propagation

xe-nvdk:
## Summary

- Only the primary writer node runs retention policies in cluster mode, preventing races on shared (S3/Azure/NFS) and per-node (local) storage
- After each file delete, `DeleteFileFromManifest` commits the removal into the Raft log, keeping the cluster manifest consistent
- Reader nodes clean up their local copies via the existing `onFileDeleted` FSM callback and delete-worker pool (shared with compaction — no new code needed)
- Standalone deployments are completely unaffected (nil gate = no check)

## Test plan

- [x] `go build ./cmd/... ./internal/...` passes clean
- [x] `go test ./internal/scheduler/... ./internal/api/... ./internal/cluster/...` all pass
- [x] Standalone: retention runs normally, no gate log messages
- [x] Cluster (3-node): only writer logs "Triggering scheduled retention"; readers log "Scheduler idle"
- [x] Cluster: after retention runs, `GET /api/v1/cluster/files` shows no orphaned entries for deleted files
- [x] Writer failover: new primary picks up retention on next cron tick

### DIFF
--- a/RELEASE_NOTES_2026.05.1.md
+++ b/RELEASE_NOTES_2026.05.1.md
@@ -149,6 +149,14 @@ Directories containing sensitive data (auth database, continuous query definitio
 
 Applied defense-in-depth SQL escaping to DuckDB `SET memory_limit` (single-quote escaping) and compaction `ORDER BY` sort key names (double-quote identifier escaping). Both already had config-level validation, but the runtime escaping provides an additional safety layer.
 
+### Cluster-Safe Retention Policy Execution (Enterprise)
+
+In clustered deployments, retention policies now run exclusively on the primary writer node and propagate file deletions through the Raft manifest to all peers.
+
+- **Writer-only execution**: The retention scheduler checks a cluster gate before starting. Only the primary writer node runs — reader and compactor nodes stay idle and log a clear message. On writer failover, the newly promoted primary picks up on the next cron tick. Standalone deployments are unaffected.
+- **Cluster manifest updates**: After each file is deleted from storage, the retention handler commits the deletion into the Raft log via `DeleteFileFromManifest`. This keeps the manifest consistent and prevents orphaned entries from interfering with peer replication catch-up.
+- **Reader node cleanup (local storage)**: The `onFileDeleted` FSM callback and delete-worker pool (shared with compaction) handle retention-triggered deletes on reader nodes, removing their local copy of the file and preventing unbounded disk growth on per-node storage deployments.
+
 ## Bug Fixes
 
 ### Row-Format MessagePack Flush Hardening
--- a/cmd/arc/main.go
+++ b/cmd/arc/main.go
@@ -1763,7 +1763,11 @@ func newRetentionClusterGate(c *cluster.Coordinator) *retentionClusterGate {
 }
 
 func (g *retentionClusterGate) CanRunRetention() bool {
-	return g.coordinator.GetLocalNode().IsPrimaryWriter()
+	node := g.coordinator.GetLocalNode()
+	if node == nil {
+		return false
+	}
+	return node.IsPrimaryWriter()
 }
 
 func (g *retentionClusterGate) Role() string {
--- a/cmd/arc/main.go
+++ b/cmd/arc/main.go
@@ -1320,7 +1320,7 @@ func main() {
 	var retentionHandler *api.RetentionHandler
 	if cfg.Retention.Enabled {
 		var err error
-		retentionHandler, err = api.NewRetentionHandler(storageBackend, db, &cfg.Retention, authManager, logger.Get("retention"))
+		retentionHandler, err = api.NewRetentionHandler(storageBackend, db, &cfg.Retention, licenseClient, authManager, logger.Get("retention"))
 		if err != nil {
 			log.Fatal().Err(err).Msg("Failed to initialize retention handler")
 		}
--- a/cmd/arc/main.go
+++ b/cmd/arc/main.go
@@ -1325,6 +1325,9 @@ func main() {
 			log.Fatal().Err(err).Msg("Failed to initialize retention handler")
 		}
 		retentionHandler.RegisterRoutes(server.GetApp())
+		if clusterCoordinator != nil {
+			retentionHandler.SetCoordinator(clusterCoordinator)
+		}
 		shutdownCoordinator.RegisterHook("retention", func(ctx context.Context) error {
 			return retentionHandler.Close()
 		}, shutdown.PriorityDatabase)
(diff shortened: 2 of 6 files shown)


xe-nvdk:
@gemini-code-assist Both findings from your review have been addressed in the latest push:

1. **Batch manifest updates** — replaced the per-file `DeleteFileFromManifest` calls with a single `BatchFileOpsInManifest` call after the deletion loop, matching how compaction handles manifest updates.

2. **Per-tick gate check** — moved `CanRunRetention()` out of `Start()` and into `runRetention()` so role transitions (failover, demotion) take effect on the next cron tick without requiring a node restart. `Status()` now reports the live `can_run` value instead of a stale startup snapshot.

Could you do another pass?

xe-nvdk:
@gemini-code-assist Three findings from the second review addressed:

1. **Chunked manifest updates** — batch is now split into 1000-op chunks to prevent oversized Raft log entries on large retention runs.
2. **Error log + honest comment** — upgraded from `Warn` to `Error` on manifest failure; removed the misleading "self-heal" comment and replaced with an explicit note that affected paths will be orphaned in the manifest until Phase 5 reconciliation ships.
3. **Removed unnecessary mutex** — `clusterGate` is immutable after construction; removed the lock/copy pattern in both `runRetention` and `TriggerNow`.

Could you do a final pass?

xe-nvdk:
@gemini-code-assist Three findings from the latest round addressed:

1. **Manifest-before-storage ordering** — the manifest is now updated *before* deleting from storage. If the Raft update fails, the file still exists in both places and the next retention run will retry it naturally. A storage-delete failure after a successful manifest update leaves a harmless ghost file on disk (invisible to the cluster, will be swept by the next compaction pass).

2. **Abort on manifest failure** — if `BatchFileOpsInManifest` fails, the retention cycle returns an error immediately rather than continuing to the next chunk. A Raft quorum loss is not transient; proceeding would only produce more orphans for the same root cause.

3. **Removed redundant `!dryRun` guard** — `deletedFilePaths`/`eligiblePaths` is only populated when `dryRun=false`, so the outer check was redundant.

Could you do a final pass?

xe-nvdk:
@gemini-code-assist Both findings addressed:

1. **Interleaved manifest+storage per chunk** — instead of two separate passes (all manifest updates then all storage deletes), each chunk now updates the manifest and immediately deletes from storage before moving to the next chunk. This limits the orphan blast radius to one chunk on a mid-run failure, matching your suggested implementation.

2. **SQL injection fix in `getFileMaxTimeAndRowCount`** — single quotes in the file path are now escaped (`strings.ReplaceAll(filePath, "'", "''")`) before interpolation into the `read_parquet()` query, since DuckDB doesn't support parameterized `read_parquet()` calls.

Could you do a final pass?

xe-nvdk:
@gemini-code-assist Logic bug fixed: `subPaths` and `subRows` are now built in parallel with `ops` during the marshal loop, so a `json.Marshal` failure skips the file from both the manifest batch **and** the storage delete loop. Previously, a marshal failure would silently exclude the file from `ops` but leave it in `chunk`, causing a storage delete without a corresponding manifest removal.

Could you do a final pass?

xe-nvdk:
@gemini-code-assist Addressed — replaced `[:0:0]` with `make([]string, 0, len(chunk))` / `make([]int64, 0, len(chunk))` for clarity. Could you do a final pass?

xe-nvdk:
@gemini-code-assist Four issues addressed from Gemini's latest round and a parallel principal-level review:

1. **Concurrent execution guard** — added `runningJob bool` flag to `RetentionScheduler`. If a previous cycle is still in progress when the next tick fires, the tick is skipped with a warn log. Applies to both the cron path and `TriggerNow`.

2. **Early abort on manifest error in `ExecutePolicy`** — when `deleteOldFiles` returns a manifest error and a coordinator is wired (cluster mode), `ExecutePolicy` now records the execution as `failed` and returns immediately instead of continuing to the next measurement. Storage-only errors (standalone mode) still continue as before.

3. **License check in `ExecutePolicy`** — added a license gate directly in `ExecutePolicy` so direct programmatic/API calls that bypass the scheduler's `Start()` check are also protected.

4. **`licenseClient` wired into `RetentionHandler`** — constructor updated to accept `*license.Client`; call site in `main.go` and test updated accordingly.

Could you do a final pass?

xe-nvdk:
@gemini-code-assist Applied the `runningJob` guard to `TriggerNow` as well — a manual trigger now returns an error immediately if a scheduled cycle is already in progress, matching the behaviour added to `runRetention`. Could you do a final pass?

xe-nvdk:
Three additional fixes from a parallel staff/principal-engineer review:

1. **`ExecutePolicy` abort in standalone mode** — removed the `if h.coordinator != nil` guard around the abort path. Standalone mode was silently `continue`-ing on errors and producing partial inconsistent state; now aborts the entire cycle on first failure in both modes.

2. **`handleExecute` context propagation** — was calling `deleteOldFiles(context.Background(), ...)`, changed to `c.Context()` so client disconnects and server timeouts cancel in-flight deletions. Also records execution as "failed" and returns 500 on error instead of swallowing it.

3. **`retentionClusterGate.CanRunRetention` nil safety** — `GetLocalNode()` can return nil during cluster startup; added guard to return `false` instead of panicking.

@gemini-code-assist please do a final review pass on these three commits.

xe-nvdk:
Fixed the reader-node API race flagged in the latest review.

`handleExecute` now checks `h.coordinator.CanRunRetention()` before executing any mutations — a reader node gets a `503 Service Unavailable` with the node's current role in the error message. Dry-run requests are allowed on any node since they make no mutations.

Implementation: extended `RetentionCoordinator` interface with `CanRunRetention() bool` and `Role() string`, implemented both on `*cluster.Coordinator` (delegates to `GetLocalNode().IsPrimaryWriter()` with nil safety). Standalone mode is unaffected — `coordinator` is nil, check is skipped.

@gemini-code-assist please do a final pass.

xe-nvdk:
Two more fixes from the latest review:

1. **Policy ID in manifest reason** — `deleteOldFiles` now takes a `reason string` parameter; callers pass `"retention:<policy_id>"` (e.g. `"retention:42"`). This propagates to `DeleteFilePayload.Reason` in every Raft manifest op, making delete-worker traces and Raft log entries auditable by policy.

2. **Storage error severity** — `storage.Delete` failures downgraded from `Error` to `Warn`. A network blip or already-deleted file is transient; the cycle continues to the next file. Manifest failures remain `Error` + abort since Raft quorum loss is non-transient.

@gemini-code-assist please do a final pass.

xe-nvdk:
Fixed the partial-progress counter issue.

`totalDeleted` and `totalFilesDeleted` are now accumulated *before* the error check in both `ExecutePolicy` and `handleExecute`. `deleteOldFiles` returns its partial progress even on abort (rows/files deleted in completed chunks before the manifest failure), so the execution record now accurately reflects all completed work rather than dropping counts from the failed measurement.

@gemini-code-assist please do a final pass.

PR #408, opened 2026-04-20
feat(delete): cluster-safe DELETE endpoint — writer gate + Raft manifest

xe-nvdk:
## Summary

- **Writer-only gate**: reader nodes return `503 Service Unavailable` before any mutation; dry-run is allowed on any node
- **Manifest-before-storage for full-file deletes**: when all rows in a file match the WHERE clause, `BatchFileOpsInManifest` is called before `storage.Delete`. Raft propagates the `CommandDeleteFile` to all nodes — reader nodes' `onFileDeleted` FSM callback cleans up local copies. Partial rewrites (file path unchanged) need no manifest entry.
- **Abort on manifest failure**: `errManifestFailure` sentinel causes the operation to abort and return 500; continuing would delete files from storage with no manifest record
- **Storage failure after committed manifest**: logged as Warn, treated as success — manifest is authoritative, orphan cleaned up by Phase 5 reconciliation
- **Single-quote escaping** on all `read_parquet()` path interpolations (SQL injection hardening)
- **`c.Context()` propagation**: replaces `context.Background()` so client disconnects cancel in-flight rewrites
- **Standalone mode unchanged**: coordinator is nil, all gates skipped

Mirrors the retention PR (#407) pattern exactly.

## Test plan

- [x] `go test ./internal/api/... -v` — all pass
- [x] Standalone: issue a delete with `confirm=true` — files rewritten/deleted, no errors
- [x] Reader node (cluster): issue delete → expect `503` with role in error message
- [x] Full-delete (cluster): after deleting all rows in a file, confirm path absent from Raft manifest on all nodes
- [x] Failover: stop writer mid-operation, promote standby → next delete succeeds on new writer

### DIFF
--- a/RELEASE_NOTES_2026.05.1.md
+++ b/RELEASE_NOTES_2026.05.1.md
@@ -157,6 +157,14 @@ In clustered deployments, retention policies now run exclusively on the primary 
 - **Cluster manifest updates**: After each file is deleted from storage, the retention handler commits the deletion into the Raft log via `DeleteFileFromManifest`. This keeps the manifest consistent and prevents orphaned entries from interfering with peer replication catch-up.
 - **Reader node cleanup (local storage)**: The `onFileDeleted` FSM callback and delete-worker pool (shared with compaction) handle retention-triggered deletes on reader nodes, removing their local copy of the file and preventing unbounded disk growth on per-node storage deployments.
 
+### Cluster-Safe DELETE Endpoint (Enterprise)
+
+The `POST /api/v1/delete` endpoint is now cluster-aware. Reader nodes reject delete requests with `503 Service Unavailable` — only the primary writer may execute mutations.
+
+- **Writer-only gate**: Checked before any storage scan, so reader nodes are rejected immediately without running expensive DuckDB queries.
+- **Manifest-before-storage for full-file deletes**: When a DELETE removes an entire file (all rows match the WHERE clause), the path is committed to the Raft manifest before `storage.Delete`. All peer nodes learn about the removal via the `onFileDeleted` FSM hook and clean up their local copies. Partial rewrites (file overwritten in-place, path unchanged) require no manifest entry.
+- **Path traversal hardening**: `database` and `measurement` inputs are now validated to reject `..`, `/`, and `\` characters.
+
 ## Bug Fixes
 
 ### Row-Format MessagePack Flush Hardening
--- a/cmd/arc/main.go
+++ b/cmd/arc/main.go
@@ -1303,6 +1303,9 @@ func main() {
 	// Register Delete handler
 	deleteHandler := api.NewDeleteHandler(db, storageBackend, &cfg.Delete, authManager, logger.Get("delete"))
 	deleteHandler.RegisterRoutes(server.GetApp())
+	if clusterCoordinator != nil {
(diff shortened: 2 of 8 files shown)


xe-nvdk:
@gemini-code-assist please review this PR.

xe-nvdk:
@gemini-code-assist Pushed two additional commits addressing your partial-rewrite manifest feedback:

**What changed:**
- `CommandUpdateFile` — new Raft FSM command that replaces a `FileEntry` in-place (size + SHA256) without triggering `onFileRegistered` (no unnecessary peer pull)
- `Node.UpdateFile` — applies the command via Raft with leader-forwarding support
- `Coordinator.UpdateFileInManifest` + `Coordinator.GetFileEntry` — the caller reads the existing manifest entry first, copies all immutable fields (database, measurement, tier, originNodeID, createdAt), then updates only `SizeBytes`/`SHA256`
- `DeleteCoordinator` interface extended with the two new methods
- `rewriteFileWithoutDeletedRows` calls `updateManifestAfterRewrite` after a successful partial rewrite; failure is non-fatal (Warn log) since the rewritten file is still queryable
- Path containment check added in `updateManifestAfterRewrite` before opening the rewritten file for SHA256 computation (defense-in-depth)

**Design notes:**
- S3 backend: SHA256 re-computation would require re-downloading; the entry is re-committed with original metadata but the Raft log index advances so peers see the change
- Full-file delete path is untouched — still uses `CommandDeleteFile` via `BatchFileOpsInManifest`

Please review for correctness, failure modes, and any gaps in the `applyUpdateFile` secondary-index handling.

xe-nvdk:
@gemini-code-assist Addressed both findings:

1. **`applyUpdateFile` now calls `onFileRegistered`** — reader nodes will detect the content change via the updated manifest entry and re-fetch the rewritten file from the writer. The callback is extracted before `Unlock()` and invoked after, matching the pattern in `applyRegisterFile`.

2. **S3 SHA256/size now accurate** — `rewriteS3File` computes size and SHA256 from the local temp file (already on disk before upload) before streaming it to S3. The result is threaded through to `updateManifestAfterRewrite`, which applies it to the manifest entry instead of leaving stale metadata.

Please re-review.

xe-nvdk:
@gemini-code-assist Addressed all findings from the latest review:

**High — Azure backend falls through to `rewriteLocalFile`**
Replaced `isS3Backend()` with `isRemoteBackend()` which covers both `*storage.S3Backend` and `*storage.AzureBlobBackend`. Also updated `getQueryPath` to build `azure://{container}/{path}` URLs for DuckDB reads from Azure Blob (matching the pattern in `tiering/router.go`).

**Medium — `FailedFiles` missing from manifest-failure abort response**
Added `FailedFiles: failedFiles` to the abort response so callers know which files were already processed before the quorum loss.

**Medium — `totalDeleted` incremented before error check**
Moved `totalDeleted += deleted` to after the error check so the count only reflects successfully processed files.

**Medium — Double-read of temp file for SHA256 + upload**
Replaced the separate `fileMetadata` call with `io.TeeReader` — SHA256 is now computed in a single pass during the upload stream, eliminating the extra read.

The stale `onFileRegistered` and S3 metadata comments were already addressed in the previous push — those findings are resolved.

PR #409, opened 2026-04-22
feat(cluster): writer-only CQ execution with Raft writer gate

xe-nvdk:
## Summary

- Adds `CQClusterGate` interface to `internal/scheduler/cq_scheduler.go` — checked at every tick in `runJob` so role transitions (failover, demotion) take effect without a restart
- Adds `CQCoordinator` interface to `internal/api/continuous_query.go` — gates `POST /:id/execute` with `503 Service Unavailable` when the node is not the primary writer
- Wires both interfaces in `cmd/arc/main.go` via the `cqClusterGate` adapter; nil when clustering is disabled (standalone/OSS — gate bypassed)
- Adds per-job overlap guard (`executing bool` + `sync.Mutex`) to prevent tick overlap when a CQ execution takes longer than the configured interval
- Mirrors the `RetentionClusterGate` pattern from #407 and `writerGate` from #408

## Behavior

| Node role | `POST /:id/execute` | Scheduler tick |
|-----------|---------------------|----------------|
| Primary writer | `200 OK` | Executes CQ |
| Non-primary writer / reader | `503 Service Unavailable` | Skips (logs `CQ tick skipped: node is not primary writer`) |
| Standalone (no cluster) | `200 OK` (no gate) | Executes CQ (no gate) |

## Test plan

- [x] `go build ./cmd/... ./internal/...` — clean build
- [x] `go test ./internal/scheduler/... -v` — all tests pass (10 CQ scheduler tests, 5 gate-specific)
  - `TestCQScheduler_ClusterGate_NilGate` — nil gate in standalone mode
  - `TestCQScheduler_ClusterGate_ReaderSkipsTick` — reader gate blocks execution
  - `TestCQScheduler_ClusterGate_WriterAllowsTick` — writer gate allows execution
  - `TestCQScheduler_ClusterGate_FailoverTransition` — gate atomically flips on failover without restart
  - `TestCQScheduler_OverlapGuard_ExecutingFlag` — overlap guard initializes to false
- [x] Integration: standalone node — CQ creation (201) and execute (no 503) confirmed via Docker test
- [ ] Integration Phase B (cluster): requires enterprise license with `clustering` feature — Phase A tests pass, Phase B ready to run with valid license key

### DIFF
--- a/RELEASE_NOTES_2026.05.1.md
+++ b/RELEASE_NOTES_2026.05.1.md
@@ -235,6 +235,26 @@ Fixed a memory retention issue where running a retention policy or the delete AP
 
 **Impact:** Memory usage should return to baseline after a retention policy run or delete operation, instead of climbing GBs per execution. Particularly noticeable in Docker/Kubernetes deployments with memory limits, where this could cause OOM kills during nightly retention runs.
 
+### Writer-Only Schedulers Skipped All Ticks Without Failover Enabled (Enterprise)
+
+Fixed a bug where the CQ and retention schedulers skipped execution on every tick when `cluster.failover_enabled=false` — which is the default and the typical single-writer cluster configuration.
+
+**Root cause:** `IsPrimaryWriter()` on a cluster node returns `true` only when `WriterState == WriterStatePrimary`. That state is set exclusively via the writer failover manager's `CommandPromoteWriter` Raft entry. When failover is disabled, no promotion command is ever issued, so `WriterState` remains at its zero value and `IsPrimaryWriter()` always returns `false` — even on writer nodes. Both the retention and CQ gate adapters in `main.go` (`retentionClusterGate`, `cqClusterGate`) called `node.IsPrimaryWriter()` directly, bypassing the coordinator entirely.
+
+**Fix:**
+- `Coordinator.IsPrimaryWriter()` now falls back to a role check (`node.Role == RoleWriter`) when no failover manager is configured. With failover enabled, the existing `WriterState == WriterStatePrimary` semantics are preserved.
+- `retentionClusterGate` and `cqClusterGate` now delegate to `coordinator.IsPrimaryWriter()` instead of calling `node.IsPrimaryWriter()` directly, so the fallback is respected.
+
+**Impact:** Retention policies and continuous query schedules now execute correctly on writer nodes in clusters running without automatic failover. Previously both would silently skip every scheduled tick, meaning retention was never applied and CQs never ran in the default cluster configuration.
+
+### Continuous Query Not Scheduled After API Creation
+
+Fixed a bug where a CQ created via `POST /api/v1/continuous_queries` was not picked up by the scheduler until the node was restarted.
+
+**Root cause:** `handleCreate` inserted the CQ into SQLite but did not notify the scheduler. The scheduler's `ReloadCQ` was only called from `handleUpdate`.
+
+**Fix:** `handleCreate` now calls `scheduler.ReloadCQ(queryID)` after a successful insert. If the scheduler is not running (no license, or standalone without license), the call is a no-op.
+
 ## Dependencies
 
 ### DuckDB v1.5.1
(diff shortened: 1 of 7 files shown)


xe-nvdk (commented):
Addressed — removed the  +  fields from , the corresponding lock/check/set block in , and the associated test. Gemini is correct that the guard was dead code:  is a single goroutine and  blocks synchronously, so  was always  when checked.

xe-nvdk:
Updates pushed — addressed all findings from the first review. @gemini-code-assist please do a second pass.

xe-nvdk:
Addressed — `handleCreate` now calls `scheduler.StartJobDirect(queryID, req.Name, req.Interval, req.IsActive)` instead of `ReloadCQ`, passing the data already in hand so no SQLite re-read is needed. Added `StartJobDirect` to both the `CQSchedulerReloader` interface and `CQScheduler`. The other three inline comments from this review are stale — the overlap guard and its test were removed in the previous push.

xe-nvdk:
@gemini-code-assist please do another pass.

xe-nvdk:
All four findings from Review 3 addressed in commit 72b6ff7:

1. **`IsPrimaryWriter` docstring** — updated to list `api.CQCoordinator` and `scheduler.WriterGate` alongside the existing interfaces.

2. **Duplicate `CQClusterGate` interface** — interface definition removed from `cq_scheduler.go`; `CQClusterGate` is now a type alias for the shared `WriterGate` defined in `retention_scheduler.go`. Same for `RetentionClusterGate`. Both aliases kept for backwards compatibility.

3. **`stopJobLocked` helper** — already extracted in the previous commit. `StartJobDirect`, `ReloadCQ`, `ReloadAll`, and `Stop` all call it.

4. **Merged `cqClusterGate` + `retentionClusterGate`** — both removed; replaced by a single `writerClusterGate` struct in `main.go`. Both call sites now use `newWriterClusterGate`.

@gemini-code-assist please do a final pass.

xe-nvdk:
All three findings from Review 4 addressed in commit f771eb0:

1. **`ReloadCQ` running guard** — added `if !s.running { return nil }` at the top (under the lock) so a stopped scheduler never restarts jobs.

2. **`StartJobDirect` running guard** — same guard added; `handleCreate` calling this when the scheduler is inactive is now a clean no-op.

3. **Release notes** — corrected to say `StartJobDirect(queryID, name, interval, isActive)` (not `ReloadCQ`), with a note that the call is a no-op when the scheduler is not running.

@gemini-code-assist please do a final pass.

xe-nvdk:
Thanks for the follow-up. Respectfully pushing back on this one — the earlier Review 1 finding was correct and we should keep the guard removed.

**Why the overlap guard is not needed here (and was correctly removed):**

`runJob` is a single goroutine with a blocking `select`. `executeJob` is called synchronously — no `go` keyword, no goroutine spawn. The goroutine is blocked inside `executeJob` for the entire duration of the query. The `time.Ticker` channel has a buffer of 1, so if a tick fires while the job is executing, exactly one tick queues up; all others are dropped. When `executeJob` returns, the goroutine processes that one queued tick and then blocks again. There is no concurrency, no fan-out, and no resource exhaustion possible.

**Why the retention scheduler is different:**

`RetentionScheduler` uses `cron.Cron`, which calls its registered function in a **new goroutine on every tick** — that's the default `cron.Cron` behavior. Without the `runningJob` flag, two retention cycles could genuinely run concurrently. The flag is necessary there precisely because of that execution model.

**The PR description / test plan discrepancy:**

The PR description mentioned an overlap guard early in the draft. Review 1 correctly identified it as dead code (the `executing` flag was always `false` when checked because `runJob` is single-threaded), and we removed it along with `TestCQScheduler_OverlapGuard_ExecutingFlag`. The PR description wasn't updated to reflect that — that's the only gap, not a missing implementation.

@gemini-code-assist please do a final pass with this context in mind.

xe-nvdk:
All four findings from Review 5 addressed in commit 86a2698:

1. **`ReloadAll` missing running guard** — fixed. Same `!s.running` early return added under the lock, consistent with `ReloadCQ` and `StartJobDirect`.

2. **Overlap on reload** — conceded and fixed properly. Added `done chan struct{}` to `cqJob`, closed by `runJob` on exit. `stopJobLocked` now returns it. `ReloadCQ`, `StartJobDirect`, and `ReloadAll` all release the lock, wait on `<-done` if a job was running, then re-acquire the lock to start the replacement. The old goroutine is guaranteed to have exited (including any in-flight `executeJob`) before the new one starts.

3. **`CQClusterGate` alias framing** — removed the misleading "Deprecated" tag. Comment now reads: "Both schedulers share the same interface; this alias lets existing code compile without changes."

4. **`cq_name` in `stopJobLocked` log** — added back.

---

Re the overlap guard pushback from my last comment: I was wrong about the `stopCh` being sufficient. Closing `stopCh` signals the goroutine but doesn't block until it exits — if `executeJob` is mid-query, the new job could start before the old one finishes. The `done` channel fix is the correct solution.

@gemini-code-assist this has had 5+ review rounds. Please do a thorough final review and approve if there are no remaining blockers. We'd like to merge this.

PR #411, opened 2026-04-23
fix(ingest): ingestion correctness hardening and performance — 26.05.1 pre-release

xe-nvdk:
## Summary

- **Multi-hour flush atomicity**: tiering/manifest registration moved to a collect-then-register pattern — all hour-bucket Parquet files must succeed before any `registerFileInTiering` call, preventing the manifest from advancing ahead of storage on a mid-loop write failure
- **WAL error observability**: WAL write failures now increment a `total_wal_errors` atomic counter exposed via the stats endpoint, giving operators an alertable signal beyond log scraping
- **Line protocol timestamp bounds**: raw timestamp values are bounds-checked before precision multipliers (`ms`, `s`) are applied; values that would overflow `int64` fall back to server time instead of wrapping silently into plausible-looking epoch timestamps
- **Column signature caching**: hot-path schema change detection string (sorted+joined column names) is now computed once when the `TypedColumnBatch` is constructed and cached as a field — zero allocation on stable-schema writes
- **Parquet writer properties pre-built**: `WriterProperties` and `ArrowWriterProperties` are built once in `NewArrowWriter` and reused across all flushes instead of being reconstructed on every flush
- **Sort permutation reuse**: `sortTypedColumnBatchByKeys` now computes the sort permutation once and reuses it for validity bitmap reordering, eliminating a duplicate sort pass
- **`mergeBatches` colInfo value types**: changed from pointer map to value map, removing per-column heap allocations during batch merges
- **`BatchToColumnar` tag lookup reduction**: per-record `tagSet` built once and reused for both schema-discovery and data-fill passes, halving tag map lookups

## Test plan

- [x] `go test ./internal/ingest/... -v -race`
- [x] `go test ./internal/api/... -v`
- [x] `go test ./internal/wal/... -v`
- [x] Sustained bench at 12 workers — baseline 19.2M RPS / p50 0.49ms — confirm no regression
- [x] Send a multi-hour msgpack columnar payload; verify two Parquet files created and manifest consistent; verify WAL replay does not duplicate them
- [x] Send a line protocol record with a timestamp that overflows `int64` at `s` precision — verify it is accepted with server time rather than wrapping


xe-nvdk:
@gemini-code-assist Please review this PR.

**Context:** This is a pre-release hardening and performance pass over Arc's ingestion pipeline (MessagePack columnar, MessagePack row, and Line Protocol paths) for the 26.05.1 Enterprise GA release. A staff/principal-level multi-agent audit was run over the complete pipeline end-to-end before implementation.

**Key areas to focus on:**

1. **Multi-hour flush atomicity** (`arrow_writer.go` — `flushPartitionedData` multi-hour branch): The collect-then-register pattern ensures `registerFileInTiering` is only called after all `storage.Write` calls succeed. Verify no manifest entries can be orphaned under partial failure.

2. **WAL error counter** (`arrow_writer.go` — `totalWALErrors` field): Confirm the counter is incremented at all WAL write failure sites and exposed correctly in `GetStats()`.

3. **Line protocol timestamp bounds** (`lineprotocol.go` — `parseLineWithPrecision`): The `const maxInt64` guard prevents silent `int64` overflow before precision multipliers are applied. Verify all four precision cases (`us`, `ms`, `s`, `ns`) are handled correctly.

4. **Column signature caching** (`arrow_writer.go` — `TypedColumnBatch.Signature`): Confirm the cached signature is propagated correctly through `sortTypedColumnBatchByKeys` and that no path leaves it stale relative to the actual column set.

5. **Sort permutation reuse** (`arrow_writer.go` — `sortTypedColumnBatchByKeys`): The `*WithPermutation` variants return a nil index slice when data is already sorted (fast path). Verify validity bitmap reordering is skipped correctly on the nil-indices path.

xe-nvdk:
@gemini-code-assist All three findings from your previous rounds have been addressed in the latest commit (`fbecc83`). Please do another pass.

**Changes made:**

1. **Column signature now type-aware** (`arrow_writer.go` — `getColumnSignature`): Signature now encodes `name:type` pairs (`i64`, `f64`, `str`, `bool`, `dec`) so a type change on the same column name triggers schema evolution detection and a buffer flush. `TestGetColumnSignature` updated with typed slices and an explicit type-change assertion.

2. **Negative timestamps allowed** (`lineprotocol.go` — `parseLineWithPrecision`): Removed the incorrect `rawTs >= 0` guard. `strconv.ParseInt` already returns `ErrRange` for out-of-range values (handled above), so rawTs is always a valid int64 here. Pre-epoch (negative) timestamps now pass through. `us` and `ns` have no overflow guard (none needed). `ms` and `s` now check both upper and lower bounds (`rawTs >= minInt64/multiplier`).

3. **`tagSet` removed** (`lineprotocol.go` — `BatchToColumnar`): Replaced the redundant `tagSet map[string]struct{}` allocations with direct `record.Tags[key]` lookups in both the schema-discovery and data-fill loops.

xe-nvdk:
@gemini-code-assist Applied your suggestion in `abbe158`. Went one step further than the `strings.Builder` swap — pre-computed the exact final byte count before `Grow()` so the builder does a single allocation with no growth re-allocations. Please do another pass.

xe-nvdk:
@gemini-code-assist Both findings addressed in `86b8658`. Please do a final pass.

**Changes made:**

1. **Single-pass size accumulation**: `getColumnSignature` now accumulates the `strings.Builder` capacity during the type-switch loop that builds `entries`, eliminating the redundant second iteration over entries that was previously computing the size.

2. **Fallback signature for external callers**: Both `WriteTypedColumnarDirect` and `writeTypedColumnarInternal` now compute the signature on the spot if `typedColumns.Signature` is empty — preventing a spurious flush-on-every-write when callers construct `TypedColumnBatch` directly without going through `convertColumnsToTyped`.

xe-nvdk:
@gemini-code-assist Good catch — that was a genuine regression introduced during the single-pass refactor. Fixed in `6055616`.

The guard is now `len(name) == 0 || name[0] == '_'`, matching the original behavior exactly. Also added a dedicated test case (`skips_empty_column_names`) to lock this in and prevent it regressing again. Please do a final pass.

xe-nvdk:
@gemini-code-assist Both of your findings addressed in `e64daee`, plus additional issues I found during a local audit. Please do another pass.

**Your findings (fixed):**

1. **nil validity entry preserved** in `sortTypedColumnBatchByKeys` — per the `TypedColumnBatch` contract (nil = all valid), the reorder loop now skips nil entries with `continue`, eliminating the panic risk you flagged.

2. **stdlib constants**: replaced local `maxInt64`/`minInt64` with `math.MaxInt64` / `math.MinInt64`.

**Additional issues I found and fixed during a local audit:**

3. **Same nil-validity bug in `sliceTypedColumnBatchByIndices`** — the inner loop with `idx < validLen` defaulted to `false` (null) for nil entries, silently converting "all valid" into "all null" for sliced hour batches. Now preserves nil per contract.

4. **`mergeBatches` contract violation**: if `batchValidity[name]` was present but the slice was nil, `copy(dest, nil)` left `dest` all-false (null) instead of treating it as all-valid. Now handled alongside the missing-entry case.

5. **Inconsistent WAL error counting**: `writeTypedColumnarInternal` was missing `totalWALErrors.Add(1)` — two of three WAL append error sites were counted, this one wasn't. Now consistent.

6. **`sliceTypedColumnBatchByIndices` wasn't propagating `Signature`** — sliced batches share the same schema as their source, so the signature is identical. Now propagated.

7. **Duplicate doc comment** on `int64SliceToTimestamps` (stale original block left in place when the H3 lifetime comment was added).

8. **Stale release notes**: removed the `BatchToColumnar tag lookup reduction` bullet (that change was reverted in response to your earlier feedback). Expanded the column-signature bullet to mention type-aware detection.

**Tests added (all pass with `-race`):**
- `TestSortTypedColumnBatchByKeys_NilValidityEntry`
- `TestSliceTypedColumnBatchByIndices_NilValidityEntry`
- `TestLineProtocolParser_ParseTimestamp_Precisions` — 8 cases covering us/ms/s/ns, pre-epoch negatives, and overflow fallback

**Verified locally:**
- `go build ./cmd/... ./internal/...` — clean
- `go vet ./internal/... ./cmd/...` — clean
- `gofmt -l internal/ingest/` — clean
- `go test ./internal/ingest/... -race -count=1` — ok
- `go test ./internal/api/...` — ok
- `go test ./internal/wal/...` — ok

PR #413, opened 2026-04-27
fix(ingest): 26.05.1 critical hardening — 5 criticals + review-pass cleanup

xe-nvdk:
## Summary

Five critical bugs surfaced by a 4-agent post-implementation review of the ingest path; this PR fixes all five plus the consolidated findings from a second 4-agent review of the fix branch (real bugs introduced, plus refactor/test/comment cleanups). All five are 26.05.1 GA blockers; fixing them does not regress the headline number — sustained 60s MessagePack columnar benchmark on this branch shows **19.04M rec/s @ p99 3.13ms vs baseline 18.6M rec/s @ p99 3.68ms** (~17% p99 improvement, ~2% throughput uplift, 0 errors).

## Critical fixes

- **C1 — graceful-shutdown panic**: `Close()` closed `flushQueue` after `cancel()`, racing concurrent writers past `shard.mu.Unlock()`. Fixed: `closing atomic.Bool` flag set before cancel, channel never closed (workers exit on `b.ctx.Done()`).
- **C2 — schema-evolution corruption under concurrent writes**: race window inside `flushBufferLocked` could install a third schema mid-flight; original caller appended into the racer's buffer. Fixed: bounded loop (8 iterations) re-checks schema after I/O + per-iteration `ctx.Err()` check.
- **C3 — WAL silent backpressure drop**: `AppendRaw`/`AppendRawWithMeta` returned `nil` on full channel, making downstream "data preserved in WAL" log lines untrue. Fixed: typed `wal.ErrWALDropped` sentinel; ingest caller differentiates via `errors.Is` + sampled Warn + dedicated `total_wal_dropped` counter; cluster receivers treat as non-fatal (don't advance `lastSeq` to silent divergence).
- **C4 — gzip decompression bomb in LP/TLE handlers**: Fiber's `c.Body()` transparently gunzipped without a size cap. Fixed: `c.Request().Body()` (raw) + magic-byte detection + pooled bounded decompressor (mirrors msgpack pattern).
- **C5 — ingest endpoints missing write-tier auth**: read-only token could write data via `/api/v1/write/*` when RBAC was disabled (OSS default). Fixed: explicit `auth.RequireWrite`/`RequireAdmin` middleware on all 9 endpoints via new `internal/api/auth_middleware.go` helpers; `SetAuthAndRBAC` takes concrete `*auth.AuthManager` directly to eliminate silent type-assert misses.

## Review-pass refactors and cleanups

- Extract `Writer.tryEnqueue` (kills 2 copies of the WAL drop pattern).
- Extract `ArrowBuffer.tryEnqueueFlush` (kills 4 copies of the closing-flag + select-with-ctx-arm + queue-full pattern). Returns a `flushSendOutcome` enum so callers know `queued` / `closing` / `ctx-canceled` / `queue-full` distinctly.
- Extract `ArrowBuffer.flushOnSchemaChangeLocked` (kills 2 copies of the schema-evolution loop).
- Extract `withWriteAuth` / `withAdminAuth` + `passthroughMiddleware` (collapses if/else registration in 4 ingest handlers).

## Test plan

- [x] `go vet ./cmd/... ./internal/...` clean
- [x] `go test ./internal/ingest/... ./internal/wal/... ./internal/api/... ./internal/cluster/...` (all green, ~95s)
- [x] Race-detector pass on new tests: `go test -race -run "CloseRace|WriteAfterClose|SchemaEvolutionConcurrent|ErrWALDropped"` all green
- [x] Sustained 60s MessagePack columnar benchmark: 19.04M rec/s @ p99 3.13ms, 0 errors
- [x] New regression tests are deterministic (no `t.Skip` on fast machines, no fixed `time.Sleep` on slow ones)
- [x] Manual: hit `/api/v1/write/msgpack` with a read-only token, verify 403 (was 200)
- [x] Manual: post a 1MB gzip-bomb to `/write` with `Content-Encoding: gzip`, verify rejection (was OOM)

## Process notes

- Two rounds of 4-agent internal review (correctness / security / code-quality / perf) ran on the branch before opening this PR; all blocking findings from both rounds are addressed in this single commit.
- Three of the four reviewers in round 2 independently flagged the cluster-receiver fallout from the C3 fix — that's the kind of cross-cutting find the new CLAUDE.md review framework is designed to catch *before* shipping to gemini.


xe-nvdk:
/gemini review

@gemini-code-assist this is the 26.05.1 GA-blocker hardening PR. Five critical fixes plus the cleanup pass. Two rounds of 4-agent internal review on the branch caught real bugs in the first round of fixes (cluster-receiver fallout from C3, log-spam under WAL backpressure, type-assert silent miss in auth wiring) — all addressed before opening this PR. Looking for any remaining findings before merge.

Sustained-load benchmark on this branch:
- 19.04M rec/s, p99 3.13ms, 0 errors over 60s (vs baseline 18.6M @ p99 3.68ms)
- Hot record path is unchanged; all new logic is per-flush or per-shutdown

xe-nvdk:
/gemini review

@gemini-code-assist round 2 fixes pushed in commit 713bbd3. All four findings addressed:

**MEDIUM (×2) — error-detail restore:**
- ✅ LP gzip-decompress response now includes `err.Error()` — line 229.
- ✅ TLE gzip-decompress response now includes `err.Error()` — line 131.
- Same change applied symmetrically to the new zstd branches in both handlers.

**MEDIUM — zstd on LP/TLE (consistency with msgpack):**
- ✅ Both handlers now detect zstd via magic bytes (`0x28 0xB5 0x2F 0xFD`) after the gzip check, with the same 100 MB decompressed-size cap as gzip.
- New helper `decompressZstdPooled(data []byte, maxSize int) ([]byte, error)` in tle.go reuses the package-level `zstdDecoderPool` and `decompressBufferPool` already declared in msgpack.go. Returns a freshly-copied `[]byte` rather than a `PooledBuffer` — keeps the LP/TLE flat-byte handler shape unchanged. The pooled buffer itself is returned to its pool inside the helper, with the same oversized-buffer discard policy as msgpack (`cap > maxPooledBufferSize → fresh small buffer for the pool, decompBufferDiscards counter incremented`).
- New release-notes subsection "Line Protocol and TLE: zstd Decompression Support" describes the addition, the magic-byte detection, the 100 MB cap, and the 3–5× speedup over gzip relevant for satellite-fleet TLE senders.

**MEDIUM — release-notes prose clarity:**
- ✅ Replaced the shell-glob `/api/v1/import/{csv,parquet,lp,tle}` with the four explicit paths and rephrased the auth bullet for readability per your suggestion.

Build clean (`go build ./internal/api/... ./cmd/...`) and full api test suite passes (`go test ./internal/api/...` → `ok 5.848s`). No new tests added for the zstd helper because it shares the contract and pool with msgpack's already-tested path; if you'd like a dedicated `decompressZstdPooled` unit test (oversized-payload cap, malformed-magic-bytes-but-valid-stream, etc.), I'm happy to add it.

Looking for any final findings before merge.

xe-nvdk:
/gemini review

@gemini-code-assist round 3 fixes pushed in commit 3c79ea1. Both findings addressed:

**CRITICAL — body-aliasing footgun (LP and TLE):**
- ✅ Both handlers now copy the raw fasthttp body at `localProcessing` entry: `body := append([]byte(nil), c.Request().Body()...)`.
- Verified: the LP and TLE parsers already produce fresh strings (`string(...)` copies in `unescape` / `parseFieldValue`), so today's behavior was actually safe — but the no-aliasing contract was undocumented, brittle, and the silent-aliasing footgun would surface as silent data corruption if anyone ever added a fast-path that retained a sub-slice. The defensive copy is one memcpy per request, dwarfed by per-record allocations.
- The decompression branches already returned fresh slices, so they're symmetric. msgpack handler is unchanged: its `decoder.Decode(payload)` is synchronous and consumes `payload` before returning, so fasthttp body reuse is fine there.

**MEDIUM — schema-evolution cap behavior:**
- ✅ `flushOnSchemaChangeLocked` now returns a typed `ingest.ErrSchemaChurnExceeded` sentinel when `schemaEvolutionMaxIters` is reached (was: log + return nil, allowing wide schema-mixed buffers to land on disk).
- ✅ LP and TLE write handlers map `errors.Is(err, ingest.ErrSchemaChurnExceeded)` → HTTP 503 so upstream senders treat it as retryable backpressure rather than permanent server error.
- ✅ New `totalSchemaChurnExceeded` atomic counter surfaced in `ArrowBuffer.Stats()` as `total_schema_churn_exceeded` so operators can alert on a non-zero rate.
- Important: the per-iteration flushes inside the loop already wrote older schemas' rows to durable Parquet — only the *current request* fails. No data loss.

**New regression tests:**
- `TestArrowBuffer_SchemaChurnExceededErrorContract` — runs 16 writers × 200 iterations with 12 rotating distinct schemas against a single shard, asserts that (a) every non-nil write error must be `errors.Is(err, ErrSchemaChurnExceeded)` (catches a regression to "return nil") and (b) `totalSchemaChurnExceeded` counter delta == churn-error count (catches counter/error decoupling). Test passes deterministically; under steady-state runs the cap doesn't fire (0 errors), the contract still holds.
- `TestArrowBuffer_ErrSchemaChurnExceeded_IsExported` — pins the sentinel as exported and unwrappable via `errors.Is`.

Full ingest + api suites pass (`go test ./internal/ingest/... ./internal/api/... → ok 66s + 5s`); build clean.

Looking for any final findings before merge.

xe-nvdk:
/gemini review

@gemini-code-assist round 4 fixes pushed in commit 0822793. All four NEW findings from round 3 addressed (the 2 earlier-rounds re-flags at lineprotocol.go:231 / arrow_writer.go:1440 were stale — already addressed by 713bbd3 / 3c79ea1 respectively, just gemini re-flagging them on shifted line numbers):

**MEDIUM — defensive body copy now deferred (LP and TLE):**
- ✅ Both handlers switched from "unconditional copy + compression detection" to a single `switch` over magic bytes. Gzip and zstd branches return fresh slices directly (decompression already produces a fresh allocation); only the uncompressed `default` branch makes the defensive `append([]byte(nil), rawBody...)` copy.
- This eliminates the doubled-memory case for large compressed payloads while preserving the silent-aliasing footgun fix from round 2.

**MEDIUM — `ErrSchemaChurnExceeded` → 503 in msgpack handler:**
- ✅ `writeMsgPack` now mirrors LP/TLE — `errors.Is(err, ingest.ErrSchemaChurnExceeded)` maps to HTTP 503 with retryable-backpressure semantics.
- ✅ Generic-error path also restored `err.Error()` in the response body for client-side diagnosis (was previously a hard-coded "Failed to write records" with no detail). Now consistent with the gzip-decompress responses.

**MEDIUM — `flushOnSchemaChangeLocked` docstring stale:**
- ✅ Updated comment block to match implementation: "`schemaEvolutionMaxIters reached — return ErrSchemaChurnExceeded so the caller can reject the write with a retryable status (HTTP 503). The per-iteration flushes inside the loop already wrote older schemas' rows to durable Parquet, so there is no data loss — only the current request fails under sustained schema-rotation churn.`"

Build clean, full ingest + api suites pass (`go test ./internal/ingest/... ./internal/api/... → ok 67s + 5s`).

Looking for any final findings before merge.

xe-nvdk:
/gemini review

@gemini-code-assist round 5 fixes pushed in commit 1aca1e8. All three NEW findings from round 4 addressed (the 2 earlier-rounds re-flags at arrow_writer.go:1440 / msgpack.go:168 were stale — already fixed in 3c79ea1 and 0822793 respectively, gemini's grep just landed on shifted lines):

**🚨 HIGH/security-high — zstd decompression bomb in all three handlers:**
- ✅ Confirmed real vulnerability in BOTH the new LP/TLE zstd path (added 713bbd3) AND the pre-existing msgpack zstd path. `zstd.Decoder.DecodeAll` grows its output buffer to fit the entire decompressed stream regardless of `WithDecoderMaxMemory` (which bounds only the per-frame window, not output growth).
- ✅ All three switched to streaming `decoder.Reset(bytes.NewReader(data))` + `io.LimitReader(decoder, max+1)` + `io.ReadAll`. Bound is enforced *during* decoding so the bomb is rejected with a bounded `len(maxSize)+1` allocation. Same shape as the existing gzip path.
- ✅ Pool semantics preserved: decoder is still pooled (the per-frame state is the expensive part); output is now a fresh slice on the streaming path because `io.ReadAll`'s growth pattern fights `decompressBufferPool`'s fixed-capacity contract. msgpack's `PooledBuffer` returned with `bufPtr=nil` so `Release()` is a safe no-op for the new path.
- ✅ New regression tests in `internal/api/decompress_bomb_test.go`:
  - `TestDecompressZstdPooled_BombRejected_WithoutOOM` — 256 MB plaintext compresses to 28 KB (ratio 9361×); `decompressZstdPooled` correctly rejects with `"exceeds 100MB limit"`.
  - `TestDecompressZstdPooled_RejectsBenignOversizedPayload` — non-adversarial cap-exceeded path.
  - `TestDecompressZstdPooled_HappyPath_RoundTrip` — pin against regression in normal-sized requests.

**MEDIUM — extract shared `decompressRequest` helper:**
- ✅ New `decompressRequest(rawBody []byte, maxSize int) (body []byte, codec string, err error)` in tle.go. Returns the fresh-decoded slice plus a codec label (`"gzip"`, `"zstd"`, or `""`) for structured logging.
- ✅ LP and TLE handlers reduced from a 30-line inline switch each to a 3-line `body, codec, err := decompressRequest(rawBody, 0)` call. All three CRITICAL invariants (no fasthttp gunzip, streaming-bounded zstd, defensive copy on uncompressed) are preserved inside the helper.

**MEDIUM — remove duplicated gzip helper:**
- ✅ `LineProtocolHandler.decompressGzip` method (lineprotocol.go:373) was a near-clone of `decompressGzipPooled` (tle.go) and pulled from the same `lpGzipReaderPool`. Method removed; all routes go through `decompressRequest` → `decompressGzipPooled`. Removed now-unused `bytes`/`io`/`gzip` imports from lineprotocol.go.

Build clean, full suites pass (`go test ./internal/api/... → 2.886s, ./internal/ingest/... → 66.912s`); 3 new bomb-regression tests included.

Looking for any final findings before merge.

xe-nvdk:
/gemini review

@gemini-code-assist round 6 fixes pushed in commit 4a05cfb. All 5 NEW round-5 findings addressed (the recurring arrow_writer.go:1440 and msgpack.go:168 re-flags remain stale — already fixed in 3c79ea1 and 0822793; gemini's grep just keeps landing on shifted lines):

**MEDIUM × 4 — sampled Warn for follower-side WAL backpressure:**
- ✅ Added `walDropLastLogNano atomic.Int64` field to both `replication.Receiver` and `sharding.ShardReceiver`.
- ✅ Added `walDropLogIntervalNano = int64(time.Second)` const next to each receiver type.
- ✅ Wrapped both Warn emissions in the same atomic-CAS sampling guard used by `ArrowBuffer.recordWALError` — at most one log line per second per receiver, suppression keyed on `(now - last) < interval && CompareAndSwap(last, now)`.
- ✅ Both Warn payloads now also emit the running `total_dropped` counter so operators can read the rate from a single line if they grep for it.
- ✅ The non-fatal/applied semantics from the 26.05.1 fix are unchanged — only the log shape changed.

**MEDIUM — formatBytes consistency:**
- ✅ `decompressZstdPooled` and `decompressGzipPooled` cap-error messages now use `formatBytes(int64(maxSize))` instead of `%dMB`. Matches msgpack's `decompressZstd` / `decompressGzip` exactly.

Build clean, full suites pass: `go test ./internal/api/... ./internal/cluster/replication/... ./internal/cluster/sharding/... → ok 4.7s + 0.8s + 3.0s`.

Looking for any final findings before merge.

xe-nvdk:
/gemini review

@gemini-code-assist round 7 fixes pushed in commit 6561c86. Both NEW HIGH findings addressed; the 6 recurring re-flags (arrow_writer.go:1440, msgpack.go:168, replication/receiver.go:91+386, sharding/shard_receiver.go:81+494) remain stale — those code shapes were already fixed in 3c79ea1 / 0822793 / 4a05cfb and your grep just keeps landing on shifted lines.

**🚨 HIGH/security-high — `decompressRequest` uncompressed branch missing maxSize cap:**
- ✅ Helper now defaults `maxSize <= 0` to 100MB up front and applies the cap to all three branches symmetrically. The uncompressed path rejects over-cap input *before* the defensive copy, closing the OOM vector you flagged.
- ✅ New regression tests:
  - `TestDecompressRequest_UncompressedRejectsOverCap` — `cap+1` raw bytes → "exceeds 100MB limit" error, codec=`""`.
  - `TestDecompressRequest_UncompressedAcceptsAtCap` — exactly `cap` bytes accepted; mutates `body[0]` and verifies `rawBody[0]` is unchanged, locking in the defensive-copy no-aliasing invariant.

**🚨 HIGH — msgpack handler missing defensive body copy:**
- ✅ `writeMsgPack` now routes through the shared `decompressRequest(rawBody, int(h.maxPayloadSize))` helper, just like LP and TLE. The defensive copy + new uncompressed-cap apply automatically.
- ✅ Deleted `MsgPackHandler.decompressGzip` and `decompressZstd` (240 lines of per-handler pooled-buffer codecs replaced by the package-level streaming helpers in tle.go). `PooledBuffer` itself remains for `TestPooledBufferRelease`.
- ✅ Removed unused `bytes`/`io`/`gzip`/`zstd` imports from msgpack.go.
- ✅ Error message shape unified: when codec is empty (uncompressed-rejection), the response is now `"Failed to read request body: <err>"` instead of the awkward `"Failed to decompress  data: <err>"`. When codec is `gzip`/`zstd`, the existing decoder-error message format is preserved.

**Verification of the round-2 conclusion (no current data corruption):**
- The Basekick-Labs/msgpack/v6 fork was inspected: string fields go through `string(b)` (always allocates), and `[]byte` fields go through `io.ReadFull` into a fresh slice (also always allocates). So today's msgpack handler was technically safe even without the defensive copy. But the contract was undocumented and brittle, and the inconsistency with LP/TLE was a real maintenance hazard — your finding to defensively copy is the right call regardless.

Build clean, full suites pass: `go test ./internal/api/... ./internal/cluster/replication/... ./internal/cluster/sharding/... → ok 4.8s + 0.9s + 2.5s`. 5 decompress regression tests (3 round-4 bomb + 2 round-6 cap/aliasing) pass deterministically.

Looking for any final findings before merge.

xe-nvdk:
/gemini review

@gemini-code-assist round 8 fixes pushed in commit 04e64cc. One NEW round-7 finding addressed plus a perf restoration:

**MEDIUM — drop poisoned zstd decoder on partial-decode error:**
- ✅ `decompressZstdPooled` now calls `decoder.Close()` on the io.ReadAll error path instead of `Reset(nil)` + returning to the pool. After a mid-stream corruption a poisoned internal frame state is not safely recoverable; closing the decoder lets the next request's `pool.Get()` allocate a fresh one. Cheap and defensive.

**Perf restoration — msgpack uncompressed fast-path:**
- After round-6 unified all three handlers behind the shared `decompressRequest` helper (which copies on the uncompressed branch), the 60-second sustained MessagePack-columnar bench dropped from **19.04M rec/s** (baseline) → **18.04M rec/s** (~5% regression). At 18M rec/s × ~50KB batches, the new per-request defensive memcpy becomes a real allocation storm.
- msgpack now inlines its codec dispatch: gzip/zstd still go through the shared helpers (those need the bomb cap and produce fresh slices anyway), but the uncompressed branch hands `rawBody` directly to `h.decoder.Decode`. The decoder is synchronous and the Basekick-Labs/msgpack/v6 fork copies every string via `string(b)` and every `[]byte` via `io.ReadFull` — so by the time `Decode` returns, no decoded value still aliases `rawBody`.
- LP and TLE keep the defensive copy in the shared helper — their throughput targets don't justify diverging.

**New regression test** — `TestMessagePackDecoder_CopiesStringFields_NoBodyAliasing` decodes a row-format payload, mutates the source bytes to `0xFF`, and asserts the decoded `Measurement` and `Tags["region"]` values are unchanged. **Locks in the no-aliasing invariant**: if a future fork update introduces zero-copy aliasing, the test fails loudly before production sees corruption. This is the load-bearing safety net for the bypass-the-copy decision.

**Perf result with this commit:** 19.13M rec/s @ p99 3.06ms over 60s — slightly above the post-round-1 baseline (19.04M / p99 3.13ms).

**Bumped** `Basekick-Labs/msgpack/v6` from v6.0.0 to v6.1.0 (user-released; verified the no-aliasing regression test still passes).

The 7 other comments in round 7 are stale re-flag noise on code shapes already fixed in 3c79ea1 / 0822793 / 4a05cfb / 6561c86.

Build clean, full suites pass, no-aliasing invariant locked in test.

Looking for any final findings before merge.

PR #416, opened 2026-05-03
fix(mqtt): nil-guard subscriber map and disabled manager

### LINKED ISSUE #304: high(mqtt): Nil pointer panics when MQTT disabled + nil subscriber in GetAllStats
mqtt.go:38, manager.go:475 — handlers dont check nil. Fix: add nil guards.

SAY-5:
Closes #304.

Two surfaces:
- `GetAllStats` only checked map presence, so a nil entry (mid-shutdown or when the subscriber failed to start) panicked on `subscriber.GetStats()`. Mirror the existing single-id `GetStats` check (`!ok || subscriber == nil`).
- `MQTTHandler.handleStats`/`handleHealth` deref the manager unconditionally, so the API endpoints crash when MQTT is disabled and the handler was wired with a nil manager. Return 503 JSON instead.

### DIFF
--- a/internal/api/mqtt.go
+++ b/internal/api/mqtt.go
@@ -97,6 +97,12 @@ func (h *MQTTHandler) handleHealth(c *fiber.Ctx) error {
 		return c.Status(fiber.StatusServiceUnavailable).JSON(fiber.Map{
 			"status":  "disabled",
 			"healthy": false,
+			"subscriptions": fiber.Map{
+				"total":   0,
+				"running": 0,
+				"stopped": 0,
+				"errors":  0,
+			},
 		})
 	}
 	stats, err := h.manager.GetAllStats(c.Context())
--- a/internal/api/mqtt.go
+++ b/internal/api/mqtt.go
@@ -40,6 +40,12 @@ func (h *MQTTHandler) RegisterRoutes(app *fiber.App) {
 
 // handleStats returns statistics for all MQTT subscriptions
 func (h *MQTTHandler) handleStats(c *fiber.Ctx) error {
+	if h.manager == nil {
+		return c.Status(fiber.StatusServiceUnavailable).JSON(fiber.Map{
+			"success": false,
+			"error":   "MQTT subsystem disabled",
+		})
+	}
 	stats, err := h.manager.GetAllStats(c.Context())
 	if err != nil {
 		h.logger.Error().Err(err).Msg("Failed to get MQTT stats")
@@ -87,6 +93,12 @@ func (h *MQTTHandler) handleStats(c *fiber.Ctx) error {
 
 // handleHealth returns MQTT subsystem health status
 func (h *MQTTHandler) handleHealth(c *fiber.Ctx) error {
+	if h.manager == nil {
+		return c.Status(fiber.StatusServiceUnavailable).JSON(fiber.Map{
+			"status":  "disabled",
+			"healthy": false,
+		})
+	}
 	stats, err := h.manager.GetAllStats(c.Context())
 	if err != nil {
 		return c.Status(fiber.StatusServiceUnavailable).JSON(fiber.Map{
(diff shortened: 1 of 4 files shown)


xe-nvdk:
Hey @SAY-5 thank you for the PR, can you review the suggestion from Gemini please?

xe-nvdk (changes requested):
Thanks for closing this out @SAY-5 — the fix is on the right shape and the diff is appropriately small. A few things before we merge:

## Looks good

- `GetAllStats` nil-subscriber guard at `internal/mqtt/manager.go:475` correctly mirrors the existing single-id `GetStats` pattern at `internal/mqtt/manager.go:443` (`!ok || subscriber == nil`). Consistent and minimal.
- `if h.manager == nil` in the API handlers is safe here. I checked the wiring at `cmd/arc/main.go:457` — when MQTT is disabled, `var mqttManager mqtt.Manager` is left as a true nil interface (no concrete type ever assigned), so the comparison works. No typed-nil hazard.
- 503 with a structured body for both `/stats` and `/health` is clearly distinguishable from genuine errors by clients.

## Please address before merge

### 1. Add regression tests

Both branches added here are easy to unit-test and currently uncovered:

- `MQTTHandler.handleStats` / `handleHealth` with `manager: nil` should return 503 with the documented body shapes. Use the standard Fiber `app.Test()` pattern.
- `SubscriptionManager.GetAllStats` with a `subscribers` map that contains a `nil` entry for a known ID should return the DB-fallback `SubscriptionStats` for that ID (not panic). A small fake repo is enough.

Without tests, the next contributor refactoring `MQTTHandler` won't realize this nil branch is load-bearing and may delete it.

### 2. `/health` disabled-response: 503 or 200?

Returning `503 Service Unavailable` + `"healthy": false` for a *configured-off* subsystem reads as "something is broken" to every uptime-check in the world. Two reasonable positions:

- **Keep 503**: loud-by-design — an operator who *thinks* MQTT is enabled gets paged. Defensible since the endpoint is admin-only. If you go this way, please add a one-line comment explaining the choice.
- **Switch to 200 + `"status": "disabled"`**: more semantically correct — disabled is a steady state, not a degraded one.

Either is fine, but please pick one explicitly.

## Non-blocking observations

### Inconsistent nil-handling philosophy across the two MQTT handlers

The comment at `cmd/arc/main.go:1518` says *"always register, handlers check if manager is nil"* — that's now true for `MQTTHandler` after this PR, but `MQTTSubscriptionHandler` next to it is gated at wiring time (`cmd/arc/main.go:1523`) and contains 11 unguarded `h.manager.X` call sites. Both strategies work; we just have one of each now. Worth either updating the comment to reflect the actual split, or picking one strategy across both handlers in a follow-up. Not blocking this PR.

### Minor: redundant body in disabled `/health`

`"subscriptions": {total:0, running:0, stopped:0, errors:0}` is derivable from `"status": "disabled"` and adds nothing for a client. Fine to keep for response-shape consistency, but trimming it cleans up the response. Take it or leave it.

---

Approving once tests are added and the `/health` 503-vs-200 choice is documented. Once green I'll tag `gemini-code-assist[bot]` for the second pass.

xe-nvdk (approved):
Thank you @SAY-5, this is a clean follow-up. Both review points addressed precisely:

- The 200 + `"status": "disabled"` choice on `/health` is the right one and the inline comment explaining *why* (so it doesn't get "fixed" back to 503 by the next contributor) is exactly the kind of small thing that pays off later.
- The integration-style test for `GetAllStats_NilSubscriber` (real SQLite repo + manually-injected nil subscriber under the mutex) is more valuable than a mocked version would have been. It locks in both the nil-guard *and* the DB-fallback branch.

I checked the PR out locally and ran the tests:

```
ok  github.com/basekick-labs/arc/internal/api   0.912s
ok  github.com/basekick-labs/arc/internal/mqtt  0.333s
```

Both new tests pass. Full package suites pass. Build is clean.

Approving and merging. Genuinely appreciate the careful work on this one — closing out #304 with proper regression coverage is exactly how this stuff should land.

PR #418, opened 2026-05-07
fix(mqtt): unify nil-guard policy across both MQTT API handlers

xe-nvdk:
## Summary

Follow-up to #416 (which added handler-side nil-guards to `MQTTHandler.handleStats` / `handleHealth`). After that PR, the two MQTT handlers had different disabled-MQTT semantics:

- `MQTTHandler` (stats/health) returned **503** with a `"MQTT subsystem disabled"` body when manager was nil.
- `MQTTSubscriptionHandler` (CRUD/lifecycle) was gated at wiring time, so when MQTT was disabled the routes **404**'d.

That split was already noted in the #416 review thread. This PR harmonizes the two so all MQTT API routes return a consistent 503 disabled-response shape, which is friendlier for monitors and ops dashboards that probe across the full surface.

## Approach

- New `requireEnabled(c)` helper on `MQTTSubscriptionHandler` returning `(handled bool, err error)`. The two-return form is needed because Fiber response writes return `nil` on success; a single-error return can't distinguish "I wrote the 503, stop" from "MQTT is enabled, proceed".
- Each of the 10 handlers (`handleCreate`, `handleList`, `handleGet`, `handleUpdate`, `handleDelete`, `handleStart`, `handleStop`, `handlePause`, `handleRestart`, `handleStats`) gains a 3-line guard at the top:
  ```go
  if handled, err := h.requireEnabled(c); handled {
      return err
  }
  ```
- `cmd/arc/main.go`: drop the wiring-side `if mqttManager != nil` gate; comment updated to explain the now-unified policy and point at the regression tests.
- `internal/api/mqtt_subscriptions_test.go`: new table-driven regression test covering 6 representative routes (CRUD + lifecycle + stats) with a nil manager. Asserts 503 status + documented body shape.

## Why this is the right shape

The alternative (gate at wiring time everywhere) would 404 on disabled MQTT, which silently breaks any client/dashboard that probes `/api/v1/mqtt/*` to discover whether the subsystem exists. 503 with a structured `"MQTT subsystem disabled"` body is unambiguous and uniform.

The duplicated 3-line guard at the top of each handler is the cost of doing this with handler-level nil-guards. It's one helper call and consciously chosen over middleware so the policy is visible in each handler.

## README

Adds @SAY-5 (Sai Asish Y) to the contributor list for #416.

## Test plan

- [x] `go test -count=1 ./internal/api/... ./internal/mqtt/...` passes (including the existing #416 tests and the new `TestMQTTSubscriptionHandler_DisabledManager` with 6 sub-tests)
- [x] Manual verification: helper signature pattern matches Fiber idiom (handled-bool + error)
- [ ] Review by `gemini-code-assist[bot]`

### DIFF
--- a/README.md
+++ b/README.md
@@ -345,7 +345,11 @@ For commercial licensing, contact: **enterprise@basekick.net**
 
 ## Contributors
 
-Thanks to everyone who has contributed to Arc:
+Thanks to everyone who has contributed code to Arc:
 
-- [@schotime](https://github.com/schotime) (Adam Schroder) - Data-time partitioning, compaction API triggers, UTC fixes
-- [@khalid244](https://github.com/khalid244) - S3 partition pruning improvements, multi-line SQL query support
+- [@schotime](https://github.com/schotime) (Adam Schroder) — Data-time partitioning, compaction API triggers, UTC fixes
+- [@SAY-5](https://github.com/SAY-5) (Sai Asish Y) — MQTT nil-guard hardening (handlers + manager) with regression coverage
+
+And a thank-you to community members whose bug reports drove fixes:
+
+- [@bjarneksat](https://github.com/bjarneksat) — reported the line-protocol null-field handling bug fixed in 26.03.1
--- a/README.md
+++ b/README.md
@@ -348,6 +348,7 @@ For commercial licensing, contact: **enterprise@basekick.net**
 Thanks to everyone who has contributed code to Arc:
 
 - [@schotime](https://github.com/schotime) (Adam Schroder) — Data-time partitioning, compaction API triggers, UTC fixes
+- [@khalid244](https://github.com/khalid244) — S3 partition pruning improvements, multi-line SQL query support
 - [@SAY-5](https://github.com/SAY-5) (Sai Asish Y) — MQTT nil-guard hardening (handlers + manager) with regression coverage
 
 And a thank-you to community members whose bug reports drove fixes:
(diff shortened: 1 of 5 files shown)


xe-nvdk:
Thanks for the review @gemini-code-assist — addressed all four findings in commit 48a9f64:

**1. (High) `@khalid244` removed from README** — restored. Their historical contribution entry stays as-is.

**2. (Medium) `main.go` comment claimed all routes return 503** — fixed. The comment now explicitly notes that `MQTTHandler.handleHealth` returns 200 with `{"status":"disabled","healthy":false}` to avoid paging operators about a configured-off subsystem, while every other MQTT route returns 503.

**3. (Medium) `requireEnabled` should be Fiber middleware** — refactored. `requireEnabled` now returns `error` and calls `c.Next()` on the enabled path; it's applied via `subs.Use(h.requireEnabled)` in `RegisterRoutes`. All 10 per-handler guard blocks dropped — net `-21` lines in `mqtt_subscriptions.go`. Future routes added to the group are auto-protected without any per-handler boilerplate. Good call.

**4. (Medium) Test resource leaks + ignored `io.ReadAll` errors** — fixed in both `mqtt_subscriptions_test.go` and the existing `mqtt_test.go` (PR #416 had the same pattern; same hygiene applied so they don't drift). `defer resp.Body.Close()` + `if err := io.ReadAll(...)` checks added.

Tests: 9 sub-tests pass (`TestMQTTHandler_DisabledManager` x2, `TestMQTTSubscriptionHandler_DisabledManager` x6, `TestSubscriptionManager_GetAllStats_NilSubscriber`). Build clean.

/gemini review

xe-nvdk:
Good catch @gemini-code-assist — fixed in 2e92b6c. The doc comment was carrying over the old per-handler reasoning even though the implementation moved to group-level middleware. Updated to describe the actual protection: `subs.Use(h.requireEnabled)` auto-guards every route in the group, so the "future contributor forgets to call the helper" failure mode no longer exists.

PR #419, opened 2026-05-07
feat(cluster): hard query gating during replication catch-up (#392)

### LINKED ISSUE #392: Hard query gating during catch-up
## Problem
Readers that are still catching up on file replication serve queries with potentially incomplete results. Operators must use external readiness probes via `/api/v1/cluster/status` to gate traffic.

## Proposed Solution
Wire `Coordinator.ReplicationReady()` (already landed in Phase 3) into the query path. When `ReplicationReady()` returns false, reject queries with a clear error. Add a config toggle (`cluster.query_gate_on_catchup`) so operators can opt in.

## Context
`ReplicationReady()` exists at `coordinator.go:1957`. Phase 3 release notes explicitly deferred this.

## Effort
Small

xe-nvdk:
Closes #392.

## Summary

Reader nodes in a clustered Arc Enterprise deployment previously accepted queries the moment they started, even while peer file replication was still pulling Parquet files the rest of the cluster already had. The Raft manifest knew about the missing files; the local storage didn't have them yet; `read_parquet()` globbed against local storage rather than the manifest. The result was **silent partial results** during the catch-up window.

26.05.1 closed half the gap (WAL replication makes unflushed writer data queryable on readers within milliseconds), but flushed Parquet files still depended on async background pullers.

This PR closes the remaining gap behind a single config flag.

## Behavior

- **`cluster.query_gate_on_catchup`** (default `false`) — when true, all user-facing read endpoints (`/api/v1/query`, `/api/v1/query/arrow`, `/api/v1/query/estimate`, `/api/v1/query/:measurement`, `/api/v1/measurements`) return `503 Service Unavailable` until peer file replication has fully converged.
- **OSS / standalone unaffected**: no puller → always ready → gate is a no-op.
- **`/api/v1/internal/cache/invalidate` deliberately excluded**: peer nodes need it to fire *during* catch-up.

## Readiness predicate (the load-bearing piece)

Phase 3 plumbed `Coordinator.ReplicationReady()` pointed at `Puller.CatchUpCompleted()`, which only signaled walker-done. Wiring it as-is would have let through exactly the silent-partial-results condition the gate exists to prevent.

This PR introduces `Puller.FullyCaughtUp()` which requires:

1. `catchupCompletedAt > 0` (walker finished)
2. `inflightCount == 0` (every queued/processing pull settled — `len(queue)` is implied since every queued entry was added to inflight before being sent)
3. `totalFailed == 0` (no pull gave up after retries)
4. `totalDropped == 0` (no entry was dropped due to queue saturation)

Failed and dropped pulls indicate files the manifest promised but this reader does not have. Both surface in the 503 body so operators see the deception if it happens.

## 503 response shape

```json
{
  \"success\": false,
  \"error\": \"replication_catch_up_in_progress\",
  \"message\": \"Reader is still catching up on replicated files. Retry shortly or check /api/v1/cluster for catch-up progress.\",
  \"catchup_status\": {
    \"queue_depth\": 7,
    \"inflight_count\": 2,
    \"failed\": 0,
    \"dropped\": 0,
    \"completed_at\": 0
  }
}
```

Plus `Retry-After: 5` header so HTTP-aware load balancers and clients back off automatically.

## Performance

`Puller.inflightCount` is a new `atomic.Int64` mirror of `len(inflight)`, updated under `inflightMu` in the same critical section as the map. Hot-path readers (gate middleware, `/api/v1/cluster`) are lock-free and don't contend with puller workers under sustained 503 storms.

`Puller.CatchUpStatus()` returns the 11 fields the gate body and operator dashboards need, built directly rather than projecting `Stats()` — saves one map allocation per gated request.

## Observability

- `QueryHandler.QueryGate503Total()` exposes cumulative gated-503 count for Prometheus / metrics dashboards.
- Sampled (1Hz) Warn log when the gate fires, with the gate counter and request path. Avoids flooding under sustained catch-up.
- `/api/v1/cluster` status endpoint exposes new `queue_depth`, `inflight_count`, `failed`, `dropped` fields under `replication_catchup_status`.

## Known limitation

Sub-millisecond window between `applyRegisterFile` committing a manifest entry to the Raft FSM and the `onRegister` callback firing `puller.Enqueue`. A query landing in that window can observe `ReplicationReady() == true` while a manifest entry the same Raft commit produced is not yet in the in-flight set. Closing this gap requires a per-query Raft `LastApplied()` barrier on the query path, which is out of scope for this opt-in gate. Documented in `ReplicationReady()`'s doc comment and the release notes; tracked follow-up.

## Internal review

Per [CLAUDE.md's post-implementation review policy](../blob/main/.claude/CLAUDE.md), four parallel staff/principal-engineer review agents (correctness, security, Go idioms, performance/observability) ran on the initial draft. They flagged **2 CRITICAL** + several HIGH/MEDIUM findings; **all addressed in the same commit before push** so gemini lands on a near-clean diff:

| Severity | Finding | Status |
|---|---|---|
| CRITICAL | Failed/dropped pulls leaving gate green | Fixed (predicate now checks `totalFailed == 0 && totalDropped == 0`) |
| CRITICAL | Raft → Enqueue race | Documented as known limitation (per-query Raft barrier out of scope) |
| HIGH | Redundant `len(queue)` check | Dropped (covered by `inflightCount == 0`) |
| HIGH | `inflightMu` contention | Atomic counter added |
| HIGH | `SetCluster` typed-nil branch | Comment clarified |
| MEDIUM | Dedicated `CatchUpStatus()` method | Added |
| MEDIUM | Gate-fire counter + sampled Warn | Added |
| MEDIUM | `Retry-After` header | Added |
| MEDIUM | Interface name `replicationReadinessProvider` | Renamed to `replicationReadiness` |
| MEDIUM | Shutdown TOCTOU on `c.puller` | Snapshot before nil-check |

## Test plan

- [x] 8 puller sub-tests covering `FullyCaughtUp` semantics + `inflightCount` invariant
- [x] 6 API sub-tests covering middleware behavior, body shape, `Retry-After` header, counter monotonicity
- [x] Full regression: `go test -count=1 ./internal/api/... ./internal/cluster/... ./internal/config/...` green
- [x] `go vet ./cmd/... ./internal/...` clean
- [ ] Manual end-to-end on a 2-node local cluster (writer + reader) per the plan in [docs/progress/](../tree/main/docs/progress) — recommend doing this before flipping the default. Not blocking PR.
- [ ] `gemini-code-assist[bot]` second-pass review


xe-nvdk:
Thanks for the second-pass review @gemini-code-assist. All three findings addressed in commit 605dee3 — the deepest one (cumulative-counter and inflight-anything stickiness) genuinely changed the architecture of the gate, so worth walking through:

## Finding #1 (HIGH) — `replication_catchup_enabled=false` + `query_gate_on_catchup=true` → permanent 503

**Fixed.** `cmd/arc/main.go` now validates the combination at startup. If both are set, the gate is auto-disabled with a clear `WARN` log line instead of locking the node into permanent 503. The operator's intent in disabling the walker is clear (they don't want bootstrap reconciliation), so a gate that depends on the walker should be a no-op too. Refusing to start would have been the more strict choice but is more disruptive; auto-disable + WARN matches the project's "fail observably, don't brick the deploy" pattern.

## Finding #2 (MEDIUM) — cumulative `totalFailed` / `totalDropped` make the gate permanently red on transient failures

**Fixed.** New catch-up-scoped counters: `Puller.catchupFailed` and `Puller.catchupDropped`. Bumped only when:
- A catch-up-tagged path's pull gives up after retries (in `processEntry`'s defer), or
- The walker drops an entry due to queue saturation (in `RunCatchUp`).

Steady-state pull failures hit `totalFailed` but **not** `catchupFailed`, so the gate doesn't stick on transient blips after the catch-up batch settles. Pinned by `TestPuller_FullyCaughtUp_IgnoresSteadyStateFailures`.

## Finding #3 (MEDIUM) — gating on any `inflightCount != 0` 503s every new flush

This was the deepest and most important one. **Fixed by scoping the inflight check to the catch-up batch:**

- New `Puller.catchupPaths` set tracks paths the walker specifically enqueued. `RunCatchUp` calls `markCatchUp` on each successful Enqueue.
- New `Puller.catchupInflight atomic.Int64` mirrors `len(catchupPaths)` for lock-free reads.
- `FullyCaughtUp` now checks `catchupInflight == 0`, not `inflightCount == 0`. Steady-state ingest's reactive Enqueues never enter `catchupPaths`, so they don't fire the gate.

The conceptual fix: the gate's contract is *"the reader has finished bootstrapping its view of the manifest as of startup,"* not *"no pulls are happening anywhere right now."* The latter would mean a 503 every few seconds on any healthy cluster — defeating the entire purpose of the feature. Pinned by `TestPuller_FullyCaughtUp_IgnoresSteadyStateInflight`.

## Finding #4 (MEDIUM) — `SetCluster` called after `RegisterRoutes` in main.go

**Resolved with a doc clarification rather than a reorder.** Looking at the actual call sequence: `RegisterRoutes` registers handlers into the Fiber app, but the HTTP server doesn't call `server.Start()` until line 1697 — well after `SetCluster` at line 1172. Field reads happen at request time, not registration time. No data race exists at startup. Updated the doc comment on `SetCluster` to reflect actual behavior: initialization-time-before-server-start is safe; runtime re-wiring is not supported. Reordering would also have required reordering `SetRouter` and several other `SetX` methods, which all share the same pattern, so the doc fix is the lower-blast-radius change.

## Bonus — flaky test fix

While running the full regression suite I caught `TestSenderBroadcastToMultipleReaders` failing 1/10 in package context. Root cause: `broadcastEntry` iterates readers in non-deterministic map order. The test drained the two conns serially, which could head-of-line-block whichever reader the broadcast picked second, tripping the per-reader `WriteTimeout` (1s) → `RemoveReader` → `conn.Close` → EOF on the test's subsequent read. Fixed by draining both conns concurrently in the test. **Stress-tested 20x, 0 failures, race-clean.** Not strictly required for #392 but caught while verifying my changes didn't introduce flakiness.

## Updated docs

- Release notes: [`RELEASE_NOTES_2026.06.1.md`](https://github.com/Basekick-Labs/arc/blob/release/26.06.1/RELEASE_NOTES_2026.06.1.md) (will land on the release branch when 26.06.1 is cut)
- Public docs: [docs.basekick.net update](https://github.com/Basekick-Labs/docs.basekick.net/commit/bbcc1e3) — corrected the "Query Gating During Replication Catch-Up" section with the new scoped semantics + the `:::warning` admonition for the catchup-disabled conflict.

## Tests

- 8 puller sub-tests for `FullyCaughtUp` semantics, including the two new ones pinning steady-state behavior.
- 6 API sub-tests for the gate middleware — unchanged from previous commit, still green.
- Full regression on `internal/api`, `internal/cluster/...`, `internal/config` is green.

/gemini review

xe-nvdk:
Both findings addressed in commit 9ff6498. Walking through:

## HIGH — global-counter race in `processEntry`'s defer

Took your suggested fix verbatim. Replaced `totalFailed.Load() > failedAtEntry` with a local `failed bool` set only when THIS worker hits the give-up path. Worker A's defer can no longer falsely attribute worker B's steady-state failure to a catch-up path. New test `TestPuller_CatchUpFailure_ConcurrentWorkerIsolation` pins the contract: bumping `totalFailed` outside `recordCatchUpFailure` must not move the gate counter.

## MEDIUM — sticky catch-up failures (doc claimed self-heal, code didn't)

You were right — the doc was lying. Implemented proper self-heal:

- New `catchupFailedPaths map[string]struct{}` tracks paths that gave up after retries.
- `recordCatchUpFailure(path)` — adds to set + increments `catchupFailed`. Idempotent.
- `clearCatchUpFailure(path)` — removes from set + decrements `catchupFailed`. No-op if path wasn't failed.
- `processEntry`'s success path (both `pulledFromPeer` and the skipped-local pre-pull check) sets a local `succeeded` bool. The defer calls `clearCatchUpFailure` on success unconditionally — safe because the helper is a no-op when the path wasn't in the failed set.

Net result: when a transient peer outage clears and a reactive FSM callback re-enqueues the same path, the worker's success path decrements `catchupFailed`, the gate re-opens automatically, no restart needed. Pinned by `TestPuller_CatchUpFailure_SelfHeals`.

`catchupDropped` deliberately does NOT self-heal — a drop means no inflight slot was ever taken, so there's no "the original drop's path just succeeded" event to attribute to. Recovery for drops requires a node restart (re-runs walker) or operator action. The doc comment explains this asymmetry; it's also called out in the release notes and the public docs page.

## Tests added

- `TestPuller_CatchUpFailure_SelfHeals` — record failure, clear, verify decrement and `FullyCaughtUp` flips back to true.
- `TestPuller_CatchUpFailure_RecordIsIdempotent` — duplicate record + over-clear don't drift the counter.
- `TestPuller_CatchUpFailure_ConcurrentWorkerIsolation` — global `totalFailed` motion from a non-catch-up worker does not move `catchupFailed`.

Plus the existing 8 gate tests still green. Full regression on `internal/api`, `internal/cluster/...`, `internal/config` is clean.

## Docs updated

- Release notes: describe local-`failed`-bool race fix + self-heal recovery.
- Public docs: [`docs.basekick.net` commit `0f219e0`](https://github.com/Basekick-Labs/docs.basekick.net/commit/0f219e0) — added a "Self-heal" paragraph to the gate section, explaining the asymmetry between failures (heal) and drops (don't).

/gemini review

xe-nvdk:
All three findings addressed in commit 74b5ceb. Walking through:

## CRITICAL — `markCatchUp` race in `RunCatchUp`

You caught a real bug. The post-`Enqueue` mark could race a fast worker that completed the pull and ran `inflightRemove` before `markCatchUp` did its work. The stale tag would then leak `catchupInflight` upward forever and brick the gate.

Fixed exactly as suggested:

- `markCatchUp` now runs **before** `Enqueue`. By the time a worker can reach `inflightRemove`, the tag is already in `catchupPaths` and `inflightRemove` cleans it up correctly.
- `markCatchUp` now returns `bool` (true if the tag was newly added). `RunCatchUp` uses the return value plus its drop-detection switch to compensate when `Enqueue` returns "dropped" — no inflight slot is taken in that case, no worker will run `inflightRemove`, so the walker calls a new `unmarkCatchUp` to clear the tag and decrement `catchupInflight`.
- The `isSelf` skip is now handled manually in `RunCatchUp` before any tagging, so origin==self entries never enter `catchupPaths`.

Pinned by `TestPuller_MarkCatchUp_DropCompensation`.

## MEDIUM — `catchupDropped` didn't self-heal

You were right that the asymmetry was a footgun. A single queue saturation during the walker pass would permanently 503 the reader until restart — same operational shape as the original sticky-failure problem.

Fixed by mirroring the failure-path self-heal:

- New `catchupDroppedPaths map[string]struct{}` tracks the affected paths.
- `recordCatchUpDrop(path)` — adds to set + increments `catchupDropped`. Idempotent.
- `clearCatchUpDrop(path)` — removes + decrements. No-op if path wasn't dropped.
- `processEntry`'s success path now calls **both** `clearCatchUpFailure` AND `clearCatchUpDrop` unconditionally (both are no-ops when the path isn't in the corresponding set). Any successful pull heals whichever flavor of catch-up gap originally affected the path.

Pinned by `TestPuller_CatchUpDrop_SelfHeals` + `_RecordIsIdempotent`.

## MEDIUM — `CatchUpStatus` broke API keys

You're right that this was a breaking change for `/api/v1/cluster/status` consumers, and dropping the cumulative counters was an observability regression. Took your suggested code shape almost verbatim:

- **Original keys preserved** (`failed`, `dropped`, `skipped_dup`) — they now carry catch-up-scoped values, so pre-#392 dashboards naturally start showing gate-relevant numbers without code changes.
- **New explicit `catchup_*` keys** (`catchup_failed`, `catchup_dropped`, `catchup_inflight`) added alongside for new dashboards that want unambiguous naming.
- **`total_failed` / `total_dropped` cumulative keys** added so operators can distinguish "catch-up batch had a hiccup" from "steady-state has been having problems for hours."

Pinned by `TestPuller_CatchUpStatus_BackwardCompatibleKeys` — asserts all the right keys exist with the right values across the three categories.

## Tests

19 catch-up + gate sub-tests pass:
- 8 `FullyCaughtUp` semantic tests (4 from this round, 4 from earlier rounds)
- `InflightCount_MirrorsMap`
- `MarkCatchUp_DropCompensation` (NEW)
- 3 `CatchUpFailure_*` self-heal/idempotency/isolation tests
- 3 `CatchUpDrop_*` (NEW: self-heal, idempotency)
- `CatchUpStatus_BackwardCompatibleKeys` (NEW)
- 5 existing `RunCatchUp*` walker tests still green after the mark-before-Enqueue restructure

Full regression on `internal/api`, `internal/cluster/...`, `internal/config` is clean. Build clean, vet clean.

## Docs

- Release notes updated to describe both self-heal flavors.
- Public docs: [`docs.basekick.net` commit `3ab37d0`](https://github.com/Basekick-Labs/docs.basekick.net/commit/3ab37d0) — drop self-heal + `total_failed`/`total_dropped` cumulative observability.

/gemini review

xe-nvdk:
Both findings addressed in commit 679f2fa.

## HIGH — `wasCatchUp` snapshot race

You caught a real correctness bug. The reactive-pull-already-in-flight + walker-tags-late sequence would silently drop the catch-up-failure signal. Fixed exactly as suggested:

- Removed the `wasCatchUp := p.isCatchUpPath(entry.Path)` snapshot at function entry.
- Defer now does `case failed && p.isCatchUpPath(entry.Path):` — reads the tag at outcome-time, before `inflightRemove` clears it.

The new test `TestPuller_ProcessEntry_LateCatchUpTag` pins this: reactive enqueue → late `markCatchUp` → failure → defer must see the late tag and call `recordCatchUpFailure`. Buggy snapshot-at-entry behavior would have left `catchupFailed == 0` and the gate would have opened on a missing file.

## MEDIUM — `failed`/`dropped` cumulative semantics

You're right, my reasoning was wrong on the previous round. Mixing cumulative and scoped under unmarked keys was inconsistent (`pulled`/`skipped_dup` were still cumulative), and overwriting the cumulative `failed`/`dropped` would silently hide steady-state issues from operators who'd been alerting on those keys.

Fixed per your suggested code:

- `failed` and `dropped` restored to **cumulative whole-puller-lifetime totals**. Pre-#392 dashboards continue to report what they always did.
- `catchup_failed` / `catchup_dropped` are the only keys carrying the catch-up-scoped values FullyCaughtUp consumes.
- Dropped the redundant `total_failed`/`total_dropped` since the cumulative semantic is back where it belongs.

The renamed test `TestPuller_CatchUpStatus_KeySemantics` asserts the contract: `failed=10` (cumulative), `dropped=20` (cumulative), `catchup_failed=3`, `catchup_dropped=5`. Distinguishable values so the test pins exactly which counter populates which key.

## Tests + regression

- 14 catch-up + gate puller sub-tests pass.
- New `TestPuller_ProcessEntry_LateCatchUpTag` for the race fix.
- Renamed and rewritten `TestPuller_CatchUpStatus_KeySemantics`.
- Full regression on `internal/api`, `internal/cluster/...`, `internal/config` clean. Build + vet clean.

## Docs

- Release notes: clarified the check-in-defer race fix and the cumulative-semantics restoration.
- Public docs: [`docs.basekick.net` commit `f03e61e`](https://github.com/Basekick-Labs/docs.basekick.net/commit/f03e61e) — clarified that `failed`/`dropped` are cumulative, with the operator-facing example for distinguishing batch-hiccup from steady-state.

/gemini review

xe-nvdk:
Applied in 22cc4fc. Agreed — the boolean-chain form reads better and the doc comment above the function carries all the WHY for each condition, so the early-return shape was just visual noise. Nice catch on the polish.

All 14 catch-up + gate sub-tests still green. Build + vet clean.

This is now PR's sixth pass with no remaining findings flagged. If you sign off on this round, ready to merge whenever @xe-nvdk gives the go-ahead.

PR #422, opened 2026-05-08
fix(storage): strip URL scheme from s3_endpoint before DuckDB

xe-nvdk:
## Summary

- The AWS SDK Go v2 accepts \`s3_endpoint\` with or without an \`http(s)://\` prefix and prepends the scheme if missing. DuckDB's \`httpfs\` extension expects a bare \`host:port\` and prepends scheme based on \`s3_use_ssl\`. The two consumers had OPPOSITE expectations.
- Result: when a user set \`s3_endpoint = "http://host:port"\` in \`arc.toml\` (matching the AWS SDK convention and \`deploy/docker-compose/oss-s3/docker-compose.yml:25\`), DuckDB built malformed URLs of the form \`http://http://host:port/...\` and every \`read_parquet()\` against S3 failed with \`Could not resolve hostname\`.
- Added a small \`stripURLScheme\` helper called at both DuckDB \`SET GLOBAL s3_endpoint\` sites: \`configureS3Access\` (startup) and \`ConfigureS3\` (runtime, for tiered storage). Either input format now works.

## Origin

This bug was caught empirically during the PR #420 memtest reproduction. Customers running Garage or other S3-compatible backends over plain HTTP would hit this if they ever set the scheme in \`arc.toml\`.

## Test plan

- [x] \`go build ./internal/database/... ./cmd/arc/...\` clean.
- [x] \`go test ./internal/database/... -run TestStripURLScheme\` — 7 cases passing (\`http://\` strip, \`https://\` strip, no-scheme passthrough, empty string, scheme-not-at-start no-match).
- [x] End-to-end memtest validation: set \`s3_endpoint = "http://minio:9000"\` in the memtest harness, ran retention against 871 small parquet files. **Pre-fix**: every \`read_parquet\` errored with \`Could not resolve hostname\`. **Post-fix**: retention completed cleanly in 2s, deleted 668 files (441,600 rows). No \`http://http://\` strings appear in the Arc logs.
- [ ] Spot-check on AWS S3 (TLS endpoint) — the path through \`stripURLScheme\` strips \`https://\` too, so this needs a sanity check post-merge.

## What this PR explicitly does NOT do

- Doesn't change SDK-side endpoint handling — that already accepts both forms correctly via the existing \`strings.HasPrefix\` check at \`internal/storage/s3.go:144\`.
- Doesn't update \`arc.toml\` comments or \`config.go\` docstrings — both forms now work, so the existing docs are no longer wrong; users who follow either convention get a working deployment.
- Doesn't add config-load validation — accepting any input the helper can normalise is the simpler contract.

### DIFF
--- a/internal/database/duckdb.go
+++ b/internal/database/duckdb.go
@@ -44,6 +44,17 @@ func escapeSQLString(s string) string {
 	return strings.ReplaceAll(s, "'", "''")
 }
 
+// stripURLScheme removes a leading "http://" or "https://" prefix.
+// DuckDB's httpfs extension expects a bare "host:port" in s3_endpoint and
+// prepends scheme based on s3_use_ssl. The AWS SDK accepts either form, so
+// users routinely include the scheme — passing it verbatim to DuckDB
+// produces "http://http://host:port/..." URLs that fail to resolve.
+func stripURLScheme(endpoint string) string {
+	endpoint = strings.TrimPrefix(endpoint, "https://")
+	endpoint = strings.TrimPrefix(endpoint, "http://")
+	return endpoint
+}
+
 // Config holds DuckDB configuration
 type Config struct {
 	MaxConnections int
@@ -195,7 +206,7 @@ func configureS3Access(db *sql.DB, cfg *Config, logger zerolog.Logger) error {
 
 	// Set custom endpoint for MinIO or S3-compatible services
 	if cfg.S3Endpoint != "" {
-		if _, err := db.Exec(fmt.Sprintf("SET GLOBAL s3_endpoint='%s'", escapeSQLString(cfg.S3Endpoint))); err != nil {
+		if _, err := db.Exec(fmt.Sprintf("SET GLOBAL s3_endpoint='%s'", escapeSQLString(stripURLScheme(cfg.S3Endpoint)))); err != nil {
 			return fmt.Errorf("failed to set s3_endpoint: %w", err)
 		}
 	}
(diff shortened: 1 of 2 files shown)


xe-nvdk:
@gemini-code-assist taken in 3781db9. You're right — RFC 3986 schemes are case-insensitive and users do paste mixed-case from docs / IDE autocomplete; the case-sensitive \`strings.TrimPrefix\` would have left \`HTTP://minio:9000\` untouched, producing the exact same \`http://HTTP://...\` malformed URL we're fixing.

Slight tweak from your suggested form: I used \`strings.ToLower\` only for the prefix check and sliced the ORIGINAL string by \`len("http(s)://")\` so the remainder preserves its case. The new "preserve case in remainder" test case asserts \`http://MyBucket.example.com\` → \`MyBucket.example.com\` (not \`mybucket.example.com\`) — important because hostnames are case-insensitive at the DNS layer but bucket names and path components can be case-sensitive depending on the S3 implementation.

Tests cover all 12 shapes: 7 original + uppercase HTTP/HTTPS + mixed-case Http/Https + case-preservation. All pass.

Ready for another pass.

xe-nvdk:
@gemini-code-assist round 2 — both taken in d056a98.

**Taken: trim whitespace + trailing slashes** ([internal/database/duckdb.go:54](internal/database/duckdb.go)) — \`http://minio:9000/\` and \` http://minio:9000\ \` would otherwise produce \`minio:9000/\` and \`minio:9000\ \` after the scheme strip; the first makes httpfs build \`http://minio:9000//arc-test/...\` and the second breaks hostname resolution. Fixed by trimming in order: whitespace → scheme → trailing slash. Restructured the prefix branch to a \`switch\` so the post-strip TrimRight runs uniformly.

**Taken: rename "https with path" → "https with port"** ([internal/database/duckdb_test.go](internal/database/duckdb_test.go)) — the test input \`https://garage.example.com:3900\` contains a port, not a path. Trivial typo fix.

Six new test cases cover the new trim behaviour: trailing slash (single, multiple, no-scheme), whitespace (leading+trailing, combined with trailing slash), whitespace-only → empty. 18 cases total, all pass.

Ready for another pass.

PR #424, opened 2026-05-10
fix(query): break streaming loop on Flush error so DuckDB result buffers free on client disconnect

xe-nvdk:
## Summary

- Heavy queries against \`/api/v1/query\` (and \`/api/v1/query/arrow\`) hold DuckDB result-set memory until query completion or the 5-min timeout when the client disconnects mid-stream — Grafana panel close, browser tab kill, dashboard refresh. The streaming goroutine has no way to learn the client closed the TCP connection because fasthttp's \`RequestCtx.Done()\` only fires on server shutdown (\`server.go:2719-2745\`), and \`c.UserContext()\` is already cancelled by the time the async stream writer runs.
- The reliable signal in fasthttp's streaming model is the error returned by \`bufio.Writer.Flush()\` — when the underlying TCP connection is closed, the next flush surfaces EPIPE. The fix captures that error and breaks the streaming loop.
- Same pattern applied to the three streaming write paths: Arrow IPC (\`query_arrow.go\`), Arrow-to-JSON (\`query_arrow_json.go\` — what the default \`/query\` endpoint uses on \`-tags=duckdb_arrow\` builds), and the pure database/sql JSON fallback (\`query_json_writer.go\`).
- Aligned the four \"stream truncated after headers committed\" log sites to \`Warn\` (from \`Error\`). Headers are already committed when streamErr fires, so the client got a partial result — by definition the server worked. Client disconnect is expected ops noise.

## Origin

Filed in response to a user-reported pattern: heavy time-bucket aggregations and wide GROUP BYs cause RSS to climb during query execution. Investigation found multiple potential leak sources; this PR fixes the one with the clearest causal mechanism and reproducer (client-disconnect path).

A throttled \`DuckDB.ClearHTTPCache()\` from the query path (parquet_metadata_cache eviction) was investigated and tested in the same branch but **dropped from this PR** — at 1200 queries on a small local-backend dataset, the with-fix and without-fix runs were indistinguishable (both converged to ~180 MB RSS post-settle). Without empirical evidence at the test scale, shipping it would be cargo-cult. May revisit at production scale.

## Test plan

- [x] \`go build ./internal/api/... ./cmd/arc/...\` clean on both default and \`-tags=duckdb_arrow\` build configurations.
- [x] \`go test -tags=duckdb_arrow ./internal/api/...\` — all existing tests pass plus new \`TestStreamArrowJSON_FlushErrorBreaksLoop\` (7 tests in the Arrow JSON test file).
- [x] New regression test uses \`errAfterNBytes\` (an \`io.Writer\` that fails after N bytes) and asserts:
  - the returned error wraps the sentinel via \`errors.Is\`
  - the row count is bounded to within \`jsonFlushInterval + batchRows\` of the failure point (i.e. break fires at the next 5000-row Flush boundary, not later)
- [ ] **Spot-check on Grafana**: open a heavy panel against the \`/query\` endpoint, kill the browser tab mid-load. Pre-fix: DuckDB result reader keeps draining for up to 5 min. Post-fix: reader stops within the next flush window.

## Internal review

Four parallel reviewer agents (correctness, security, code-quality, performance) per the CLAUDE.md review protocol. Findings addressed:

- **Same bug in non-Arrow JSON path** (code-quality HIGH) → applied identical fix in \`query_json_writer.go:169\`.
- **Asymmetric severity Error vs Warn** (3 agents) → aligned all four streaming-truncation log sites to Warn.
- **Stale test comment + loose assertion** (multiple agents) → fixed comment, tightened bound to \`jsonFlushInterval + batchRows\` instead of \`>= totalRows\`.
- **No IPC-path test coverage** (code-quality MEDIUM) → noted as a follow-up. The IPC path uses structurally identical code to the JSON path that the test exercises. Extracting a testable helper from the Fiber handler is a separate refactor.

Pushed back on:
- **Convert reader.Release() to defer** (correctness MEDIUM) — pre-existing pattern, not introduced by this fix, separate hardening.
- **Short-circuit envelope-close on streamErr** (correctness HIGH) — the design is intentional (\"emit valid JSON regardless\" for clients parsing as bytes arrive); cost on disconnect is one no-op JSON marshal.
- **Cancel ctx immediately on flush-error break** (performance MEDIUM) — re-read the code, this is already correct: the cancel() at end of closure fires after the break, and \`reader.Next()\` isn't called again because we broke the loop.
- **Add sentinel error / metric** (multiple LOW) — separate observability PR.
- **Convert test to CheckedAllocator** (perf NIT) — the existing \`buildArrowBatch\` helper leaks builder allocations on every test (pre-existing bug); fixing it would scope-creep. Filed mentally as a follow-up.

### DIFF
--- a/RELEASE_NOTES_2026.06.1.md
+++ b/RELEASE_NOTES_2026.06.1.md
@@ -112,6 +112,28 @@ The eliminated +72 MB-during-idle growth is the customer's exact symptom — RSS
 
 **Note on `MALLOC_ARENA_MAX=2`.** A separate experiment with `MALLOC_ARENA_MAX=2` cut residue further but caused 25–100% latency regression across the standard query suite, so it was not adopted. The 30s throttle on `ReleaseToOS()` keeps allocator-lock contention bounded without restricting glibc's per-thread arena count.
 
+### Query Path: Abort Streaming on Client Disconnect (PR #TBD)
+
+The Arrow-based streaming query handlers (`/api/v1/query` and `/api/v1/query/arrow`) write results to the client via Fiber's async `SetBodyStreamWriter`. When a client closed the connection mid-stream — closing a Grafana panel, killing a browser tab, Cmd-W on a dashboard — the streaming goroutine had no way to learn the client had gone away. fasthttp's `RequestCtx.Done()` only fires on server shutdown, not per-request disconnect ([fasthttp@v1.51.0/server.go:2719-2745](https://github.com/valyala/fasthttp/blob/v1.51.0/server.go#L2719-L2745)), and the streaming code used `context.Background()` deliberately because `c.UserContext()` is cancelled when the handler returns (before the async stream writer runs).
+
+The streaming loop would keep calling `reader.Next()` on the Arrow record reader, draining DuckDB result batches into a buffer nobody was reading, until either the query naturally completed or the per-request `queryTimeout` (default 300s) fired. For heavy time-bucket aggregations and wide GROUP BYs on long time ranges, that's tens of MB of result-set memory held per cancelled query.
+
+The fix is mechanical: capture the error from `bufio.Writer.Flush()` and break the streaming loop on the first failed flush, which is the canonical signal in fasthttp's streaming model that the underlying TCP connection has been closed. Six lines of change per handler in [internal/api/query_arrow.go](internal/api/query_arrow.go) and [internal/api/query_arrow_json.go](internal/api/query_arrow_json.go). Regression test in `query_arrow_json_test.go` uses an `io.Writer` that fails after N bytes and asserts the loop breaks before draining the full result set.
+
+This is **complementary to but distinct from** the #420 retention/delete leak: that fix targeted DuckDB native heap residue after S3 reads; this one targets in-flight Arrow record batches held on the goroutine stack when the client abandons a query mid-stream.
+
+### S3 Endpoint Scheme Normalization for DuckDB (PR #422)
+
+The AWS SDK Go v2 accepts `s3_endpoint` with or without an `http(s)://` prefix; DuckDB's `httpfs` extension expects a bare `host:port` and prepends scheme based on `s3_use_ssl`. With `s3_endpoint = "http://host:port"` in `arc.toml` (matching the AWS SDK convention), DuckDB built malformed URLs of the form `http://http://host:port/...` and every `read_parquet()` against S3 failed with `Could not resolve hostname`.
+
+Added a small `stripURLScheme` helper in [internal/database/duckdb.go](internal/database/duckdb.go), called at both `SET GLOBAL s3_endpoint` sites (startup and runtime reconfigure for tiered storage). Case-insensitive, also trims whitespace and trailing slashes — accepts `http://host:port`, `https://host:port/`, `HTTP://host:port`, ` host:port `, and the bare `host:port` form transparently. 18 unit test cases.
+
+### DELETE Rewrite on Non-TLS S3 (PR #423)
+
+The DELETE API rewrites parquet files to remove rows matching a WHERE clause and uploads them back to S3. Against plain-HTTP S3 (MinIO, Garage), every rewrite failed with `compute input header checksum failed, unseekable stream is not supported without TLS and trailing checksum`. AWS SDK Go v2 (`aws-sdk-go-v2/service/s3 v1.99.0`, post-Feb 2025) requires either TLS or a seekable body for the mandatory request checksum, and the previous `io.TeeReader` single-pass SHA256+upload pattern lost the underlying `*os.File`'s seekability.
+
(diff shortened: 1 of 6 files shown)


xe-nvdk:
@gemini-code-assist round 1 — both taken in 4109b43.

**Taken: \`ipcWriter.Close()\` log downgrade** ([internal/api/query_arrow.go:201](internal/api/query_arrow.go)) — you're right. When the loop broke because the client disconnected, \`ipcWriter.Close()\` is guaranteed to fail flushing trailing IPC metadata over the already-closed connection. Logging that at \`Error\` defeated the ops-noise reduction. Now \`Warn\` with a comment explaining the cause/effect.

**Taken: tighter test assertion** ([internal/api/query_arrow_json_test.go](internal/api/query_arrow_json_test.go)) — I added a temporary \`t.Logf\` to verify the actual rowCount before tightening:

\`\`\`
query_arrow_json_test.go:415: actual rowCount=5000, jsonFlushInterval=5000
\`\`\`

Empirically: bufio's internal 4 KiB auto-flush hits the failing writer well before row 5000 and stores a sticky error; subsequent \`WriteByte\`/\`WriteString\` calls silently no-op until the explicit \`Flush()\` at row 5000 returns the stored error. So the exact-boundary check holds. Changed \`rowCount > tighterBound\` to \`rowCount != jsonFlushInterval\`.

Ready for another pass.

xe-nvdk:
@gemini-code-assist round 2 — all four findings addressed in 743736b. The change ended up bigger than expected because finding #2 required introducing the sentinel error to disambiguate cleanly.

**HIGH 1+2: wasted work between flush boundaries** ([query_arrow_json.go:287](internal/api/query_arrow_json.go), [query_json_writer.go:176](internal/api/query_json_writer.go))

Considered per-row Write-error check but rejected as too expensive for hot-path queries — adds a branch per WriteByte across millions of rows, and \`bufio.Writer\` has no public API to inspect its stored error without forcing a Flush (which incurs a syscall when buffered bytes exist).

Lowered \`jsonFlushInterval\` from 5000 to **1000**. Compromise: 5× faster disconnect detection, ~5× more no-op Flush calls (essentially free), worst-case wasted formatting work bounded to ~1000 rows (~10-100µs at typical query rates). Acceptable.

**MEDIUM 1: log level disambiguation** ([query.go:1536](internal/api/query.go))

You're right — downgrading every stream-truncation error to Warn hid genuine server-side failures (scanner errors, db iteration failures). Fixed by introducing \`errClientDisconnected\` sentinel ([query.go:107-114](internal/api/query.go)) and wrapping the Flush-failure sites with it. Caller log sites now disambiguate:

\`\`\`go
ev := h.logger.Error()
if errors.Is(streamErr, errClientDisconnected) ||
    errors.Is(streamErr, context.DeadlineExceeded) ||
    errors.Is(streamErr, context.Canceled) {
    ev = h.logger.Warn()
}
\`\`\`

Applied at four call sites: \`query_arrow_json.go\`, \`query.go\` (×3 — two streamTypedJSON callers and queryMeasurement), and the newly-refactored \`query_arrow.go\` (see MEDIUM 2 below).

**MEDIUM 2: Arrow IPC handler inconsistency** ([query_arrow.go:201](internal/api/query_arrow.go))

Caught a real asymmetry. The IPC handler logged inline and broke without setting a streamErr, so the closure always logged \"Arrow streaming query completed\" even on disconnect, AND \`IncQueryErrors\` was never incremented for those events. Refactored to mirror the JSON handler: capture all error conditions (ctx cancel, decimal cast, ipcWriter.Write, Flush) into \`streamErr\`, log/metric once at closure end with the same severity disambiguation.

**Test update**: added an \`errors.Is(err, errClientDisconnected)\` assertion alongside the existing underlying-I/O-sentinel check. Tests confirm both ancestors are in the error chain.

Also **closes #428** (sentinel was deferred there as \"defer until a caller actually needs to branch\" — gemini's log-disambiguation finding made the need concrete).

Ready for another pass.

xe-nvdk:
@gemini-code-assist round 3 — both findings taken in aa70be1.

**Taken: \`isClientError\` helper** ([internal/api/query.go:116](internal/api/query.go)) — you're right, the 4-line errors.Is block was repeated 5 times. Extracted two helpers:

\`\`\`go
func isClientError(err error) bool {
    return errors.Is(err, errClientDisconnected) ||
        errors.Is(err, context.DeadlineExceeded) ||
        errors.Is(err, context.Canceled)
}

func (h *QueryHandler) streamErrEvent(err error) *zerolog.Event {
    if isClientError(err) {
        return h.logger.Warn()
    }
    return h.logger.Error()
}
\`\`\`

Each of the 5 sites now reads as a single line:

\`\`\`go
h.streamErrEvent(streamErr).Err(streamErr).
    Int("rows_sent", rowCount).
    Float64("execution_time_ms", ...).
    Msg(...)
\`\`\`

Net **-11 lines** across the diff. Behaviour is identical (same severity disambiguation, same call shape), just expressed once instead of five times.

**Taken: stale test comment** — \`jsonFlushInterval (5000)\` → references the constant symbolically now so it doesn't drift again if we ever retune the interval.

Ready for another pass.

PR #433, opened 2026-05-12
fix(query): mask bare FROM inside EXTRACT/SUBSTRING/TRIM/OVERLAY + Arrow IPC exec-time trailer

xe-nvdk:
## Summary

- **Parser fix**: `SELECT EXTRACT(YEAR FROM time) FROM citibike_trips` (and `SUBSTRING(s FROM 1 FOR 3)`, `TRIM(LEADING '0' FROM x)`, `OVERLAY(s PLACING 'x' FROM 2)`) no longer produces a DuckDB `Binder Error: read_parquet is a table function but used as a scalar`. The regex rewriter previously matched the inner `FROM time` and rewrote `time` as a measurement. `time` is the canonical column name across InfluxDB/Telegraf/Prometheus-derived schemas — every customer migrating hit this.
- **Arrow exec time**: `/api/v1/query/arrow` now emits an `Arc-Execution-Time-Ms` HTTP response trailer so clients can report authoritative server-side timing instead of wall-clock (which overstates Arc perf when network is in the way — playground from Costa Rica saw 4,014 ms wall-clock vs 2,830 ms server-side).

## How it works

Pre-pass before the table-rewrite regexes: paren-depth scanner masks each bare `FROM` inside `EXTRACT/SUBSTRING/TRIM/OVERLAY` argument lists with a unique content-addressed placeholder `__FROM_MASK_<n>__`. The regex passes operate on the masked SQL, then `strings.Replace` restores the originals. Content-addressed (not offset-based) is essential — the original offset-based draft corrupted SQL whenever EXTRACT appeared after FROM/JOIN because the regex rewrite shifts every downstream byte. `TestUnmaskAfterLengthChangingRewrite` locks in that regression.

Helpers live in `internal/sql/mask.go` next to the existing `MaskStringLiterals` / `UnmaskStringLiterals` pair. Both single-table fast paths bail to the masking path when `EXTRACT/SUBSTRING/TRIM/OVERLAY` is detected.

## Overhead per cache-miss query

Cached queries skip the entire rewriter, so this is paid once per unique SQL string:

| Bench | ns/op | B/op | allocs/op |
|---|---:|---:|---:|
| `ContainsFromKeywordFunction_Miss` | 348 | 0 | 0 |
| `ContainsFromKeywordFunction_Hit` | 21 | 0 | 0 |
| `MaskFromKeywordsInFunctionBodies_Miss` | 348 | 0 | 0 |
| `MaskFromKeywordsInFunctionBodies_Hit` | 316 | 336 | 4 |
| `UnmaskFromKeywordsInFunctionBodies` (3 masks) | 285 | 560 | 3 |

Miss path (the 99% of queries): ~350 ns, **0 allocations**. Hit path: ~625 ns, ~900 bytes total.

## Why not a real SQL parser

Evaluated `cockroachdb-parser`, `pg_query_go`, DuckDB's `json_serialize_sql`, and `auxten/postgresql-parser` before writing this fix. No Go SQL parser currently handles DuckDB's full syntax (lambdas `x -> y`, list literals `[1,2,3]`, `QUALIFY`, `EXCLUDE`/`REPLACE`, FROM-first, `:=` named args, `PIVOT`/`UNPIVOT`). The parser migration is tracked as a separate, larger initiative; this PR is a narrow regex pre-pass that closes the user-facing bug now.

## Review process

Two rounds of four parallel internal reviews (correctness/failure-modes, security, code-quality, perf/observability per CLAUDE.md). Round-1 caught a CRITICAL: the original offset-based unmask corrupted SQL when EXTRACT appeared after FROM. Switched to content-addressed placeholders. Round-2 returned ship-it across all four reviewers; remaining LOW/NIT items applied in this branch (allocation-free `isFromKeywordFuncBeforeParen`, `strconv.AppendInt` placeholder formatting, sync.Once-gated AddTrailer warn, dropped verbose comments).

## Test plan

- [x] `go test ./internal/sql/ ./internal/api/` — green
- [x] `go test -tags duckdb_arrow ./internal/sql/ ./internal/api/` — green
- [x] `go build ./cmd/... ./internal/...` — green
- [x] `go build -tags duckdb_arrow ./cmd/... ./internal/...` — green
- [x] `go test ./internal/sql/ -bench=. -benchmem` — numbers above
- [ ] Manual end-to-end against `h01-customers-basekick-net` demo:
  - [ ] `curl POST /api/v1/query` with `SELECT EXTRACT(YEAR FROM time) FROM citibike_trips LIMIT 1` returns one row, not a Binder Error
  - [ ] `curl POST /api/v1/query` with `SELECT SUBSTRING(s FROM 1 FOR 3)`, `TRIM(LEADING '0' FROM x)`, `OVERLAY(s PLACING 'x' FROM 2)` — all rewrite correctly
  - [ ] `curl POST /api/v1/query/arrow` followed by reading `r.headers['Arc-Execution-Time-Ms']` — matches server log `execution_time_ms` within ±1 ms
- [ ] Verify rewriter cache invalidation works for queries that now exercise the mask path
- [ ] Confirm no perf regression on the cache-hit path (rewriter cache should absorb identical queries)

## Discovered

May 12, 2026 during playground build.

### DIFF
--- a/RELEASE_NOTES_2026.06.1.md
+++ b/RELEASE_NOTES_2026.06.1.md
@@ -2,6 +2,59 @@
 
 > **Status:** In progress. Entries are added as PRs land.
 
+## Bug fixes
+
+### Parser no longer mis-resolves bare `time` column inside `EXTRACT(YEAR FROM time)`
+
+The regex-based query rewriter previously matched the `FROM` keyword inside `EXTRACT(YEAR FROM time)` (and the same shape in `SUBSTRING(s FROM 1 FOR 3)`, `TRIM(LEADING '0' FROM x)`, `OVERLAY(s PLACING 'x' FROM 2)`) and rewrote the column reference as a measurement, producing:
+
+```
+Binder Error: Function "read_parquet" is a table function but it was used as a scalar function.
+LINE 1: SELECT EXTRACT(YEAR FROM read_parquet('/app/data/arc/.../time/**/*.parquet'...
+```
+
+`time` is the canonical column name across InfluxDB / Telegraf / Prometheus-derived schemas, so every customer migrating from those systems hit this. Workarounds were `YEAR(time)`, `date_trunc('year', time)`, or quoting / qualifying the column.
+
+26.06.1 adds a pre-pass that masks `FROM` keywords inside the argument list of `EXTRACT`, `SUBSTRING`, `TRIM`, and `OVERLAY` before the table-rewrite regex runs, then restores them afterwards. The masker tracks paren depth so nested calls like `EXTRACT(YEAR FROM CAST(t AS DATE))` are handled correctly. Both the standard and `x-arc-database` header-optimized rewriter paths are covered, and the fast paths bail to the masking path when these functions are present. The single-table query optimization continues to apply for queries that don't use these functions.
+
+**Overhead** (per cache-miss query — cached queries skip the entire rewriter): ~350 ns and **0 allocations** for queries that don't use these functions (the common case); ~625 ns and ~900 bytes for queries that do. The rewriter cache absorbs repeated identical queries, so this is paid once per unique SQL string. Bench numbers from `internal/sql/mask_test.go` on an M3 Max:
+
+```
+BenchmarkContainsFromKeywordFunction_Miss        348 ns/op      0 B/op   0 allocs/op
+BenchmarkContainsFromKeywordFunction_Hit          21 ns/op      0 B/op   0 allocs/op
+BenchmarkMaskFromKeywordsInFunctionBodies_Miss   348 ns/op      0 B/op   0 allocs/op
+BenchmarkMaskFromKeywordsInFunctionBodies_Hit    316 ns/op    336 B/op   4 allocs/op
+BenchmarkUnmaskFromKeywordsInFunctionBodies      285 ns/op    560 B/op   3 allocs/op
+```
+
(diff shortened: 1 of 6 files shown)


xe-nvdk:
Round-2 Gemini comments addressed in d5c826b. Summary:

- HIGH (subquery FROM at deeper paren depth) — fixed by masking FROM only when `depth == stack[top]`. Tests `subquery FROM inside EXTRACT must not be masked` and `deeply nested subquery` added. Verified end-to-end against ClickBench: `SELECT EXTRACT(YEAR FROM (SELECT MIN(EventDate) FROM hits))` now reaches DuckDB binder (failing on `USMALLINT -> DATE` type, not on Arc's rewriter).
- MEDIUM (inline comment between name and paren) — fixed with `skipWhitespaceAndCommentsForward` / `skipWhitespaceAndCommentsBack`. `EXTRACT /* c */ (...)`, `SUBSTRING -- inline\n(...)`, multi-line `/* ... */` all pass.
- MEDIUM (backtick identifier) — rejected alongside double-quote. DuckDB's parser doesn't accept backticks today, but the defensive guard is cheap.
- MEDIUM (NewReplacer) — `UnmaskFromKeywordsInFunctionBodies` switched to `strings.NewReplacer`, single-pass walk.

Also fixed (caught by manual integration test, not in Gemini review): runtime panic at `c.Context().Response.Header.Set(...)` inside the Arrow stream-writer closure — `c *fiber.Ctx` is pooled and reset after the handler returns. Now captures the fasthttp RequestCtx pointer before SetBodyStreamWriter and threads it in as a local (560369f).

End-to-end validation: 305 ms server-side trailer observed via raw socket on `SearchPhrase` top-10 GROUP BY of 100M-row ClickBench; identical query via JSON endpoint reports 311 ms.

/gemini review

PR #434, opened 2026-05-13
feat(arcx): Arc-side loader for the proprietary arcx DuckDB extension

xe-nvdk:
## Summary

Wires Arc Enterprise to **LOAD** a proprietary external DuckDB extension called **arcx** at startup. The extension itself lives in a separate private repo (`Basekick-Labs/arcx`, not yet pushed). v0.1 of arcx is a single `arcx_version()` proof-of-life UDF — real operators (partition-aware scan, manifest-backed `read_parquet`, partition-aligned aggregation fast path) land later. This PR is **loader plumbing only**.

## Why

CPU profile against ClickBench (244 cold queries, 90s capture, 999s CPU): **99.95%** of query CPU is inside DuckDB's C++ or the Go↔C border. Go-side has zero perf headroom. The DuckDB extension API is the only remaining lever for Enterprise-grade query perf.

Locked strategic direction documented in [memory/project_duckdb_extension_for_arc](memory link). Other paths (own SQL engine, Rust sidecar, validation cache, pool affinity) were ruled out on profile data.

## What ships

- `internal/config/config.go` — `DatabaseConfig.ArcxExtensionPath` + `database.arcx_extension_path` viper key (env: `ARC_DATABASE_ARCX_EXTENSION_PATH`)
- `internal/license/license.go` + `client.go` — `FeatureArcx = "arcx"` + `CanUseArcx()` on both types
- `internal/database/duckdb.go` — new `openDuckDB()` routes the pool through `duckdb.NewConnector` with a `connInitFn` that runs `LOAD '...'` on every new connection; `verifyArcxLoaded()` pins a connection via `db.Conn(ctx)` and runs `SELECT arcx_version()` for proof-of-life
- `cmd/arc/main.go` — license gate before `database.Config` build; clears the path with a Warn if `!licenseClient.CanUseArcx()`
- `internal/database/duckdb_arcx_test.go` — unit tests for `buildDSN` + opt-in end-to-end (`ARCX_TEST_PATH` points at a built `.duckdb_extension`)
- `RELEASE_NOTES_2026.06.1.md` — new `## Enterprise` section

## Design decisions worth flagging

**License enforcement is Arc-side only.** No in-extension crypto verification. The `.duckdb_extension` binary is the licensing perimeter — distribute it only inside Arc Enterprise builds. License expiry mid-process does **not** unload arcx (DuckDB has no `UNLOAD`); operators who need to revoke must restart Arc.

**`allow_unsigned_extensions=true` in the DSN** is a database-scope security relaxation. Mitigated by Arc's existing `dangerousSQLPattern` regex denying `LOAD`/`INSTALL` in user SQL (verified against the regression suite at `internal/api/query_test.go`).

**Per-connection LOAD via `connInitFn`** — DuckDB's `LOAD` is per-connection (no `SET GLOBAL` equivalent). A bare `db.Exec("LOAD …")` would only register arcx on whichever pool member `database/sql` happened to hand us. Going through `duckdb.NewConnector` with the init callback guarantees every pool connection has arcx loaded. This was caught in internal review as a CRITICAL bug in the original commit and fixed in the follow-up.

## Internal review

Ran the four-agent parallel review per CLAUDE.md (correctness, security, code-quality, perf/observability) **before pushing**. The original commit had a CRITICAL: per-connection LOAD only loaded arcx on one pool member. The follow-up commit (`6f2ffb6`) addresses that plus the MEDIUMs: dead placebo test deleted, release-note wording corrected (database-scope not connection-scope), `cfg != nil` dead-guard dropped, license godoc expanded.

## Test plan

- [x] `go build ./cmd/... ./internal/...` clean (default + `duckdb_arrow` tags)
- [x] `go test ./internal/database/ ./internal/license/ ./internal/config/ ./internal/api/` all green
- [x] Negative path verified end-to-end: Arc with `ARC_DATABASE_ARCX_EXTENSION_PATH` unset starts cleanly, `SELECT arcx_version()` returns `Catalog Error` (arcx genuinely not loaded)
- [x] License-denial path verified: setting the env without a licensed `arcx` feature produces `WRN arcx extension configured but license does not include 'arcx' feature` + `arcx_enabled=false` + Arc proceeds normally
- [ ] Positive path with a real built `.duckdb_extension` — pending the `make` in the arcx repo (~30 min DuckDB compile, deferred to a separate task). `TestArcxLoadsAndReportsVersion` will run that path with `ARCX_TEST_PATH=` set.

## Out of scope

- The arcx repo itself (separate private repo `Basekick-Labs/arcx`, not yet pushed to GitHub).
- Real operators (partition-aware scan, manifest-backed `read_parquet`, partition-aligned aggregation fast path) — separate PRs.
- ed25519 / in-extension license verification — explicitly rejected (single source of truth = Arc).
- Phase 2 RBAC integration when operators land — the future `arc_scan(db, m)` operator runs **inside DuckDB** with no Arc Fiber context; we'll need to gate it via SQL pre-rewrite after RBAC, otherwise it becomes an RBAC-bypass primitive. Flagged in internal security review; not blocking this PR.


xe-nvdk:
Round-1 Gemini findings addressed in 91e1144:

- **HIGH (duckdb.go:184)**: `connInitFn` now uses `context.WithTimeout(30s)` so a corrupt/network-mounted extension file can't hang pool acquisition indefinitely. 30s is generous; real `dlopen` + Load() takes tens of ms.
- **MEDIUM (duckdb.go:181)**: comment-vs-code mismatch fixed. The closure now captures `path := cfg.ArcxExtensionPath` into a local and references that everywhere (including the logger field), so the closure no longer retains the full `*Config` for its lifetime.

Build + tests green on default and `duckdb_arrow` tags.

@gemini-code-assist review

xe-nvdk:
@gemini-code-assist The MEDIUM finding at duckdb.go:191 in this round is already addressed by commit 91e1144 — the closure captures `path := cfg.ArcxExtensionPath` into a local on line 182 and uses `path` (not `cfg.ArcxExtensionPath`) in both the SQL interpolation (line 183) and the logger field (line 198). Could you verify against the latest HEAD?

Current state of the file (lines 179-203):

```go
// Capture the path into a local so the closure does not retain the
// whole *Config (the logger field below would otherwise pull cfg in
// for its entire lifetime — gemini round 1).
path := cfg.ArcxExtensionPath
loadSQL := fmt.Sprintf("LOAD '%s'", escapeSQLString(path))
componentLogger := logger.With().Str("component", "duckdb").Logger()

const arcxLoadTimeout = 30 * time.Second

connector, err := duckdb.NewConnector(dsn, func(execer driver.ExecerContext) error {
    ctx, cancel := context.WithTimeout(context.Background(), arcxLoadTimeout)
    defer cancel()
    _, execErr := execer.ExecContext(ctx, loadSQL, nil)
    if execErr != nil {
        componentLogger.Error().Err(execErr).
            Str("path", path).
            Msg("arcx LOAD failed on new DuckDB connection")
        return fmt.Errorf("arcx LOAD on new connection: %w", execErr)
    }
    return nil
})
```

The closure no longer references `cfg`.

xe-nvdk:
@gemini-code-assist Round-3 partially addressed in ba177b7:

**Accepted (✓)**: Windows path compat — added `filepath.ToSlash(cfg.ArcxExtensionPath)` so a Windows operator configuring `C:\Program Files\arcx\arcx.duckdb_extension` doesn't trip DuckDB's LOAD on backslash handling. No-op on Linux/macOS.

**Disagreed (✗)**: the claimed type mismatch on `duckdb.NewConnector`'s callback signature. The actual signature in `duckdb-go/v2@v2.10501.0` ([duckdb.go:59](https://pkg.go.dev/github.com/duckdb/duckdb-go/v2@v2.10501.0)) is:

```go
func NewConnector(dsn string, connInitFn func(execer driver.ExecerContext) error) (*Connector, error)
```

— the callback parameter genuinely IS `driver.ExecerContext`, not `driver.Conn`. The code builds clean on both `go build ./cmd/... ./internal/...` and `go build -tags duckdb_arrow`, and all package tests pass. Inserting an `(any).(driver.ExecerContext)` type assertion is unnecessary and would actually fail to compile against this driver version.

If you're checking against a different/older version of duckdb-go, the discrepancy is on that side. Could you re-verify against the current HEAD on the branch?

xe-nvdk:
@gemini-code-assist Round-4 findings addressed in d9df8f7:

**Fixed**:
- **MEDIUM (duckdb.go:207)** log flooding — connInitFn Error log is now gated by sync.Once. The error itself still propagates to database/sql on every call (misconfiguration not hidden), only the log line is throttled. Same pattern as arrowTrailerWarnOnce from PR #433.
- **MEDIUM (duckdb_arcx_test.go:74)** test reused idle pool conn — rewrote the test to acquire n concurrent *sql.Conn via db.DB().Conn(ctx) so database/sql is forced to open distinct connections. Each pinned conn then runs SELECT arcx_version(); if the per-connection LOAD ever regresses, conn 2+ fails with "function does not exist". Thanks for the catch — my original loop was a false sense of coverage.

**Acknowledged but not actionable**:
- **MEDIUM (duckdb.go:191)** SQL injection via fmt.Sprintf — DuckDB's LOAD is DDL-class and does not accept prepared-statement parameters. The path is operator-supplied at startup (config file or env var), not user-supplied at runtime, and goes through the existing escapeSQLString helper that's the standard Arc pattern for DuckDB string-literal interpolation (same as configureS3Access for S3 credentials). I'll keep this approach for consistency unless DuckDB ever ships parameterized LOAD.

Build + tests green on both default and duckdb_arrow tags.

PR #435, opened 2026-05-13
feat(database): wire arcx storage_root + denylist arc_partition_agg in user SQL

xe-nvdk:
## Summary

Companion to **[arcX#1](https://github.com/Basekick-Labs/arcX/pull/1)** — adds the Arc-side wiring needed to make arcx's first real operator (`arc_partition_agg(database, measurement, unit)`) usable from Arc's DuckDB pool, and closes the RBAC bypass that direct user-SQL access to the function would create.

- `database.Config.ArcxStorageRoot` field + `connInitFn` extension: runs `SET arcx.storage_root = '<LocalPath>'` immediately after the existing `LOAD '<extension>'` on every pooled connection. Same `arcxLoadTimeout` budget. Skipped when arcx isn't configured, so OSS connInit is unchanged.
- `cmd/arc/main.go`: populates `ArcxStorageRoot` from `cfg.Storage.LocalPath` only when the arcx loader is enabled.
- `internal/api/query.go`: adds `arc_partition_agg(` to `ValidateSQLRequest`'s denylist, matching the existing `read_parquet(` block. Without this, an authenticated user scoped to `db1` could call `arc_partition_agg('db2', 'mem', 'hour')` directly and enumerate row counts in databases they don't own — the function takes raw `db/measurement` strings and globs the filesystem. Only Arc's query rewriter (which has the user's identity) may emit calls to it.
- `internal/database/duckdb_arcx_test.go`: opt-in integration test `TestArcxStorageRootIsSetOnEveryConn` confirms the setting is applied across **distinct pool connections** (so a rolling failover to a fresh pool member doesn't break the function).
- `RELEASE_NOTES_2026.06.1.md`: Enterprise-section entry with the measured speedups and explicit call-out that the Arc query rewriter is not yet wired to emit `arc_partition_agg` automatically.

## Measured impact (local M1, real Arc data)

| Workload                              | Files  |   Rows | Speedup on `date_trunc … GROUP BY` |
| ------------------------------------- | -----: | -----: | ---------------------------------: |
| citibike (14 yr, daily-compacted)     | 4,782  | 137 M  |                              5.1× |
| production (1 hr, hourly compacted)   |     5  | 393 M  |                               35× |
| synth (20 days × 24 hr × 5/hr)        | 2,400  | 189 B  |                               76× |

Full Q1 scale / Q2 predicate pushdown / Q3 multi-tenant correctness review is in the arcx repo at `docs/2026-05-13-operator-1-partition-agg-readiness.md`.

## Security context (the denylist)

Same shape as the existing `read_parquet(` block in `ValidateSQLRequest`. Without the new entry:

```sql
-- User authenticated for db1 issues:
SELECT * FROM arc_partition_agg('db2', 'mem', 'hour');
-- arcx globs /<root>/db2/mem/**/*.parquet, returns row counts for db2.
-- RBAC bypassed.
```

With this PR:

```
Bad request: Direct arc_partition_agg() calls are not allowed in user SQL
```

Arc's query rewriter (which has the authenticated user's identity) remains the only path that may emit `arc_partition_agg(...)` calls. The denylist runs on string-literal-masked, comment-stripped SQL, so literals containing the text `arc_partition_agg` are not false-positives.

## Test plan

- [x] `go test ./internal/database/ ./internal/api/` — all passing, including new cases.
- [x] `TestValidateSQLRequest_BypassesAndFalsePositives` — 5 new cases (direct, uppercase, whitespace, CTE, literal-text false-positive). All passing.
- [x] `TestArcxStorageRootIsSetOnEveryConn` — opt-in (requires `ARCX_TEST_PATH`). Verified locally against the [arcX#1](https://github.com/Basekick-Labs/arcX/pull/1) build.
- [x] Manual smoke test: arcx loader fires with `SET arcx.storage_root = '/data/arc'` on every pool connection; `SELECT current_setting('arcx.storage_root')` returns the configured path across multiple distinct connections.
- [x] OSS path unchanged: with `ArcxExtensionPath=""`, neither `LOAD` nor `SET` runs; `database/sql.Open` takes the bare driver path.


xe-nvdk:
@gemini-code-assist please review this PR. Companion to arcX#1 (private repo). The wiring is minimal (one SET statement appended to the existing connInitFn closure; one regex added to ValidateSQLRequest); the bulk of the operator lives in the arcx repo. Highlighting the denylist as the load-bearing security boundary — without it, user SQL can bypass Arc RBAC by globbing other customers' databases via arcx.

PR #438, opened 2026-05-15
feat(query): columnar MessagePack query response endpoint (experimental)

xe-nvdk:
## Summary

- New `POST /api/v1/query/msgpack` returns query results as **columnar MessagePack** — `data` is an array of per-column arrays, with a parallel `types[]` field. Matches Arc's columnar storage and ingest architecture.
- Same execution pipeline as `POST /api/v1/query` (auth, RBAC, governance, ctx-timeout, profile, query-registry, slow-query logging); only the response serialization differs. Gated by the `duckdb_arrow` build tag.
- **2.49× faster than JSON** on `LIMIT 1M` (334ms → 134ms). **78% of Arrow IPC's throughput.**
- Experimental: the endpoint may move, change shape, or be removed based on production feedback. Don't hard-code against it from production clients yet.

## End-to-end head-to-head

393M-row `cpu` measurement on M3 Max, p50 latency in milliseconds:

| Query | JSON | msgpack (columnar) | Arrow IPC | msgpack vs JSON |
|---|---:|---:|---:|---:|
| `LIMIT 100K` |  48 |  33 |  31 | **1.45×** |
| `LIMIT 500K` | 173 |  81 |  61 | **2.14×** |
| `LIMIT 1M`   | 334 | 134 | 105 | **2.49×** |

DuckDB-bound queries (`GROUP BY`, `Percentile`, aggregations) are within run-to-run noise across all three wire formats — serialization is sub-millisecond and DuckDB execution dominates.

## The columnar shape was the win

Two implementation rounds:

1. **Row-oriented draft**: `data: [[v,v,v], ...]`. Delivered 1.74× over JSON on `LIMIT 1M` (193ms). Bottleneck: 1M rows × 7 cols = 7M per-cell type assertions in `writeArrowValueMsgPack`.
2. **Columnar redesign**: `data: [[col0_v, col0_v, ...], [col1_v, ...]]`. Hoists the type-switch outside the row loop — one switch per column. Delivered 2.49× over JSON (134ms).

The redesign was the only change that moved the needle. A subsequent optimization round (per-batch ctx check, `EncodeInt64`/`EncodeUint64` direct, `enc.EncodeTime` for timestamps, drop NaN/Inf check, drop per-column flush, 256 KiB bufio) added ~4ms — within run-to-run noise.

## Why we stopped (pprof receipts)

A CPU profile (`go tool pprof` against the live server under 100 concurrent `LIMIT 1M` queries, 14-core box) shows the Go-side encoder accounts for **zero measurable samples**. 25-second profile, 100ms of total Go-side work captured:

- 30ms `runtime.cgocall` — DuckDB query execution through the Go/C boundary, opaque to Go pprof.
- 30ms `syscall.rawsyscalln` — TCP writes via fasthttp's chunked encoder.
- The msgpack encode loop, type-switches, bufio writes: **all rounded to zero on the profile.**

So the remaining 28ms gap to Arrow IPC lives in **DuckDB's Arrow record materialization** on the C++ side. Arrow IPC writes DuckDB's internal column buffers via `ipcWriter.Write(batch)` — a memcpy plus a small framing header. The msgpack endpoint forces DuckDB to walk each record's cells via `c.Value(i)`, which makes DuckDB do per-cell materialization work Arrow IPC's batch-buffer write skips entirely. No amount of Go-side optimization closes that gap; it's structural.

The release notes include the full optimization summary table.

## Wire format

Single msgpack map per response:

```
map(7 or 8) {
  "success":           bool
  "columns":           [string, ...]
  "types":             [string, ...]                          // arrow.DataType.String() per column
  "data":              [[col0_v, ...], [col1_v, ...], ...]    // numCols outer, numRows inner
  "row_count":         uint
  "execution_time_ms": uint                                   // milliseconds
  "timestamp":         string                                 // RFC3339
  "profile":           map                                    // only when ?profile=true
}
```

- **Timestamps** use msgpack's native timestamp Ext type (4/8/12 bytes, no RFC3339 round-trip).
- **Binary columns** use msgpack's native `bin` type.
- **Null cells** are msgpack nil.
- **Errors, `SHOW DATABASES`, `SHOW TABLES`** also use this envelope shape via new `respondError` / `respondSuccessRows` / `respondEmptySuccess` helpers that branch on a `wire_format` request-local.

## Implementation notes worth flagging

- **No `database/sql` fallback.** When the Arrow driver is unavailable (build without `duckdb_arrow`, or driver capability mismatch), the route returns `501 Not Implemented`. A `Scan`-based fallback would defeat the contract — the entire point of the endpoint is the typed Arrow → msgpack encode.
- **Parallel-partition execution is bypassed** for msgpack queries. The parallel executor's merge iterator is wired for JSON streaming; coupling them would multiply the experimental surface area.
- **Buffered, not streamed.** msgpack arrays need a length prefix, so the endpoint drains all Arrow batches into memory before emitting the data array. The only row ceiling is `governance.max_rows` (Enterprise token policy); without it, the response is unbounded.
- **Profile field tagging.** The msgpack library defaults to `msgpack:"..."` struct tags. `QueryProfile` has only `json:"..."` tags, so the encoder calls `SetCustomStructTag("json")` once to keep field names `snake_case` on the wire.

## Internal review

Two rounds of four parallel staff/principal-engineer review agents (correctness/failure-modes, security, code-quality/Go-idioms, performance/observability) per CLAUDE.md before opening. Blockers addressed across the rounds:

- Profile struct tag (`SetCustomStructTag("json")`) — fixed.
- `SHOW DATABASES` / `SHOW TABLES` JSON contract leak — fixed via `respondSuccessRows`.
- 501 short-circuit when `arrowMsgPackQueryFunc == nil` (non-tag build) — fixed.
- Governance cap-before-Retain — fixed (trim trailing batch with `NewSlice`).
- Dead `pendingRow` type, `scratch` return value, `cap` shadowing builtin — all cleaned up.
- Per-cell type-switch (the dominant cost in the row draft) — solved by the columnar redesign.

## Test plan

- [x] `go build ./cmd/... ./internal/...` and `go build -tags=duckdb_arrow ./cmd/... ./internal/...` — both clean.
- [x] `go test -tags=duckdb_arrow ./internal/api/` — all tests pass.
- [x] `go test ./internal/api/ -run TestArrowMsgPack` (non-tag) — nil-symbol assertion passes.
- [x] Manual end-to-end: rebuild with `-tags=duckdb_arrow`, run `query_suite` against all three targets (`arc`, `arc-msgpack`, `arc-arrow`), confirm 134ms/105ms/334ms split on `LIMIT 1M`.
- [x] CPU profile under saturating concurrent load — confirms encoder is not the bottleneck.

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,29 +113,32 @@ Benefits:
 
 ### Query (May 2026)
 
-Arrow IPC format provides up to 2.9x throughput vs JSON for large result sets. Both protocols hit the same DuckDB engine; the speedup comes from skipping JSON encoding for typed columnar batches.
+Arc speaks three wire formats from the same query engine. **Arrow IPC** is the throughput leader for analytical clients (Grafana, pyarrow, polars) that can take an Arrow dependency — zero-copy from DuckDB's internal columnar buffers. **MessagePack** (experimental, columnar) is the choice for clients that don't speak Arrow but want smaller bytes and faster decode than JSON — same envelope shape as JSON, native binary types for timestamps and binary columns. **JSON** stays the default for ergonomic compatibility.
 
-Benchmark: 393.7M-row `cpu` measurement, 5 iterations per query, DuckDB 1.5.1.
+Benchmark: 393.7M-row `cpu` measurement, 5 iterations per query, M3 Max, DuckDB 1.5.1. Latency is p50 in milliseconds. The five SELECT-LIMIT rows were measured back-to-back in the same session so the three columns are apples-to-apples; the DuckDB-bound rows (Time Bucket, Date Trunc, GROUP BY) are dominated by query execution and converge across wire formats.
 
-| Query | Arrow (ms) | JSON (ms) | Speedup |
-|-------|------------|-----------|---------|
-| COUNT(*) - 393.7M rows | 0.86 | 1.03 | 1.20x |
-| SELECT LIMIT 10K | 15.0 | 18.0 | 1.20x |
-| SELECT LIMIT 100K | 32.2 | 50.0 | 1.55x |
-| SELECT LIMIT 500K | 70.5 | 177.1 | **2.51x** |
-| SELECT LIMIT 1M | 118.7 | 339.2 | **2.86x** |
-| Time Range (7d) LIMIT 10K | 15.5 | 15.0 | 0.97x |
-| Time Bucket (1h, 7d) | 4.7 | 4.7 | 1.00x |
-| Date Trunc (day, 30d) | 413 | 416 | 1.01x |
-| GROUP BY host | 450 | 452 | 1.00x |
-| GROUP BY host + hour | 672 | 645 | 0.96x |
+| Query | JSON (ms) | MessagePack (ms) | Arrow IPC (ms) | msgpack vs JSON | Arrow vs JSON |
+|-------|----------:|-----------------:|---------------:|----------------:|--------------:|
+| COUNT(*) — 393.7M rows | 1.03 | 1.03 | 0.86 | 1.00x | 1.20x |
+| SELECT LIMIT 10K | 18.4 | 16.6 | 14.7 | 1.11x | 1.25x |
+| SELECT LIMIT 100K | 48.1 | 33.2 | 31.0 | **1.45x** | **1.55x** |
+| SELECT LIMIT 500K | 173.2 | 81.1 | 61.1 | **2.14x** | **2.84x** |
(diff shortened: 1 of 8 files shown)


xe-nvdk:
@gemini-code-assist /review

xe-nvdk:
@gemini-code-assist /review

xe-nvdk:
@gemini-code-assist /review

xe-nvdk:
@gemini-code-assist /review

PR #441, opened 2026-05-19
feat(server): support binding to a specific host/address — closes #439

### LINKED ISSUE #439: feat(server): support binding to a specific host/address (currently always wildcard)
## Summary

Arc has no native configuration option to bind the HTTP server to a specific host/address — the listen socket is always wildcard (`0.0.0.0` + `::`). Operators wanting loopback-only deployments (e.g. Arc fronted by a sidecar adapter on the same host) currently have to reach for systemd cgroup filters, ufw, or a reverse proxy.

## Current state

`internal/api/server.go:369` builds the listen address from port alone:

```go
addr := fmt.Sprintf(":%d", s.port)
```

The `[server]` section of `arc.toml` exposes `port` + `tls_enabled` + `tls_cert_file` + `tls_key_file`, but no `host`/`bind`/`address`. The matching env-var override path (`ARC_SERVER_*`) likewise has no entry.

## Proposed change

Add a `host` field to `config.ServerConfig` (default empty for backward-compat = wildcard):

```toml
[server]
host = "127.0.0.1"   # default "" preserves current wildcard behaviour
port = 8000
```

…and use it in `server.go`:

```go
addr := net.JoinHostPort(s.host, strconv.Itoa(s.port))
```

`net.JoinHostPort` handles IPv6 bracketing correctly (`[::1]:8000`).

## Workaround for now

For deployments wanting loopback-only access we're using a systemd drop-in:

```ini
[Service]
IPAddressDeny=any
IPAddressAllow=127.0.0.1/32 ::1/128
```

This is effective (off-host connections time out before reaching the socket), but `ss -ltnp` still shows `*:8000` which surprises ops folks on first look, and it depends on cgroup-v2 + BPF.

xe-nvdk:
## Summary

- Adds `server.host` config (and `ARC_SERVER_HOST` env var) so operators can bind Arc's HTTP listener to a specific interface — useful behind a reverse proxy where Arc should only be reachable on loopback (#439).
- Default is the empty string, which preserves the pre-26.06.1 wildcard behavior with Linux dual-stack semantics (IPv4 + IPv6 via IPv4-mapped addresses). `net.JoinHostPort(\"\", \"8000\")` produces `\":8000\"`, **byte-identical** to the historical `fmt.Sprintf(\":%d\", port)` — zero behavioral change for operators upgrading without setting `server.host`.
- Listener switches from `fmt.Sprintf(\":%d\", port)` to `net.JoinHostPort(host, port)` so IPv6 addresses are bracketed correctly (e.g. `[::1]:8000`).
- Finishes plumbing that was already half-wired: viper read `server.host` into `config.Server.Host` but `cmd/arc/main.go` dropped it on the floor when copying into `api.ServerConfig`.

### Why empty default and not `0.0.0.0`

Setting `0.0.0.0` forces **IPv4-only** binding — silent break for IPv6 clients. Empty string keeps dual-stack on Linux. Called out in the release notes and the docs.

## Test plan

- [x] `go build -tags=duckdb_arrow ./cmd/... ./internal/...` clean
- [x] `go test ./internal/api/` — new `TestListenAddrForHost` (7 cases) + `TestServerCapturesHostFromConfig` (4 cases) pass
- [x] `go test ./internal/config/` passes
- [x] Empirically verified `net.JoinHostPort(\"\", \"8000\") == \":8000\"` (byte-identical to historical `fmt.Sprintf`)
- [x] Manual smoke: `ARC_SERVER_HOST=127.0.0.1 arc` → listener bound to loopback only
- [x] Manual smoke: no env var set → listener bound to all interfaces (dual-stack on Linux)
- [x] Docs PR for `docs.basekick.net` already merged: documents the new setting

## Related

- Closes #439
- Docs: Basekick-Labs/docs.basekick.net@85b13e7

### DIFF
--- a/RELEASE_NOTES_2026.06.1.md
+++ b/RELEASE_NOTES_2026.06.1.md
@@ -203,6 +203,32 @@ go run benchmarks/query_suite/main.go --target arc-arrow  --measurement cpu --it
 
 ## Hardening
 
+### HTTP server now supports binding to a specific host/address — closes #439
+
+The HTTP listener has always bound the wildcard address (`":<port>"`), giving operators no way to restrict the bind from the config file. Deployments wanting loopback-only behavior (Arc fronted by a sidecar adapter on the same host) reached for `systemd` `IPAddressDeny`, `ufw`, or a reverse proxy. The `[server]` config block was missing the corresponding knob.
+
+26.06.1 adds `server.host` to the `[server]` config block, plumbed through to the Fiber listener via `net.JoinHostPort`. Any address Go's `net` package recognizes works — IPv4 literals, IPv6 literals (with correct bracketing), hostnames, the explicit IPv4 wildcard `0.0.0.0`, the explicit IPv6 wildcard `::`. The matching env override is `ARC_SERVER_HOST`.
+
+**Default is empty string** — the listener constructs `":<port>"`, byte-identical to the historical `fmt.Sprintf(":%d", port)`. This preserves Linux dual-stack wildcard behavior (IPv4 + IPv6 via IPv4-mapped addresses); explicit `"0.0.0.0"` would force IPv4-only and silently break IPv6 clients on upgrade, so it's an opt-in. **Zero behavioral change on the listener for operators who don't touch their config.**
+
+```toml
+[server]
+host = "127.0.0.1"   # loopback-only
+# host = "::1"        # IPv6 loopback
+# host = "192.0.2.10" # specific NIC
+# host = "0.0.0.0"    # explicit IPv4-only wildcard
+# host = "::"         # explicit IPv6-only wildcard
+# host = ""           # default; dual-stack wildcard (matches pre-26.06.1 behavior)
+```
+
+The startup log line now also reports the bound host, so `journalctl -u arc | grep 'Starting Arc server'` answers "what did we bind to?" without grepping config files:
+
+```
+INF Starting Arc server component=api-server host=127.0.0.1 port=8000 tls_enabled=false protocol=HTTP
+```
+
+Operators currently relying on the `systemd` `IPAddressDeny` workaround can keep it — `ss -ltnp` will now show the bind address Arc actually opened, but the systemd filter remains valid defense-in-depth.
+
 ### Hard Query Gating During Replication Catch-Up (Enterprise, opt-in) — closes #392
 
 Reader nodes in a clustered Arc Enterprise deployment previously accepted queries the moment they started, even while peer file replication was still pulling Parquet files the rest of the cluster already had. The Raft manifest knew about the missing files; the local storage didn't have them yet; `read_parquet()` globbed against local storage rather than the manifest. The result was **silent partial results** during the catch-up window. The Phase 3 release explicitly deferred this fix; 26.05.1 closed half the gap (WAL replication makes unflushed writer data queryable on readers within milliseconds), but flushed Parquet files still depended on async background pullers.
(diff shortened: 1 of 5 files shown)


xe-nvdk:
@gemini-code-assist /review

xe-nvdk:
@gemini-code-assist — addressed all three findings in 627b16a:

1. **Double-bracketing on already-bracketed IPv6** — extracted `listenAddr(host, port)` in [internal/api/server.go](https://github.com/Basekick-Labs/arc/blob/feat/server-host-bind/internal/api/server.go#L67-L82) that strips surrounding brackets before `net.JoinHostPort`. Production path and tests now share the same code.
2. **Log clarity for wildcard variants** — the startup log now distinguishes:
   - `* (wildcard, dual-stack)` for the empty default
   - `0.0.0.0 (wildcard, IPv4-only)`
   - `:: (wildcard, IPv6-only)`
3. **Release notes wording** — clarified that users write addresses *without* brackets; surrounding brackets in user input are stripped defensively.

Tests gained two cases pinning the bracket-strip: `[::1]` → `[::1]:8000` and `[::]` → `[::]:8000`. All 13 sub-tests pass; both `-tags=duckdb_arrow` and default builds clean.

/review

xe-nvdk:
@gemini-code-assist — addressed both findings in 9962d0a:

🔴 **High — `cmd/arc/main.go:864` IPv6 fmt bug** — confirmed and fixed. The cluster coordinator's `apiAddr` now calls `api.ListenAddr` instead of `fmt.Sprintf("%s:%d", …)`, so an IPv6 host like `::1` produces `[::1]:8000` rather than the invalid `::1:8000`. Good catch — this was a latent bug that the IPv4-only-ever world masked, and the `server.host` plumbing in this PR is what would have exposed it in production.

🟡 **Medium — export `listenAddr`** — done; renamed to `api.ListenAddr` and called from both the server's `Start()` and the cluster coordinator setup. The shared helper is the right shape: a future caller that needs the same canonicalization gets it for free, and the format duplication that hid the IPv6 bug can't reappear.

The `0.0.0.0` special-case is preserved deliberately: operators who explicitly bind the IPv4 wildcard probably don't want a literal `0.0.0.0:<port>` peer URL advertised to the cluster — they want the historical `:<port>` shape. Adjusting that behavior is out of scope here.

Both builds clean; all tests pass.

/review

xe-nvdk:
@gemini-code-assist — addressed one of two findings in 5d28713:

🟡 **Fixed: `::` log label** — you're right, Go's `net.Listen("tcp", "[::]:port")` is dual-stack on most modern Linux (`bindv6only=0`) and v6-only on systems where `IPV6_V6ONLY` is explicitly set. The family is platform-dependent, not literal-dependent. Relabeled to `:: (wildcard, platform default)` so the log doesn't overclaim. The empty default's `(wildcard, dual-stack)` label stays because that case is byte-identical to the historical `:<port>` bind and we've always intended/observed dual-stack there.

🟡 **Declined: `0.0.0.0` cluster apiAddr special-case** — leaving it as-is, for one concrete reason: `0.0.0.0:8000` is not a valid advertise address for cluster peers. Peers can't dial "all interfaces" of a remote host; they need a routable address. Switching to always-`ListenAddr` would change the fallback from `:8000` (working but vague) to `0.0.0.0:8000` (functionally broken for peer dial-out) for any operator who explicitly bound the IPv4 wildcard and didn't set `cluster.advertise_addr`.

The fallback `apiAddr` at `cmd/arc/main.go:866` only kicks in when `cluster.advertise_addr` is unset — that's the canonical knob, and it already exists at `internal/config/config.go:344` (set via `ARC_CLUSTER_ADVERTISE_ADDR` env var or `[cluster] advertise_addr = "…"` in TOML). Operators running on `0.0.0.0` who care about clustering should set that explicitly; the special-case is just the lesser-broken fallback. Changing this without a separate UX pass on the advertise-address story would silently degrade clustering for `0.0.0.0` users on upgrade.

If you (or a future reviewer) think the right answer is "fail loud and require `cluster.advertise_addr` when `server.host = 0.0.0.0`", happy to do that in a follow-up — but that's a behavior change, not a refactor.

Both builds clean; all tests pass.

Issue #448, opened 2026-05-21
internal/cluster/sharding is unwired dead code — decide: finish, remove, or document


xe-nvdk:
## Status

Discovered while securing cluster WAL replication for GHSA-wfgr-8x84-22q7 (X1 audit). The `internal/cluster/sharding` package compiles and has tests, but **none of its constructors are called from production code**:

- `NewShardReplicationManager` — only called from `internal/cluster/sharding/shard_replication_test.go`
- `NewShardReceiverManager` — only called from `internal/cluster/sharding/shard_replication_test.go`
- `NewShardRouter` — only called from `internal/cluster/sharding/router_test.go` (route helpers `RouteShardedWrite` / `RouteShardedQuery` exist in `internal/api/routing.go` but are not invoked from any non-test caller either)

The coordinator dispatch in `internal/cluster/coordinator.go` has **no server-side listener** for the sharding `MsgReplicateSync` flow. Both `shard_replication.go` (primary side) and `shard_receiver.go` (replica side) define `connect()` functions that **dial out**, leaving no side that accepts the inbound handshake. The flow is asymmetric and incomplete.

## What Arc Enterprise actually uses today

The single-writer + multi-reader Raft + replication model from `internal/cluster/coordinator.go` + `internal/cluster/replication/` (the path secured by GHSA-wfgr-8x84-22q7). Customer deployments run this model. **Sharding is not a deployed feature.**

## Why this matters

1. **Confused-reviewer hazard.** Security audits + Gemini reviews keep flagging `internal/cluster/sharding/*` as relevant attack surface. It isn't — but every reviewer has to re-derive that fact from scratch. The X1 fix had to explicitly carve it out of scope in the release notes.
2. **Maintenance surface without a customer.** ~1.2k lines of replication code that compiles and tests but ships dead in every binary. Every refactor in `internal/cluster/replication/protocol.go` or `internal/cluster/security/auth.go` has to either also touch the sharding mirrors or knowingly skip them.
3. **Half-finished design.** When sharding IS picked up, the implementer should redesign the auth model from scratch — the current handshake direction is ambiguous (both ends dial). Retrofitting the X1 HMAC onto today's shape would be wasted work.

## Options

[...]

PR #449, opened 2026-05-21
fix(security): HMAC-authenticate cluster WAL replication (closes GHSA-wfgr-8x84-22q7)

xe-nvdk:
## Summary

- Closes audit finding **X1** / [GHSA-wfgr-8x84-22q7](https://github.com/Basekick-Labs/arc/security/advisories/GHSA-wfgr-8x84-22q7) / **CVE-2026-48106** (Alex Manson / @NeuroWinter).
- Three-layer HMAC scheme on the writer→reader replication stream: handshake HMAC, per-entry truncated MAC tag (8 bytes, HKDF session key), periodic full-HMAC checkpoint every 1024 entries.
- Per-entry overhead **391 ns / 552 B / 8 allocs** at 256B payload — ~0.08% of a single core at Arc's typical ingest rate.
- Two adjacent pre-existing bugs fixed alongside (surfaced by the two-node smoke): writer success-ack wire format mismatch (`unknown message type: 19`) and 30s-idle disconnect cycle (errors.As unwrap on net.Error timeout). Enterprise WAL replication was effectively broken before this PR.

## Test plan

- [x] `go test ./internal/cluster/replication/... ./internal/cluster/security/...` — all green
- [x] `go test -race ./internal/cluster/replication/... ./internal/cluster/security/...` — all green
- [x] `go vet ./internal/cluster/...` clean
- [x] `gofmt -l` clean on touched files
- [x] 7 new round-trip tests in `internal/cluster/replication/auth_roundtrip_test.go` cover refuse-when-unconfigured (both ends), per-entry tag verification (green path + tampered payload + sequence-shifted tag), checkpoint emission at interval, checkpoint suppression before interval, per-connection key isolation, and multi-reader distinct-tag invariant
- [x] Benchmarks in `internal/cluster/replication/auth_bench_test.go` measure per-entry and per-validation cost at 64B/256B/1KB/8KB payloads
- [x] **Two-node Arc Enterprise cluster smoke test** (writer + reader, license `ARC-ENT-9A99-...`, no auth, no telemetry): handshake authenticated, 31 ingested records replicated end-to-end via WAL stream, 0 handshake failures, 0 disconnects across 55s idle, all per-entry tags verified, no checkpoint hash mismatches
- [x] Internal review with explicit configuration matrix (Cluster + Writer ± secret, Cluster + Reader ± secret, OSS standalone, Standalone role, Compactor role, TLS dimensions) before reviewer agent; addressed M2/M3/M4 documentation, S2/S4 hot-path nits, S5 release-notes correction
- [ ] Gemini Code Assist review (requested after open)

## Operator-facing change

`cluster.shared_secret` is now **required** for WAL replication. Pre-26.06.1 deployments running with an empty secret will see refuse-when-unconfigured log lines on both ends until the secret is populated. Rolling upgrades are NOT supported for this change — pre-26.06.1 readers send no auth fields and 26.06.1 writers refuse them; full cluster restart with the secret pre-staged is the supported upgrade path. Full operator-facing notes in `RELEASE_NOTES_2026.06.1.md` under "Cluster replication stream now HMAC-authenticated end-to-end (Enterprise)".

## Out of scope

The per-shard replication code in `internal/cluster/sharding/` is unwired dead code today — tracked in #448 for a separate "finish, remove, or document" decision.


xe-nvdk:
Addressed round-1 findings in 98da97c:

- **R1 (HIGH)**: Receiver now closes `r.conn` alongside clearing the session key when `receiveLoop` returns. No more leaked sockets on HMAC fail / malformed entry / read-error reconnects.
- **R2 (MEDIUM)**: Added `security.ComputeReplicationEntryTagFromHash` variant; `broadcastEntry` computes the payload SHA-256 once per entry and reuses it across every reader. Bench: 386 ns → 282 ns (27% faster per reader after the first). At N readers, drops per-broadcast hash cost from N × sha256(payload) to 1 × sha256(payload).

/gemini review

xe-nvdk:
Round-2 fixes pushed in 5608e75 — three low-severity polish items from your PR-level summary:

1. **Centralized tolerance**: Added `security.HMACTimestampTolerance = 5 * time.Minute` and replaced 7 hardcoded literals across coordinator.go, forward_apply.go, receiver.go, and cmd/arc/main.go (cache-invalidate alias). Single source of truth.
2. **`remote_addr` in receiver logs**: receiveLoop now snapshots `conn.RemoteAddr()` once and derives a sub-logger; every error/warn/debug line in the loop carries it for free.
3. **Doc cross-reference**: `MsgReplicateCheckpoint` doc now explicitly names `SenderConfig.CheckpointInterval` and the `defaultCheckpointInterval = 1024` constant.

/gemini review

xe-nvdk:
Round-2 fixes pushed in cc439aa:

**Addressed in code (4/5)**

- **(receiver.go:452)** Length-check `entry.Tag` before `hex.DecodeString`. Prevents an attacker who sends an arbitrarily long Tag from making us allocate a multi-MB byte slice in DecodeString just to throw it away. Length error and decode error are now logged separately.
- **(receiver.go:525)** Same length-before-decode pattern for `cp.CumulativePayloadHashHex` (expected `sha256.Size*2 = 64` hex chars).
- **(receiver.go:537)** Comment line-number drift fixed; replaced `line 499` with a structural reference ("subtle.ConstantTimeCompare check just above") so it survives future line-number drift.
- **(sender.go:380)** Latent `entry.Tag` race under future parallelization — fixed structurally. `sendToReader` now stamps Tag on a shallow copy of the entry (struct headers only — payload bytes aren't copied), never on the broadcast-shared pointer. Updated the `ReplicateEntry.Tag` doc comment to match.

**Declined (1/5)** with rationale, not a revert

- **(receiver.go:431)** You're right that `continue → return` on ParseEntry error is a behavior change — and it's intentional. A JSON parse failure on an *authenticated* stream means the framing is corrupt or tampered (the per-entry tag couldn't even be checked because we couldn't decode the envelope). Continuing would let the loop drift past unrelated entries and lose sequence sync — and the next entry's monotonic-sequence check would either fail (drop the connection anyway) or silently accept a sequence gap. Dropping the connection and forcing a fresh handshake (with a new session key) is the correct security posture for this case. The `continue` was load-bearing pre-X1 because the stream was unauthenticated; post-X1 the right move is to assume the stream is poisoned and re-handshake. Happy to revisit if you see a scenario where transient JSON corruption is a real concern.

/gemini review

xe-nvdk:
Post-Gemini-sign-off final sanity pass caught one more bug that all 3 Gemini rounds + the matrix-driven internal review missed. Fixed in 96af752.

**Concurrency: success-ack write raced concurrent broadcast on the same conn**

After Gemini round 3 signed off, ran one more deep review specifically asking about lock hierarchies across the file boundaries this PR touches (because the wire-format fix moved the success-ack write from `Sender.AcceptReader` to `Coordinator.AcceptReplicationConnection`).

The original write happened while `s.mu` was still held inside `AcceptReader`. Post-refactor:

1. `Sender.AcceptReader` publishes the reader to `s.readers` under `s.mu`, spawns `receiveLoop`, returns (releasing `s.mu` via defer).
2. Coordinator writes the success ack via `protocol.SendMessage` — without holding `reader.writeMu`.
3. Meanwhile, `distributionLoop` is free to RLock `s.mu`, see the new reader, call `sendToReader` (which **does** hold `reader.writeMu`), and write a `MsgReplicateEntry` frame to the same conn.

`net.TCPConn.Write` is per-call goroutine-safe, but a 3-call framed protocol message (`[4-byte length][1-byte type][JSON]`) from each side interleaves at the byte level. Receiver reads garbage → tears down the connection → reconnect retry hides the symptom. Reachable under sustained ingest at hundreds of broadcasts/sec; not surfaced by the 2-node 31-record smoke.

**Fix**: added `Sender.DoWithReaderWriteLock(readerID, fn)` that runs `fn` while holding `reader.writeMu` — same lock the broadcast path takes. Coordinator now wraps the success-ack write in it, so the two writers are serialised by construction. No hot-path impact (lock only fires once per handshake; entry-tag bench unchanged at 386 ns/op).

Verified: all replication+security tests green under `-race`, 2-node smoke still completes handshake cleanly.

/gemini review

xe-nvdk:
Round-4 fixes pushed in 86eecf6:

**G4.1 (HIGH) — Two-phase reader accept**

You're right — my previous fix in 96af752 only addressed the *write* race (byte-level interleave under `reader.writeMu`), not the *delivery-order* race. Even with writeMu serialisation, the broadcast path could acquire the lock first and write an entry frame BEFORE the ack — receiver reads `MsgReplicateEntry` as its first message and fails the handshake.

Structural fix: split `AcceptReader` into `PrepareReader` + `ActivateReader`.

- `PrepareReader` validates inputs, derives the session key, builds the `*ReaderConnection` — but does NOT publish to `s.readers` or start `receiveLoop`.
- Coordinator writes the success ack on the raw conn while the reader is still invisible to `broadcastEntry`.
- `ActivateReader` publishes the reader so the broadcast path can start streaming.
- If the ack write fails, `reader.Discard()` tears down the prepared (but never activated) reader cleanly — no leaked goroutines or map entries.

Race eliminated by **ordering**, not by locks. `DoWithReaderWriteLock` is removed since it's no longer needed (no other writer can reach the conn until after the ack lands and the reader is published).

The legacy single-call `AcceptReader` is kept as a Prepare+Activate wrapper so test code that doesn't exercise the ordering invariant doesn't need to change.

**G4.2 (MEDIUM) — Checkpoint sequence `>` → `!=`**

Agreed — on a single TCP connection the checkpoint follows entry N immediately, so `cp.LastSequence` must EXACTLY equal `r.lastSeq.Load()`. Anything else is a protocol desync. Tightened to `!=` with the suggested error message; desync now surfaces in the log directly instead of via the downstream hash mismatch.

Verified: all replication + security tests green under `-race`; two-node Enterprise smoke completes handshake cleanly with zero "unexpected message type" or sequence-mismatch errors.

/gemini review

xe-nvdk:
Round-5 fixes pushed in 96526d9 — three hot-path allocation cleanups:

- **(receiver.go entry tag decode)** `hex.DecodeString` → `hex.Decode` into a stack-allocated `[ReplicationEntryTagLen]byte`. Fires per entry at ingest rate (~200k/sec on Arc's typical profile), so the heap-allocation savings compound.
- **(receiver.go checkpoint hash verify)** Two heap allocations per checkpoint (one from `cumulativeHash.Sum(nil)`, one from `hex.DecodeString`) → `Sum(buf[:0])` + `hex.Decode` into stack arrays. Also folded the separate `hashArr` copy into the same `theirHash [sha256.Size]byte` since `ValidateReplicationCheckpointHMAC` already takes `[32]byte`.
- **(sender.go checkpoint snapshot)** Same `Sum(nil)` + `copy` pattern in `emitCheckpointLocked` → `Sum(hashSnap[:0])` into the existing array. sha256's Sum doesn't mutate internal state, so the cumulative hash keeps accumulating across checkpoints as before.

All tests green under `-race`. Compute-tag benches unchanged (those measure the sender's tag-compute side, not the receiver's decode path).

/gemini review

xe-nvdk:
Round-6 fixes pushed in 3b23942:

**G6.1 (MEDIUM) — Unknown message type now drops the connection.** Agreed — on an authenticated stream an unknown type is a desync/tamper indicator, not a transient anomaly. Forces fresh handshake (fresh session key, fresh cumulative hash). Consistent with the rest of the receiver's defensive posture (parse fail, tag fail, sequence backwards all already drop).

**G6.2 (MEDIUM) — Per-reader reusable hmac.Hash.** Added:
- `security.NewReplicationEntryHMAC(sessionKey) hash.Hash` — long-lived MAC keyed by the session key
- `security.ComputeReplicationEntryTagWithMAC(mac, seq, payloadHash, dst)` — Reset() + write + Sum into a caller-managed buffer

`ReaderConnection` now holds `entryMAC` (initialized once in `PrepareReader` from the session key). `sendToReader` uses the reusable MAC; `writeMu` already serialises every sendToReader call for the same reader, so the shared mac is safe.

**Bench** (256B payload, M3 Max):
```
ComputeReplicationEntryTag         512 ns / 784 B / 11 allocs
ComputeReplicationEntryTagFromHash 376 ns / 784 B / 11 allocs
ComputeReplicationEntryTagWithMAC  120 ns /  72 B /  3 allocs   ← production hot path
```

**3.1× speedup, ~11× lower allocations** vs the naive call. At Arc's 200k entries/sec target the per-entry overhead drops from ~0.08% to ~0.024% of a single core. Release notes updated with the new bench triple.

All tests green under `-race`. `TestSenderStampsValidTagOnEveryEntry` specifically validates that the new `WithMAC` path produces tags the receiver's `ValidateReplicationEntryTag` accepts end-to-end — so the contract is preserved.

/gemini review

xe-nvdk:
Round-7 fix pushed in 65d5d2e:

**G7.1 (HIGH) — nil-deref on malformed error frame.** Correct catch — `errMsg, _ := ParseError(payload)` discarded the error, and a writer that sends a malformed JSON error frame (misbehaving or tampered) would crash the reader with nil-deref on `errMsg.Code`. Fixed to log the parse error and drop the connection. Same posture as the other "stream is poisoned" branches (entry parse fail, malformed tag, hash mismatch).

All tests still green under `-race`.

/gemini review

xe-nvdk:
Round-8 fixes pushed in 1309333:

**G8.1 (SECURITY-MEDIUM) — Uniform handshake rejection.** You're right — the PR claimed all rejection paths return the same string, but two paths leaked specific state:
- `cfg.SharedSecret == ""` → `"peer replication is not configured on this node"`
- missing auth fields → `"missing authentication fields"`
while the HMAC-fail / cluster-mismatch / nonce-replay paths already returned `"authentication failed"`. Unified all five paths to the same generic string; specific reason remains in the operator log via `logger.Warn().Msg(...)`.

**G8.2 (MEDIUM) — Drop on MsgReplicateError.** Agreed and consistent with the rest of the receiver's defensive posture. An error frame from the writer means the sender declared the stream unhealthy (sequence gap too large, WAL rotated, apply failed, etc.) — continuing would process further entries on a stream that's already in a fatal state. Forces a fresh handshake + session key on reconnect.

All tests green under `-race` for replication + security packages.

/gemini review

xe-nvdk:
Round-9 fixes pushed in 4986573:

**Receiver-side reusable HMAC + shared SHA-256** — symmetric fix to round 6's sender-side change. Added `security.ValidateReplicationEntryTagWithMAC` and refactored `receiveLoop` to compute the payload SHA-256 ONCE per entry, then reuse it for both the per-entry tag verification AND the cumulative checkpoint hash. `entryMAC` lives on the goroutine stack (same lifetime as `cumulativeHash`).

Bench (256B payload, M3 Max):

```
ValidateReplicationEntryTag         494 ns / 776 B / 10 allocs
ValidateReplicationEntryTagWithMAC  127 ns /  72 B /  3 allocs   ← production hot path (~4× faster)
```

Both sender and receiver hot paths now run ~120-127 ns/entry with 3 allocations. At Arc's 200k entries/sec target the per-side overhead is ~0.024% of a single core.

**Note on the stale PR-level summary comments about `coordinator.go`** — the round-9 summary re-flagged "uniform authentication failed string" and "missing fields error message" as if they were unresolved. Those were addressed in 1309333; please re-check against HEAD (`grep 'sendReplicationSyncError(conn,' internal/cluster/coordinator.go` confirms all 6 rejection paths now return the same generic `"authentication failed"` string). Likely a stale-snapshot replay on the review side.

All tests green under `-race` for replication + security.

/gemini review

xe-nvdk:
Round 10 returned the same PR-level summary as round 9 with **zero new inline comments**. That summary re-flags items already addressed:

```
$ grep 'sendReplicationSyncError(conn,' internal/cluster/coordinator.go
1290:		c.sendReplicationSyncError(conn, "authentication failed")
1298:		c.sendReplicationSyncError(conn, "authentication failed")
1309:		c.sendReplicationSyncError(conn, "authentication failed")
1319:		c.sendReplicationSyncError(conn, "authentication failed")
1331:		c.sendReplicationSyncError(conn, "authentication failed")
1341:		c.sendReplicationSyncError(conn, "authentication failed")
```

All 6 handshake rejection paths return the identical generic string at HEAD (1309333); the specific reason is captured only in the operator log. This is what the round-8 summary asked for and it was implemented immediately.

Treating this as the stale-snapshot pattern documented in our review notes — no new code changes needed. **PR is ready for merge from my side**: 11 commits over 10 Gemini rounds, all real findings addressed, race + tests green, two-node Enterprise smoke validated end-to-end.

xe-nvdk:
Round-11 findings — both declined with measurements:

**G11.1 (`[]byte(entry.Tag)` allocation in `hex.Decode`)** — measured, the conversion is **elided by the compiler**. Tried both paths:

```
BenchmarkHexDecodeStdLib-14            156255304    7.637 ns/op    0 B/op    0 allocs/op
BenchmarkDecodeHexStringInto-14         79047264   14.92  ns/op    0 B/op    0 allocs/op
```

Go's escape analysis elides the `[]byte(string)` conversion when the resulting slice doesn't escape the call (`hex.Decode` doesn't retain it). Both paths show **zero allocations**. The std-lib `hex.Decode` is **2× faster** than a custom in-place decoder. I implemented `DecodeHexStringInto` and reverted it — the benchmark proved it was a pessimization.

**G11.2 (`fmt.Sprintf` allocation in `computeReplicateSyncHMACRaw`)** — declined. That function is only called on the **handshake**, not the per-entry hot path. One call per reader reconnect, which is at most every few minutes — not 200k/sec. The PR-level summary asserted this was on a high-performance path; it isn't. Caller trace:

```
$ grep -rn 'ComputeReplicateSyncHMAC\|ValidateReplicateSyncHMAC' --include='*.go' .
internal/cluster/coordinator.go:1314  (handleReplicateSync — once per reader handshake)
internal/cluster/replication/receiver.go:308  (connect — once per receiver reconnect)
```

The handshake budget is dominated by the network RTT (10s timeout) and HKDF (~µs), not by an Sprintf.

**Net: no code changes from round 11. PR is ready to merge.**

xe-nvdk:
Round-12 re-flags the same `[]byte(entry.Tag)` allocation I declined in round 11 (above, [comment 4513425038](https://github.com/Basekick-Labs/arc/pull/449#issuecomment-4513425038)) with bench data. To summarise that for the next review pass:

```
BenchmarkHexDecodeStdLib-14            156255304    7.637 ns/op    0 B/op    0 allocs/op
BenchmarkDecodeHexStringInto-14         79047264   14.92  ns/op    0 B/op    0 allocs/op
```

Go's escape analysis elides `[]byte(string)` when the resulting slice doesn't escape the callee. `hex.Decode` doesn't retain the source slice, so the conversion is stack-allocated and the std lib reports **0 allocs/op**. The custom in-place decoder I tried (per the round-11 suggestion) was **2× slower** and also 0 allocs. **The std lib path is already optimal.**

I implemented `DecodeHexStringInto` to test the suggestion and reverted it after the benchmark — both code states are visible in the branch history if needed.

**Not re-triggering the loop on this one** — the data is in the prior comment. **PR ready to merge.**

PR #475, opened 2026-06-03
fix(import): parse CSV/Parquet in-process via ArrowBuffer (#472)

### LINKED ISSUE #472: import(csv,parquet): replace DuckDB-on-temp-file introspection with in-process row streaming via ArrowBuffer
## Summary

The CSV and Parquet import handlers (`internal/api/import.go#handleCSVImport`, `handleParquetImport`) currently introspect the uploaded temp file by issuing DuckDB queries against `read_csv(...)` / `read_parquet(...)`. This is architecturally inconsistent with the LP and TLE import handlers (which parse rows in-process and push through `ArrowBuffer`), and it's fragile under the new DuckDB sandbox model from #442.

The immediate symptom — surfaced while building [arcctl PR #3](https://github.com/Basekick-Labs/arcctl/pull/3) — is that **CSV imports fail with a DuckDB parser error** before any data lands:

```
arc: failed to read file: query failed: Parser Error: syntax error at or near "("

LINE 1: SELECT column_name FROM (DESCRIBE read_csv('/Users/nacho/dev/basekick-labs/arc/.tmp/arc-uploads/arc-import-XXXX/import.csv', auto_detect=true, header=true))
                                                  ^
```

Came back as HTTP 422 from `POST /api/v1/import/csv` against a tiny 91-byte CSV with valid headers + 3 rows + every flag at defaults. Same failure against a 0-byte CSV (errors during introspection, before the empty-file check fires).

Same handler also fails for Parquet (untested in PR #3 but uses the identical `getColumns` → `read_parquet(...)` path).

## Architecture problem (the bigger fix)

Per @xe-nvdk's clarification on PR #3:

> *"You don't query the CSV, you query the data ingested through `/api/v1/query`. There is no CSV in arc when the data is ingested."*

That's the canonical contract: arc serves queries over **its own Parquet partitions**, not over arbitrary uploaded files. The current CSV/Parquet import handlers violate this by routing the file through DuckDB's `read_csv` / `read_parquet` directly:

```go
// internal/api/import.go (current)

readExpr := h.buildReadExpression(filePath, opts)  // → read_csv('/tmp/.../import.csv', ...)

// 1. schema introspection — direct DuckDB query against temp file
columns, err := h.getColumns(readExpr)              // SELECT column_name FROM (DESCRIBE read_csv(...))

// 3. stats — another direct DuckDB query against temp file
statsQuery := fmt.Sprintf("SELECT COUNT(*), MIN(%s)::VARCHAR, MAX(%s)::VARCHAR FROM %s", timeCast, timeCast, readExpr)
```

This works *only* because the upload directory is in the sandbox `allowed_directories` list (added in #442). It's still:

- **Inconsistent with LP/TLE** — both use `arrowBuffer.WriteColumnarRecord(...)` and never issue DuckDB queries against the temp file. LP parses in-process via `ingest.NewLineProtocolParser`; TLE has its own parser.
- **Fragile against DuckDB version drift** — the `SELECT column_name FROM (DESCRIBE ...)` SQL is rejected by the DuckDB version arc ships today (`duckdb_version=v1.5.1` per the smoke). `DESCRIBE` can't always be used as a derived table.
- **A new sandbox hole every time a temp dir moves** — every directory the import handler writes to has to be allowlisted explicitly. #442 kept the list intentionally minimal; this pattern keeps adding to it.

## Proposed fix

Mirror the LP/TLE handler pattern in CSV and Parquet:

1. **Parse rows in-process** (not via DuckDB).
   - CSV: use `encoding/csv` or an Arrow-aware reader. The header row provides column names directly — no `DESCRIBE` needed.
   - Parquet: use `github.com/apache/arrow-go/v18/parquet` (or whatever arc already uses for compaction). Schema is in the file metadata; no `DESCRIBE` needed.

2. **Push rows through `arrowBuffer.WriteColumnarRecord(...)`** — same code path as LP. This means:
   - Time-column validation moves to the row parser (fail-fast per row, not a separate up-front query).
   - Time-format casting happens in the parser, not via a DuckDB SQL expression.
   - Partition layout, hot/cold tier handling, replication, compaction triggers — all automatic via the ArrowBuffer pipeline.

3. **Post-ingest stats come from the query layer over the ingested partitions**, if the response needs them at all. The current `ImportResult` includes `time_range_min` / `time_range_max` and `partitions_created`; those can be computed by the ArrowBuffer (which already knows what it wrote) without any after-the-fact `SELECT` against `read_csv`.

4. **Drop the upload directory from the sandbox allowlist** once nothing in the import path issues DuckDB queries against `/tmp/arc-uploads/...`. Pure-Go parsing means the temp file is read by `os.Open` + `csv.Reader` (or the parquet equivalent) and deleted before any query runs.

## Out of scope

- LP and TLE imports — already use this pattern.
- **Patching `getColumns` SQL syntax in isolation** (e.g. swap `SELECT ... FROM (DESCRIBE ...)` for `pragma table_info(...)`) — would fix the symptom without fixing the architecture. Don't.
- Audit of every other DuckDB query that touches external files — separate issue; this one is about the import path specifically.

## Severity

**Medium.** Production CSV/Parquet imports are broken today against the current DuckDB version + sandbox config. Workaround: convert source data to line protocol and use `import lp`, which works. But that's a 1-format workaround for an API that ostensibly supports 3 formats.

## Test plan (for the eventual fix)

- [ ] CSV import against a 3-row file with explicit `--time-column` + `--time-format` produces the same `ImportResult` shape as today (rows_imported, partitions_created, time_range_min/max, columns, duration_ms).
- [ ] CSV import against a 0-byte file fails with `"file is empty"` (the same 400 LP returns today), not a parser error.
- [ ] CSV with `delimiter=';'` and `skip_rows=1` ingests correctly.
- [ ] Parquet import against a basic file with `time` column produces matching `ImportResult`.
- [ ] Imported data is queryable via `arcctl query` (or `POST /api/v1/query`) against the same database/measurement.
- [ ] Sandbox `allowed_directories` no longer contains the upload temp dir, and CSV/Parquet imports still succeed (proves the temp file isn't being read via DuckDB anymore).
- [ ] LP and TLE imports continue to work unchanged.

## Discovered while

Building `arcctl import csv` in https://github.com/Basekick-Labs/arcctl/pull/3. arcctl correctly surfaces the verbatim 422 from the server; this is purely an arc-side fix.

xe-nvdk:
## Summary

Fixes #472. Replaces the CSV/Parquet import handlers' DuckDB-on-temp-file introspection (`read_csv`/`read_parquet` + `DESCRIBE` + `COPY TO`) with **in-process row parsing pushed through `ArrowBuffer`**, mirroring the existing Line Protocol and TLE import handlers.

The immediate bug: the `SELECT column_name FROM (DESCRIBE read_csv(...))` SQL is rejected by the DuckDB version Arc ships (v1.5.1), so CSV imports failed with an HTTP 422 parser error before any data landed. The 0-byte case also errored during introspection, before the empty-file check.

## What changed

- **`internal/api/import_inprocess.go`** (new) — `handleCSVImport` / `handleParquetImport` now parse in-process:
  - CSV via `encoding/csv`, with per-column type inference (`int64` → `float64` → `bool` → `string`) so numeric columns stay numeric instead of degrading to strings.
  - Parquet via the arrow-go `pqarrow` reader (already a dependency); schema from metadata, no `DESCRIBE`.
  - Time column normalized to `int64` microseconds (`epoch_s|ms|us|ns`, or auto: magnitude detection for numerics / RFC3339-family parsing for strings).
  - Ingest via `WriteColumnarRecord` → `FlushAll`, same path as LP/TLE.
  - `ImportResult` computed in-process (rows, partitions = distinct hour buckets `t / 3_600_000_000`, time range, columns) — no follow-up DuckDB query.
- **`internal/api/import.go`** — removed the old path (`importFile`, `buildReadExpression`, `buildTimeCast`, `getColumns`, `generateStoragePath`, `escapeSQLString`); trimmed the handler struct (`db`/`storage`/`uploadDir` fields removed); `NewImportHandler(logger)`.
- **`cmd/arc/main.go`** — updated the `NewImportHandler` call. **`delete.go` wiring and the `UploadDir` sandbox allowlisting are intentionally left intact** (see below).
- **`internal/api/import_test.go`** — removed tests for deleted functions; added unit tests for the new pure helpers (`inferAndConvertColumn`, `oneTimeValueToMicros`, `buildImportResult`, `validateImportHeader`).
- **`RELEASE_NOTES_2026.06.2.md`** — Bug fixes section + upgrade note.

## Behavior changes (user-visible)

- 0-byte / header-only upload → `400 "file is empty"` / `"file contains no rows"` (was a 422 parser error).
- Up-front rejection (`400`) of: `time_column` rename colliding with an existing `time` column, and duplicate header column names — both previously could silently drop a column.
- CSV uploads now subject to the same 500 MB cap as the other formats.
- Parquet `DECIMAL` → `DOUBLE` on import (Arc's import path carries no per-column decimal precision).

## Scope note — sandbox allowlist (issue checkbox 4 NOT done, by design)

The issue asked to drop the upload dir from the DuckDB sandbox `allowed_directories`. That is **blocked by `delete.go`**: `cmd/arc/main.go` passes the same `dbConfig.UploadDir` to both `NewImportHandler` and `NewDeleteHandler`, and `delete.go rewriteS3File` does `COPY ... TO <uploadDir>/...` for S3 deletes, which requires the dir to be allowlisted. Import no longer *depends* on the allowlist (no DuckDB queries against the temp file — verified: zero `read_csv`/`read_parquet`/`DESCRIBE` in logs during import), but the entry stays for delete. Full removal is a follow-up that reworks the S3 delete rewrite.

## Test plan

- [x] `go build ./... && go vet ./internal/api/... && gofmt -l` (clean)
- [x] `go test ./internal/api/...` (green)
- [x] Integration (built binary, auth off, telemetry off):
  - [x] CSV import (3 rows) → 200, correct rows/partitions/time-range; queried back; types preserved (`value`→DOUBLE, `flag`→BOOLEAN, `host`→VARCHAR, `time`→TIMESTAMP)
  - [x] 0-byte CSV → `400 "file is empty"`
  - [x] `delimiter=';'` + `skip_rows=1` → correct ingest
  - [x] `time_format=epoch_s` → correct micros
  - [x] missing time column → `400` with available-columns list
  - [x] Parquet import (DECIMAL column) → 200, decimal → DOUBLE, queried back
  - [x] time-column rename collision → `400`
  - [x] duplicate header → `400`
  - [x] zero `read_csv`/`read_parquet`/`DESCRIBE` in server logs during import
- [x] LP/TLE imports unchanged

### DIFF
--- a/RELEASE_NOTES_2026.06.2.md
+++ b/RELEASE_NOTES_2026.06.2.md
@@ -12,12 +12,26 @@ This release strengthens internal verification routines on the Arc Enterprise ac
 
 Operators on 26.06.1 should plan to upgrade.
 
+## Bug fixes
+
+**CSV and Parquet bulk imports now work against the shipped DuckDB version.** `POST /api/v1/import/csv` and `/api/v1/import/parquet` previously introspected the uploaded file by running DuckDB queries against it (`read_csv` / `read_parquet` + `DESCRIBE`); the `DESCRIBE`-as-subquery form was rejected by the DuckDB version Arc ships, so CSV imports failed with an HTTP 422 parser error before any data landed. Both formats now parse rows in-process and ingest through the same streaming pipeline as Line Protocol and TLE imports — no DuckDB queries against the uploaded file.
+
+User-visible behavior changes:
+
+- A 0-byte or header-only upload now returns `400 "file is empty"` / `"file contains no rows"` (previously a confusing parser error).
+- Imports validate up front and reject, with a clear `400`, a file whose `time_column` rename would collide with an existing `time` column, or a file with duplicate column names — cases that could previously have silently dropped a column.
+- CSV uploads are now subject to the same 500 MB size cap already enforced for the other import formats.
+- Parquet `DECIMAL` columns are imported as `DOUBLE` (Arc's ingest path does not carry per-column decimal precision for imports). Use Line Protocol with a configured decimal column if exact decimal precision is required.
+
+Line Protocol and TLE imports are unchanged. CSV/Parquet imports no longer depend on the DuckDB sandbox's allowed-directories list.
+
 ## Upgrade notes
 
 1. **No configuration change required.** Drop in the new binary; existing `arc.toml` and license keys work as-is.
 2. **Active licenses keep working.** Arc binaries running against `enterprise.basekick.net` continue to operate normally; no re-activation or license-key reissuance is required.
 3. **OSS-only deployments** (no `[license]` block in `arc.toml`) are unaffected by the license-verification change. The database-API authorization change applies only when authentication is enabled in `arc.toml`.
 4. **Token review for operators using non-admin tokens for database management**: if any automation provisions databases via `POST /api/v1/databases`, ensure the token it uses has the `admin` permission. Auto-create-on-write (databases that come into existence as a side-effect of line-protocol or msgpack writes) is unaffected — write tokens continue to work for ingestion.
+5. **CSV/Parquet import callers**: imports that relied on a Parquet `DECIMAL` column round-tripping as a decimal will now receive a `DOUBLE`. Imports of malformed files (empty, duplicate headers, or a `time_column` rename that collides with an existing `time` column) now fail fast with a `400` instead of partially succeeding — review any automation that ignored import error responses.
 
 ## Dependencies
 
--- a/RELEASE_NOTES_2026.06.2.md
+++ b/RELEASE_NOTES_2026.06.2.md
@@ -19,7 +19,8 @@ Operators on 26.06.1 should plan to upgrade.
 User-visible behavior changes:
 
 - A 0-byte or header-only upload now returns `400 "file is empty"` / `"file contains no rows"` (previously a confusing parser error).
-- Imports validate up front and reject, with a clear `400`, a file whose `time_column` rename would collide with an existing `time` column, or a file with duplicate column names — cases that could previously have silently dropped a column.
+- Imports validate up front and reject, with a clear `400`: empty, blank, or duplicate column names; a `time_column` rename that would collide with an existing `time` column; or (for Parquet) a `NaN`/`Inf` value in a floating-point time column — cases that could previously have silently dropped a column, crashed the import, or written corrupt partition boundaries.
+- Time columns now accept **fractional (floating-point) epochs** (e.g. `1609459200.123`), preserving sub-second precision; the old DuckDB path tolerated these and the in-process rewrite restores parity. Integer epochs keep full precision (no float round-trip). For Parquet, the time column may be a `TIMESTAMP`, an integer or floating-point epoch, or a timestamp string.
 - CSV uploads are now subject to the same 500 MB size cap already enforced for the other import formats.
 - Parquet `DECIMAL` columns are imported as `DOUBLE` (Arc's ingest path does not carry per-column decimal precision for imports). Use Line Protocol with a configured decimal column if exact decimal precision is required.
 
@@ -31,7 +32,7 @@ Line Protocol and TLE imports are unchanged. CSV/Parquet imports no longer depen
 2. **Active licenses keep working.** Arc binaries running against `enterprise.basekick.net` continue to operate normally; no re-activation or license-key reissuance is required.
 3. **OSS-only deployments** (no `[license]` block in `arc.toml`) are unaffected by the license-verification change. The database-API authorization change applies only when authentication is enabled in `arc.toml`.
 4. **Token review for operators using non-admin tokens for database management**: if any automation provisions databases via `POST /api/v1/databases`, ensure the token it uses has the `admin` permission. Auto-create-on-write (databases that come into existence as a side-effect of line-protocol or msgpack writes) is unaffected — write tokens continue to work for ingestion.
-5. **CSV/Parquet import callers**: imports that relied on a Parquet `DECIMAL` column round-tripping as a decimal will now receive a `DOUBLE`. Imports of malformed files (empty, duplicate headers, or a `time_column` rename that collides with an existing `time` column) now fail fast with a `400` instead of partially succeeding — review any automation that ignored import error responses.
+5. **CSV/Parquet import callers**: imports that relied on a Parquet `DECIMAL` column round-tripping as a decimal will now receive a `DOUBLE`. Malformed files now fail fast with a `400` instead of partially succeeding (empty file, empty/blank/duplicate column names, a `time_column` rename that collides with an existing `time` column, or a `NaN`/`Inf` float time value) — review any automation that ignored import error responses.
 
 ## Dependencies
 
--- a/cmd/arc/main.go
+++ b/cmd/arc/main.go
@@ -1850,9 +1850,12 @@ func main() {
 	tleHandler.RegisterRoutes(server.GetApp())
 
 	// Register Import handler (CSV, Parquet, Line Protocol, TLE bulk import).
-	// uploadDir is the same path that database.New added to the DuckDB
-	// sandbox's allowed_directories — staying in sync keeps imports working.
-	importHandler := api.NewImportHandler(db, storageBackend, dbConfig.UploadDir, logger.Get("import"))
+	// All formats parse in-process and ingest through the ArrowBuffer pipeline;
+	// no import path issues DuckDB queries against the uploaded file, so the
+	// handler no longer needs the sandbox-allowlisted upload directory.
+	// (dbConfig.UploadDir is still allowlisted for delete.go's S3 COPY ... TO
+	// staging — see NewDeleteHandler below.)
+	importHandler := api.NewImportHandler(logger.Get("import"))
 	importHandler.SetArrowBuffer(arrowBuffer)
 	if authManager != nil && rbacManager != nil {
 		importHandler.SetAuthAndRBAC(authManager, rbacManager)
--- a/internal/api/import.go
+++ b/internal/api/import.go
@@ -1,19 +1,13 @@
 package api
 
 import (
-	"context"
 	"fmt"
 	"io"
-	"os"
-	"path/filepath"
-	"strings"
 	"sync/atomic"
 	"time"
 
 	"github.com/basekick-labs/arc/internal/auth"
-	"github.com/basekick-labs/arc/internal/database"
 	"github.com/basekick-labs/arc/internal/ingest"
-	"github.com/basekick-labs/arc/internal/storage"
 	"github.com/basekick-labs/arc/pkg/models"
 	"github.com/gofiber/fiber/v2"
 	"github.com/rs/zerolog"
@@ -50,13 +44,17 @@ type TLEImportResult struct {
 	DurationMs     int64    `json:"duration_ms"`
 }
 
-// ImportHandler handles bulk CSV, Parquet, and Line Protocol file imports
+// ImportHandler handles bulk CSV, Parquet, Line Protocol, and TLE file imports.
+//
+// All formats parse rows in-process and ingest through the ArrowBuffer pipeline
+// (the same path as streaming writes). No import path issues DuckDB queries
+// against the uploaded file, so imports do not require the upload directory to
+// be in the DuckDB sandbox allowlist. (The allowlist entry that still exists is
+// retained solely for delete.go's S3 COPY ... TO staging — see cmd/arc/main.go.)
 type ImportHandler struct {
-	db      *database.DuckDB
-	storage storage.Backend
-	logger  zerolog.Logger
+	logger zerolog.Logger
 
-	// ArrowBuffer for LP import (uses the streaming ingest pipeline)
+	// arrowBuffer is the streaming ingest pipeline used by every import format.
 	arrowBuffer *ingest.ArrowBuffer
 
 	// authManager holds the concrete *auth.AuthManager. See
@@ -65,37 +63,17 @@ type ImportHandler struct {
 	authManager *auth.AuthManager
 	rbacManager RBACChecker
 
-	// uploadDir is the directory under which multipart uploads land. Must
-	// be one of the prefixes in the DuckDB sandbox's allowed_directories
-	// list, otherwise read_csv/read_parquet on the uploaded file fails.
-	// Empty means use the OS default (only safe before the sandbox lockdown
-	// was introduced — kept for legacy/test paths that don't go through
-	// main.go).
-	uploadDir string
-
 	// Stats
 	totalRequests atomic.Int64
 	totalRecords  atomic.Int64
 	totalErrors   atomic.Int64
 }
 
-// NewImportHandler creates a new ImportHandler. uploadDir must be the same
-// path cmd/arc/main.go added to the DuckDB sandbox's allowed_directories,
-// otherwise read_csv/read_parquet on the uploaded file fails post-lockdown.
-// Logs a Warn on empty uploadDir; the request handler also fails-closed at
-// every POST so misconfiguration surfaces both at startup and at the first
-// import attempt. Tests that exercise the handler should pass a non-empty
-// path (typically a t.TempDir() inside LocalStorageRoot).
-func NewImportHandler(db *database.DuckDB, storage storage.Backend, uploadDir string, logger zerolog.Logger) *ImportHandler {
-	componentLogger := logger.With().Str("component", "import-handler").Logger()
-	if uploadDir == "" {
-		componentLogger.Warn().Msg("NewImportHandler called with empty uploadDir — imports will fail under the DuckDB sandbox; cmd/arc/main.go should pass the sandbox-allowlisted upload directory")
-	}
+// NewImportHandler creates a new ImportHandler. The ArrowBuffer must be set via
+// SetArrowBuffer before any import is served.
+func NewImportHandler(logger zerolog.Logger) *ImportHandler {
 	return &ImportHandler{
-		db:        db,
-		storage:   storage,
-		uploadDir: uploadDir,
-		logger:    componentLogger,
+		logger: logger.With().Str("component", "import-handler").Logger(),
 	}
 }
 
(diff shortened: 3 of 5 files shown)


xe-nvdk:
Thanks @gemini-code-assist — addressed in 4acfbc1:

- **`isBoolLiteral` dead code (high):** fixed — now accepts `"1"`/`"0"`, so mixed boolean representations like `["true","1"]` infer as boolean. (Pure `0/1` columns still infer as integer, since `isInt` wins over `isBool` in the switch — that's intentional.)
- **Missing integer types (high):** fixed — `arrowColumnToInterfaces` and the Parquet time-column path now handle `Int8/Int16` and `Uint8/16/32/64` (all widen to `int64`). Verified end-to-end importing a Parquet with UTINYINT/USMALLINT/UBIGINT/SMALLINT columns.
- **Per-row layout rescan (medium):** fixed — `stringsToTimeMicros` caches the matched timestamp layout and tries it first on subsequent rows.

**Declined: `WriteColumnarRecord` → `WriteTypedColumnarDirect` (the double-conversion comments).** The allocation cost is real, but imports are one-shot bulk operations, not the streaming hot path, so the per-import boxing cost isn't latency-critical. More importantly, `WriteTypedColumnarDirect` bypasses `convertColumnsToTyped`, which is where type coercion, validity bitmaps, and per-measurement decimal handling live — hand-rolling that for the import path adds correctness risk for no hot-path gain. This is deliberately the same call the Line Protocol and TLE import handlers use. If import throughput becomes a measured bottleneck, the typed path is a clean follow-up.

xe-nvdk:
Full pass over all review comments (two rounds, 11 inline). Status:

**Fixed in 4acfbc1:**
- **`isBoolLiteral` dead code** (`:555`) — now accepts `"1"`/`"0"`; mixed reps like `["true","1"]` infer as boolean. Pure `0/1` stays integer (intentional — `isInt` precedes `isBool`).
- **Missing integer types** (`:693`) — `arrowColumnToInterfaces` + the Parquet time-column path handle `Int8/Int16` and `Uint8/16/32/64`. Verified importing a Parquet with UTINYINT/USMALLINT/UBIGINT/SMALLINT.
- **Per-row layout rescan** (`:602`) — `stringsToTimeMicros` caches the matched layout and tries it first on later rows.

**Fixed in 08b1990:**
- **Negative-epoch misclassification** (`:655`) — good catch. `autoEpochToMicros` now magnitude-detects on `abs(n)`, so pre-1970 negative milli/micro/nano epochs resolve to the right unit instead of always hitting the seconds branch. Added regression tests.

**Declined (with reasoning): the `WriteTypedColumnarDirect` / zero-boxing cluster** (`:203`, `:343`, `:546`, `:759`, `:18`, and the dependent test rewrite `import_test.go:34`).

These six comments are one interdependent suggestion: have `inferAndConvertColumn` / a new `arrowColumnToTyped` return typed slices + validity bitmaps and call `WriteTypedColumnarDirect` to skip the `[]interface{}` round-trip. I'm not taking it in this PR:

1. **Not the hot path.** Imports are one-shot bulk operations; the boxing cost is paid once per import, not per write/sec. There's no measured latency problem to solve here, and our convention is to benchmark before accepting a perf change — I haven't, and this PR's goal is correctness parity with the LP/TLE handlers.
2. **`WriteTypedColumnarDirect` bypasses `convertColumnsToTyped`**, which is where type coercion, schema-churn detection, and per-measurement decimal handling live. Hand-rolling the typed batch + validity contract for the import path trades real correctness surface for a non-hot-path allocation win.
3. CSV/Parquet import deliberately uses the **same `WriteColumnarRecord` call as Line Protocol import** — keeping them on one code path is the point of #472.

If import throughput is ever measured as a bottleneck, the typed path is a clean, self-contained follow-up (and the same optimization would apply to the LP importer too).

xe-nvdk:
Round 3 — addressed in ed858ff:

- **`io.LimitReader` silent truncation** (`:89`/`:158`/`:855`, HIGH) — real bug, good catch. `io.LimitReader` returns `io.EOF` at the cap, so `csv.Reader` would treat an oversized upload as a clean EOF and import a **truncated** file with `200 OK`. Replaced with a `limitedReader` that returns an explicit `errImportTooLarge` once the cap is passed; `importCSV` detects it via `errors.Is` (csv wraps the underlying error) and returns `413`. Added a unit test covering both under- and over-limit.
- **`math.MinInt64` in `autoEpochToMicros`** (`:655`/`:12`, MEDIUM) — fixed; negating `MinInt64` overflows, so it's clamped to `MaxInt64` for the magnitude check. `math` imported.
- **Nil guards** (`:683`/`:772`, MEDIUM) — added `col == nil || col.Data() == nil` checks to both Arrow conversion helpers.

The `WriteTypedColumnarDirect` zero-boxing cluster remains declined for the reasons in my previous comment (imports are one-shot, not the hot path; bypassing `convertColumnsToTyped` adds correctness surface; same call as the LP/TLE importers). Happy to take it as a follow-up if import throughput is ever measured as a bottleneck.

xe-nvdk:

xe-nvdk:
Round 4 — addressed in e69d5c4. Accepted the float-epoch support:

- `epochToMicros` / `autoEpochToMicros` now take `float64`, so fractional epochs (e.g. `1609459200.123`) parse correctly with sub-second precision instead of being rejected by `strconv.ParseInt`. Both the explicit `epoch_*` and auto paths use `strconv.ParseFloat` with `NaN`/`Inf` guards. `intTimeToMicros` casts to `float64`.
- Bonus: the `float64` magnitude check drops the `math.MinInt64` abs special-case entirely.
- Tests: float `epoch_s`/`epoch_ms`, auto float, and `NaN`/`Inf` rejection. Verified end-to-end — `1609459200.123` with `epoch_s` imports as `2021-01-01T00:00:00.123Z` and round-trips through a query.

Good catch on the regression — the old DuckDB path tolerated float epochs and the in-process rewrite had narrowed that to integers only.

xe-nvdk:

xe-nvdk:
Round 5 — addressed in a226889. Accepted:

- `arrowColumnToInterfaces` and the Parquet time-column path now handle `*array.Binary` and `*array.FixedSizeBinary` (BYTE_ARRAY columns without a UTF8 logical annotation, common from Spark / some writers) by converting the bytes to a Go string, instead of failing with `unsupported arrow type binary`.
- Verified by importing a DuckDB `BLOB` column (BYTE_ARRAY physical type) — round-trips correctly as VARCHAR.

xe-nvdk:

xe-nvdk:
Round 6 — addressed in 21beb56. All accepted:

- **Integer-epoch precision** (`:609`/`:635`/`:671`) — good catch; round 4 over-corrected by routing *all* epochs through `float64`, which loses precision above 2^53 (e.g. ns epoch `1609459200000001900`). Now: no decimal point → `int64` math (`intTimeToMicros` / `autoIntEpochToMicros`); decimal point → `float64` for fractional epochs. Regression test asserts the large-ns case stays exact.
- **Float64/Float32 Parquet time columns** (`:888`) — `parquetColumnToTimeMicros` now handles them via `floatTimeToMicros`. Verified a DuckDB `DOUBLE` time column (fractional epoch) imports with sub-second precision (`.123Z`/`.5Z`).
- **All-empty CSV column** (`:520`) — now defaults to `string` instead of inferring `int64`, avoiding cross-file schema mismatch. Test added.

The `WriteTypedColumnarDirect` zero-boxing cluster remains the only declined item (non-hot-path; bypasses `convertColumnsToTyped`).

xe-nvdk:

xe-nvdk:
Round 7 — addressed in 2fd55c1. All accepted, and the critical one is a genuine catch:

- **CRITICAL/DoS — empty column name** (`:423`): confirmed real. A trailing comma in a CSV header (`time,value,`) produced an empty column name that passed validation, then panicked the server at `name[0]` in the Arrow schema builder (`arrow_writer.go:393/455` index without a length guard). `validateImportHeader` now rejects empty column names with `400`. Verified end-to-end: the trailing-comma CSV returns `400 "column name cannot be empty"` and the server stays up (previously crashed).
- **HIGH — NaN/Inf float time** (`:897`): `Float64`/`Float32` Parquet time columns with NaN/Inf were silently converted to garbage micros (corrupt partitions). Now rejected with `400`, matching the validation in `oneTimeValueToMicros`. Verified a NaN DOUBLE time column returns `400`, server alive.
- **MEDIUM — eager `string(a.Value(i))` on null rows** (`:792`): switched Binary/FixedSizeBinary to an explicit branch so the conversion is skipped for nulls.

Full suite (incl. `-race`), `go vet`, `gofmt` all clean. Note: this round followed a full internal review pass (config matrix + holistic re-read) on my side, which separately caught a wrong test assertion and a Parquet double-conversion — both fixed in e2c7b84.

xe-nvdk:

xe-nvdk:
Round 8 — accepted the `WriteTypedColumnarDirect` zero-boxing path for Parquet, in 00b51b8. I'd declined this in earlier rounds pending a measurement; I benchmarked it (1M rows × 4 cols):

| path | time | mem | allocs |
|---|---|---|---|
| `arrowColumnToInterfaces` (boxing) | 41.9 ms | 104 MB | 3,999,746 |
| `arrowColumnToTyped` (typed) | 3.3 ms | 40 MB | 8 |

~13× faster, 2.6× less memory, ~500,000× fewer allocations — and a lower bound, since the boxing path additionally re-typed everything in `convertColumnsToTyped`. The OOM/GC concern on large imports is real, so it's worth it.

`importParquet` now extracts flat typed slices + a null-validity bitmap directly from the Arrow chunks (`arrowColumnToTyped`) and ingests via `WriteTypedColumnarDirect` (same call the TLE importer uses). Verified end-to-end with a Parquet covering timestamp/int-widths/float/decimal/string/blob/bool **plus a null float** — the null round-trips correctly, confirming the validity-bitmap path. Full suite + `-race` green.

CSV still uses `WriteColumnarRecord` (its values come from string type-inference, not Arrow arrays — converting it is a separate change; happy to do it as a follow-up if worth it).

xe-nvdk:

xe-nvdk:
Round 9 — in f8affdc. Two accepted, one rejected with evidence:

- **`LazyQuotes=true`** (`:112`) — accepted; a field with an embedded unescaped quote now imports.
- **Skip ParseFloat while still int** (`:517`) — accepted; every base-10 integer is a valid float, so float parsing only runs once a value disqualifies the int type.
- **Allocation-free `isBoolLiteral`** (`:588`) — accepted; `EqualFold` + direct compares, no `ToLower` allocation.

- **`ReuseRecord=true`** (`:112`) — **rejected, it introduces data corruption here.** `ReuseRecord` overwrites the backing array of the previously returned slice, and this handler retains the `header` slice across every subsequent `Read()` (for `validateImportHeader`, column naming, and `ImportResult.Columns`). I tried it: a 4-row CSV came back with `columns: ["2021-01-01T00:00:03Z","delta","40"]` — the header clobbered by the last row. Reverted and left a comment so it isn't re-suggested. (`ReuseRecord` would only be safe if we deep-copied the header up front; not worth it for a one-shot import.)

Verified: multi-row CSV with distinct per-row values round-trips intact; full suite + `-race` green.

xe-nvdk:

xe-nvdk:
Round 10 — fixed in 3067708. Good catch: this was a regression my round-9 change introduced. When I added the `isFloat && !isInt` float-skip, I dropped the existing early `break`, so a text column kept scanning every remaining row after all type flags went false. Restored the break.

Safe to break there: `hasValue` is already true at that point, and string columns ignore `hasEmpty` (empty cells are valid `""` values, not nulls), so stopping early can't change the inferred type or the result. Verified the `mixed -> strings` and `int then float -> float` cases still pass; full suite + `-race` green.

xe-nvdk:

xe-nvdk:
**Gemini (BOM, `:144`):** accepted in a685f7d. Excel-on-Windows CSVs prepend a UTF-8 BOM, which made the first column name parse as `\xef\xbb\xbf<name>` so `time_column` wasn't found → 400. Now stripped from the first header field; verified a BOM-prefixed CSV imports with the header parsed correctly.

---

**Also ran a second independent review (DeepSeek v4 Pro).** Dispositions:

- **B1** (bool/int inference misclassification): the reviewer traced it, then self-retracted after re-tracing — confirmed no bug (`isInt` wins in the switch; `isBool` only wins when a non-numeric value is present, in which case every value is a checked bool literal). Agreed, no change.
- **B2** (LP/TLE use unbounded `io.ReadAll` before the size check — pre-existing DoS): **real, but pre-existing and out of scope for this PR**, which only touches CSV/Parquet. Tracking as a follow-up to add `limitedReader` to the LP/TLE handlers (the same fix this PR introduced for CSV/Parquet).
- **B3** (`limitedReader` returns bytes+error together): the reviewer rates it "acceptable defense-in-depth tradeoff" — the error is checked right after `Read()`, so the import still fails cleanly. No change.
- **P1** (type inference double-parses; single-pass-with-discard could avoid it), **P2/P3/P4** (prealloc sizing, Arrow-table 2x peak memory, vectorization): genuine perf opportunities, but optimizations rather than correctness, and the zero-boxing typed path already landed the big win (~13× measured). Deferring to a follow-up rather than expanding this already-large PR; would bench before adopting P1.
- **S1** (nil-guard message wording differs between CSV/Parquet and LP/TLE): leaving as-is — the CSV/Parquet messages name the actual field, which is arguably clearer; not worth churning. **S3** (duplicate `maxImportSize`): LP/TLE define their own locals; consolidating touches those handlers — folding into the B2 follow-up. **S2** (`fiberContext` alias): cosmetic, leaving.

Net from the DeepSeek pass: 0 new in-scope correctness bugs; B2 + the perf items become a tracked follow-up. The BOM fix above was the only actionable correctness item across both reviews this round.

xe-nvdk:
Reconsidered **DeepSeek B2** (LP/TLE unbounded `io.ReadAll`) and fixed it here rather than deferring — it's the same import-size-limiting concern this PR introduced for CSV/Parquet, in the same file, so finishing it in-scope is the right call (commit 3fcd4b7).

Context that adjusts the severity: a global Fiber `BodyLimit` (default 1GB) already backstops these reads, so it was never *truly* unbounded — but LP/TLE could still buffer up to 1GB before the 500MB check fired, 2× the cap the CSV/Parquet handlers now enforce. Both LP and TLE now wrap the read in `io.LimitReader(file, maxImportSize+1)` and reject overflow with 413, matching CSV/Parquet. The post-decompress check still guards gzip that expands past the limit. Verified LP import still works end-to-end.

Remaining DeepSeek items (perf P1 double-parse, P3 Arrow 2× peak memory, S3 `maxImportSize` consolidation) stay as a tracked follow-up — they're optimizations/cleanup, not correctness, and the big perf win (zero-boxing) already landed.

PR #476, opened 2026-06-03
perf(import): single-parse CSV inference + buffer pre-sizing (follow-up to #475)

xe-nvdk:
Follow-up to #475, addressing the performance items from the DeepSeek v4 Pro review.

## P1 — single-parse CSV type inference (highest impact)
`inferAndConvertColumn` parsed every numeric value **twice**: once during type inference (`ParseInt`/`ParseFloat`), once during the conversion pass. It now accumulates parsed values into a typed buffer **during** the inference scan, so a homogeneous int/float column is parsed once. On an int→float demotion the already-parsed ints are migrated into the float buffer; on a demotion to string the buffers are abandoned. Both buffers are lazily allocated, so string columns don't pay for them.

Benchmark (`BenchmarkInferAndConvert`, 1M-row column):

| column | before | after |
|---|---|---|
| int | 24.8 ms | 14.4 ms (~1.7×) |
| float | 62.3 ms | 32.1 ms (~1.9×) |
| string | 3.1 ms | 3.2 ms (no regression) |

## P2 — pre-size per-column buffers
`rawCols[i]` was allocated with a fixed `cap=1024`, forcing ~14 doublings on a 10M-row file. Now pre-sized from an upload-size row estimate (`fileHeader.Size / 80`, floored at 1024).

## S3 — dedup `maxImportSize`
Dropped the duplicate local `maxImportSize` consts in the LP/TLE handlers; they now use the single package-level const (same value).

## Skipped (documented non-goals)
- **P3** (release the Arrow table incrementally to cut Parquet ~2× peak memory): needs an arrow-go ownership / record-batch-streaming rewrite with real use-after-free risk, for bounded benefit (imports cap at 500 MB). Not worth it in this pass.
- **P4** (unsafe slice casting for integer widening): wants `unsafe` in the import path for a marginal, hard-to-measure gain. Declined.

## Test plan
- [x] `go build ./... && go vet ./internal/api/... && gofmt -l` clean
- [x] `go test -race ./internal/api/...` green
- [x] New `inferAndConvertColumn` tests for the lazy-alloc/demotion edges (leading-empty-then-int, leading-empty-then-float-demotion) on top of the existing cases
- [x] Internal deep review of the rewritten function (demotion correctness, nil-buffer indexing, validity alignment, bool/int precedence, early-break) — no bugs
- [x] Integration: CSV with int + int→float-demotion + bool(`true`/`false`/`1`) + null + all-empty columns round-trips with correct types (`BIGINT`/`DOUBLE`/`BOOLEAN`) and the null preserved; LP import still works

### DIFF
--- a/RELEASE_NOTES_2026.06.2.md
+++ b/RELEASE_NOTES_2026.06.2.md
@@ -24,6 +24,8 @@ User-visible behavior changes:
 - CSV uploads are now subject to the same 500 MB size cap already enforced for the other import formats.
 - Parquet `DECIMAL` columns are imported as `DOUBLE` (Arc's ingest path does not carry per-column decimal precision for imports). Use Line Protocol with a configured decimal column if exact decimal precision is required.
 
+Performance: CSV imports now parse each numeric value once (the type-inference and conversion passes were merged), roughly halving CPU on numeric-heavy files; per-column buffers are pre-sized from the upload size to avoid repeated reallocation on large files. No behavior change.
+
 Line Protocol and TLE imports are unchanged. CSV/Parquet imports no longer depend on the DuckDB sandbox's allowed-directories list.
 
 ## Upgrade notes
--- a/internal/api/import.go
+++ b/internal/api/import.go
@@ -172,12 +172,10 @@ func (h *ImportHandler) handleLineProtocolImport(c *fiber.Ctx) error {
 	}
 	defer file.Close()
 
-	// Size limit for bulk LP import (applies to both compressed and uncompressed)
-	const maxImportSize = 500 * 1024 * 1024 // 500MB
-
-	// Bound the read at the limit (+1 to detect overflow) so an oversized upload
-	// is rejected without buffering the whole body — the global BodyLimit (1GB)
-	// is a coarser backstop; this matches the CSV/Parquet handlers.
+	// Size limit (maxImportSize, package-level) applies to both compressed and
+	// uncompressed data. Bound the read at the limit (+1 to detect overflow) so an
+	// oversized upload is rejected without buffering the whole body — the global
+	// BodyLimit (1GB) is a coarser backstop; this matches the CSV/Parquet handlers.
 	data, err := io.ReadAll(io.LimitReader(file, maxImportSize+1))
 	if err != nil {
 		h.totalErrors.Add(1)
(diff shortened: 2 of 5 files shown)


xe-nvdk:
Addressed in 10f81e3.

**Gemini (HIGH, `:99`/`:118`/`:168` — wide-CSV OOM from `fileHeader.Size/80`):** correct and important — a size-only row estimate over-allocates badly on wide files (500MB/100-col → ~6.25M rows × 100 slices ≈ 10GB of backing arrays). Fixed exactly as suggested: `importCSV` now takes `fileSize int64` and computes the estimate **after** reading the header, dividing by row width (`8 × len(header)` bytes/row) so per-column capacity scales *down* with column count. Added a 1M cap as a secondary guard. Verified a 50-column CSV imports correctly (`c48` → `[48, 96]`); full suite + `-race` green.

**DeepSeek Q1 (string benchmark misleading):** also correct — a text column breaks type inference on row 0, so the old `_String` benchmark measured the early-break + copy path, not a full scan. Renamed to `_StringEarlyBreak` and added `_LateStringDemotion` (numeric until the final row) to measure the optimization's genuine worst case:

| benchmark | time | mem |
|---|---|---|
| Int | 13.8 ms | 8 MB |
| Float | 33.2 ms | 8 MB |
| StringEarlyBreak | 4.7 ms | 16 MB |
| LateStringDemotion (full parse, then abandon) | 17.7 ms | 24 MB |

DeepSeek's other items (Q2 coverage adequate, Q3 late-demotion garbage — now quantified by the new benchmark, ~8MB abandoned in the pathological case) are acknowledged non-issues; it found **0 bugs**. The OOM finding was the only actionable correctness item this round, and it's fixed.

xe-nvdk:

xe-nvdk:
Round addressed in b9cfd89.

**Gemini MEDIUM (`:167`, 32-bit overflow):** fixed — `bytesPerCell * len(header)` was multiplied as a 32-bit `int` before the int64 conversion, so a pathological column count could overflow on 32-bit arches. Now multiplies both operands as `int64`. (The division-by-zero half of the concern isn't reachable — the empty-header case returns earlier with "file is empty" — added a comment noting the invariant.)

The earlier `:99`/`:118`/`:168` HIGH comments are the wide-CSV OOM finding already addressed in 10f81e3 (re-shown because they're anchored to the prior commit). Full suite + vet clean.

PR #477, opened 2026-06-03
feat(compaction): use batch-delete APIs for compaction cleanup on S3/Azure

### LINKED ISSUE #400: Batched compaction-delete on S3/Azure backends
## Problem

Compaction's cleanup path (\`Job.deleteOldFiles\` in \`internal/compaction/job.go\`) iterates \`j.compactedFiles\` and calls \`StorageBackend.Delete(ctx, fileKey)\` per file. On large compaction cycles (hundreds of source files), this is one S3/Azure API call per file — wasteful and slow.

Both S3 and Azure expose batch-delete APIs:
- S3: \`DeleteObjects\` — up to 1000 keys per call
- Azure Blob: \`BlobBatch\` — up to 256 deletes per batch

## Scope

**Only the compaction cleanup path** (\`Job.deleteOldFiles\`). The cluster Raft manifest has already committed the logical delete via \`CommandDeleteFile\`; this is pure physical cleanup.

**Not affected:**
- User-facing \`DELETE\` API (keeps all existing guardrails: enabled flag, row limits, confirmation, dry run)
- Retention policies (separate code path, different semantics)
- Local storage backend (just a loop — no batch equivalent needed)

## Proposed Solution

1. Add \`DeleteBatch(ctx context.Context, keys []string) error\` to \`StorageBackend\` interface.
2. S3 backend: use \`s3.DeleteObjects\` with 1000-key chunks.
3. Azure backend: use \`BlobBatch\` with 256-key chunks.
4. Local backend: default implementation = loop over \`Delete\`.
5. \`Job.deleteOldFiles\` calls \`DeleteBatch(ctx, j.compactedFiles)\`.

## Impact

- Reduces S3 DELETE API calls by up to 1000× on large compactions
- Lowers latency of the compaction cleanup phase proportionally
- No behavior change from user's perspective — logical deletes still go through Raft

xe-nvdk:
## Summary

Closes #400.

Compaction's cleanup path (`Job.deleteOldFiles`) was calling `StorageBackend.Delete` once per compacted source file. The `BatchDeleter` interface already existed in all three backends, but the compaction path never used it. Azure's `DeleteBatch` was also a fake (per-file loop).

## Changes

1. **`internal/storage/azure.go`** — Rewrote `DeleteBatch` to use the actual Azure SDK BlobBatch API (`container.Client.NewBatchBuilder()` + `SubmitBatch()`), chunked at 256. Added proper 404-vs-other-error discrimination for sub-request failures.

2. **`internal/storage/s3.go`** — Added inspection of `DeleteObjectsOutput.Errors` for per-object failures. Treats `NoSuchKey` as success.

3. **`internal/compaction/job.go`** — `deleteOldFiles` now prefers `BatchDeleter`, falling back to per-file `Delete`.

4. **`RELEASE_NOTES_2026.06.2.md`** — Added performance improvements section.

## Test Plan

- [x] `go build` — clean
- [x] `go test ./internal/storage/...` — PASS
- [x] `go test ./internal/compaction/...` — PASS
- [x] `gofmt` — clean
- [x] `go vet` — clean
- [x] Internal deep review with configuration matrix — findings addressed

### DIFF
--- a/RELEASE_NOTES_2026.06.2.md
+++ b/RELEASE_NOTES_2026.06.2.md
@@ -28,6 +28,12 @@ Performance: CSV imports now parse each numeric value once (the type-inference a
 
 Line Protocol and TLE imports are unchanged. CSV/Parquet imports no longer depend on the DuckDB sandbox's allowed-directories list.
 
+## Performance improvements
+
+**Compaction cleanup now uses batch-delete APIs on S3 and Azure.** `deleteOldFiles` previously called `StorageBackend.Delete` once per compacted source file — on large compaction cycles (hundreds of files), this produced hundreds of sequential S3/Azure API calls. Both cloud backends already implemented `DeleteBatch` (S3 `DeleteObjects`, Azure `BlobBatch`) but the compaction cleanup path never used it. The path now prefers `BatchDeleter.DeleteBatch`, falling back to per-file `Delete` if the backend does not support batching. This reduces S3 DELETE API calls by up to 1000× and Azure calls by up to 256× on large compactions, with a proportional drop in cleanup-phase latency. Local-storage deployments are unaffected (a per-file loop is the correct implementation for a local filesystem).
+
+Additionally, the Azure `DeleteBatch` implementation was rewritten to use the actual Azure SDK `BlobBatch` API (`container.Client.NewBatchBuilder()` + `SubmitBatch`) instead of a per-file loop, bringing it into parity with the S3 implementation.
+
 ## Upgrade notes
 
 1. **No configuration change required.** Drop in the new binary; existing `arc.toml` and license keys work as-is.
--- a/internal/compaction/job.go
+++ b/internal/compaction/job.go
@@ -719,6 +719,10 @@ func (j *Job) uploadFile(ctx context.Context, localPath, key string) error {
 // This ensures we don't delete files that were skipped due to corruption or other issues.
 // Prefers batch delete (BatchDeleter interface) when the backend supports it
 // (S3 DeleteObjects, Azure BlobBatch) to reduce API call overhead.
+// Does NOT fall back to per-file delete on batch failure — a failing batch
+// (e.g. auth error, rate limit) means individual calls will also fail, and
+// the per-file loop would cause severe latency spikes. The next compaction
+// cycle will retry.
 func (j *Job) deleteOldFiles(ctx context.Context) error {
 	if len(j.compactedFiles) == 0 {
 		j.logger.Debug().Msg("No files to delete (none were compacted)")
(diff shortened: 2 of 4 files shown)


xe-nvdk:

xe-nvdk (commented):
Reviewed locally (checked out the branch, traced the orchestration + both cloud SDKs, ran build/vet/test on `internal/storage` and `internal/compaction` — all clean). **Verdict: correct and safe to merge.** No blockers, no high-severity issues. A couple of small cleanups below.

## What's right (the parts that matter)

- **Manifest-before-storage ordering preserved.** The change is entirely inside `deleteOldFiles` (storage-only cleanup); it doesn't touch the Raft manifest. In cluster mode the `output_written` completion manifest is written at `job.go:372-377` *before* `deleteOldFiles` at :380, and on batch failure :381 logs without failing the job and retains the manifest for next-cycle retry. The Cluster Operations invariant holds.
- **Partial-failure semantics are safe and idempotent.** S3/Azure collect per-object sub-errors and return a joined error; `deleteOldFiles` surfaces it as a non-fatal warning and lets the next cycle retry. Re-deleting already-gone keys is success on both backends (S3 reports them under `Deleted`; Azure 404 → treated as success via `isAzureNotFoundError`), so retries don't hard-fail.
- **No per-file fallback on batch error** is the right call (a failing batch — auth/rate-limit — means per-file would fail too, just slower). Matches the checklist's "no latency-spike fallback."
- **Azure BlobBatch usage is correct:** `NewBatchBuilder` + `bb.Delete(path,nil)` + `SubmitBatch`; per-item failures come back as `err==nil` with populated `item.Error` (a `*azcore.ResponseError`, which `isAzureNotFoundError` classifies correctly); top-level err is transport/auth and correctly fatal; 256 chunk size matches the Azure limit. `prefixedKey` usage is consistent between single `Delete` and `DeleteBatch` on both backends (no orphan-key risk).

## Medium — worth fixing before merge

1. **S3 `NoSuchKey` branch is effectively dead code** (`s3.go:456`). `DeleteObjects` is idempotent: a missing key is reported under `Deleted`, never `Errors` (and with `Quiet=true` it's suppressed entirely). So the `if *e.Code == "NoSuchKey"` branch never fires for the normal already-deleted case. It's harmless (real error codes like `AccessDenied`/`InternalError` are still collected) and retry idempotency is unaffected — but the comment at :452-454 describes a path S3 doesn't exercise. Either drop the branch or reword the comment to "defensive."

## Nit

2. **`deleteOldFiles` doc-comment is slightly self-contradictory** (`job.go:721-724`). It says "Does NOT fall back to per-file delete on batch failure," but the code *does* use per-file `Delete` when the backend lacks `BatchDeleter`. The comment means "a batch that *returns an error* isn't retried per-file" — reword to that to avoid conflating it with the no-`BatchDeleter` fallback path.

## Doc accuracy note

The release-notes line "Local-storage deployments are unaffected (a per-file loop is the correct implementation)" is true in effect, but note `LocalBackend` *does* implement `BatchDeleter` (`local.go:552`), so local deployments take the batch branch — which internally loops `Delete` + `errors.Join`. Behavior is identical, just reached via the batch path. The per-file fallback in `deleteOldFiles` is currently unreachable by any shipping backend (all three implement `BatchDeleter`); fine as defensive code, just worth knowing it's not exercised.

Build/vet/gofmt/tests clean. Once the two cleanups above are in, this is good to go.

xe-nvdk:
Re the two @gemini-code-assist findings:

**`azure.go` — "`BatchResponseItem` does not have a `BlobName` field" (HIGH): incorrect.** This is reasoning about a different/older SDK surface. We vendor `github.com/Azure/azure-sdk-for-go/sdk/storage/azblob v1.6.3` (go.mod), where the type is:

```go
// azblob@v1.6.3/internal/exported/blob_batch.go:151
type BatchResponseItem struct {
    ContentID     *int
    ContainerName *string
    BlobName      *string   // <-- exists
    RequestID     *string
    Version       *string
    Error         error
}
```

(`container.BatchResponseItem` is a type alias to this — `container/responses.go:66`; the SDK's own `container/examples_test.go:455` prints `*resp.BlobName`.) `item.BlobName` compiles and is correct — `go build ./internal/storage/` passes clean. No change needed; the index-based workaround would be a downgrade (the SDK gives us the authoritative blob name directly).

**`job.go:744` — "don't fall back to per-file on batch failure" (HIGH): already implemented.** The current code returns the batch error directly and does NOT loop per-file on failure — the per-file branch is only reached when the backend doesn't implement `BatchDeleter` at all. Gemini's suggested snippet is identical to what's already there, so no change. (Worth a doc-comment tweak to make that distinction clearer — see my review's nit #2.)

Verified both against the vendored SDK source rather than at face value.

PR #478, opened 2026-06-03
feat(cluster): paginated FSM manifest walks for catch-up and API

### LINKED ISSUE #396: Paginated FSM manifest walks
## Problem
\`fsm.GetAllFiles()\` allocates a full O(N) snapshot copy of the manifest. For 1M+ files this causes ~50ms apply-path spikes and ~50MB transient memory.

## Proposed Solution
Add a paginated or streaming variant for the catch-up walker and API handler. The FSM keeps the full map internally but exposes an iterator or cursor-based API.

## Context
Deferred from Phase 3. Emergency kill-switch exists (\`replication_catchup_enabled=false\`).

## Effort
Medium

xe-nvdk:
## Summary

Closes #396.

`ClusterFSM.GetAllFiles()` previously allocated a full O(N) copy of the file manifest — for 1M+ files this caused ~50ms apply-path spikes (RLock held while copying) and ~50MB transient memory. This PR adds cursor-based pagination.

## Changes

1. **`internal/cluster/raft/fsm.go`** — Added `keysCache []string` + `keysCacheGen int64` fields with cache invalidation on register/delete/update/batch/Restore. Added `GetFilesPaginated(cursor, limit)` using sorted-key pagination with RLock released between pages.

2. **`internal/cluster/filereplication/catchup.go`** — `RunCatchUp` now accepts a paginated fetch function `func(cursor, limit) ([]*FileEntry, string, error)` instead of `[]*FileEntry`. Pages through the manifest at 1000 entries/page with backpressure between pages.

3. **`internal/cluster/coordinator.go`** — Added `GetFileManifestPaginated`. Catch-up now uses paginated fetch closure.

4. **`internal/api/cluster.go`** — `GET /api/v1/cluster/files` supports `?cursor=<path>&limit=<n>` for pagination. Omitting both params returns all files (backward-compatible).

5. **`RELEASE_NOTES_2026.06.2.md`** — Added paginated manifest walks section.

## Internal Review Findings Addressed

- **Blocker**: `Restore()` now invalidates `keysCache` after snapshot load
- **High**: Added comment explaining why capturing `fsm` in closure is safe
- **Medium**: Clarified cursor stability docs — newly added entries sorting before cursor may be missed (acceptable; catch-up also receives reactive FSM callbacks)
- **Style**: Added `nextCursor` semantics comment in `paginateSlice`

## Test Plan

- [x] `go build ./internal/cluster/... ./internal/api/...` — clean
- [x] `go vet` — clean
- [x] `go test ./internal/cluster/...` — all 7 packages PASS
- [x] `gofmt -l` — clean
- [x] Internal review with configuration matrix — findings addressed

### DIFF
--- a/RELEASE_NOTES_2026.06.2.md
+++ b/RELEASE_NOTES_2026.06.2.md
@@ -34,6 +34,14 @@ Line Protocol and TLE imports are unchanged. CSV/Parquet imports no longer depen
 
 Additionally, the Azure `DeleteBatch` implementation was rewritten to use the actual Azure SDK `BlobBatch` API (`container.Client.NewBatchBuilder()` + `SubmitBatch`) instead of a per-file loop, bringing it into parity with the S3 implementation.
 
+**Paginated FSM manifest walks replace full-snapshot copies.** `ClusterFSM.GetAllFiles()` previously allocated a full O(N) copy of the file manifest — for 1M+ files this caused ~50ms apply-path spikes (RLock held while copying) and ~50MB transient memory. The replication catch-up walker and the `/api/v1/cluster/files` endpoint now use cursor-based pagination via `GetFilesPaginated(cursor, limit)`, which:
+
+- Maintains a lazily-built sorted-key cache (invalidated on manifest mutation)
+- Returns pages of entries, releasing the RLock between pages so Raft apply-path latency is unaffected by long-running walks
+- Supports `?cursor=<path>&limit=<n>` query parameters on the API endpoint (backward-compatible — omitting both params returns all files as before)
+
+The emergency kill-switch `replication_catchup_enabled=false` remains available.
+
 ## Upgrade notes
 
 1. **No configuration change required.** Drop in the new binary; existing `arc.toml` and license keys work as-is.
--- a/internal/api/cluster.go
+++ b/internal/api/cluster.go
@@ -369,6 +369,7 @@ func (h *ClusterHandler) handleGetFiles(c *fiber.Ctx) error {
 // paginateSlice applies cursor-based pagination to an in-memory slice.
 // Used for database-filtered results where the FSM's paginated API doesn't
 // natively support database filtering yet.
+// nextCursor is the last path in this page; the next call will resume after it.
 func paginateSlice(files []*clusterraft.FileEntry, cursor string, limit int) ([]*clusterraft.FileEntry, string) {
 	start := 0
 	if cursor != "" {
--- a/internal/api/cluster.go
+++ b/internal/api/cluster.go
@@ -1,6 +1,7 @@
 package api
 
 import (
+	"sort"
 	"strconv"
 
 	"github.com/basekick-labs/arc/internal/auth"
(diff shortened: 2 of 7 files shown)


xe-nvdk (commented):
Reviewed locally — checked out the branch, traced the FSM cache-invalidation across all mutation paths, the pagination cursor logic, and the catch-up paging loop; ran build/vet/test on `internal/cluster/...` (all 7 packages PASS) and `internal/api/...`. **Informational review — no blockers; safe to merge as-is.** A few notes below, mostly minor.

## What I verified as correct (the parts that matter for a Raft FSM change)

- **Cache invalidation is complete.** `keysCache = nil` is set on every manifest-mutating path: register (`fsm.go:1077`), delete (`:1126`), update (`:1208`), and `Restore` (`:2188`). The batch path (`applyBatchFileOps`) delegates to `applyRegisterFileStruct`/`applyDeleteFileStruct` per op (`:104`/`:106`), so it inherits invalidation — no separate bypass. This was the highest-risk area (a missed invalidation = silent replication gaps) and it's airtight.
- **The `Lock`→`Unlock`→`RLock` gap in `GetFilesPaginated` is safe.** `keys := f.keysCache` captures a slice header; invalidation is always reassignment (`= nil`) + rebuild via fresh `make`, never in-place mutation, so the captured `keys` is an immutable stable snapshot even if a concurrent mutation invalidates the cache mid-call. Staleness between the two locks is handled: a deleted key → `f.files[k]` miss is skipped (`:2353 if entry, exists`); a newly-added key → missed this page, which is **documented and acceptable** because catch-up also gets reactive FSM callbacks for new entries.
- **Cursor pagination terminates correctly.** `sort.SearchStrings` + skip-on-exact-match advances past the cursor; final partial page returns `nextCursor=""` (since `end >= len(keys)`); the catch-up loop breaks on both `cursor == ""` (`catchup.go:61`) and empty page — no infinite-loop or skipped-tail risk.
- **API backward-compat holds.** No `cursor`/`limit` params → returns all files (the prior O(N) behavior); with params → paginated. RLock released between pages keeps the apply path off the critical section, which is the whole point of #396.

## Minor / informational (not blocking)

1. **`keysCacheGen` is dead code.** It's incremented on rebuild (`fsm.go:2345`) but never read anywhere; the comment says "used for sanity checks" but nothing checks it. Either wire it into a sanity check or drop it + the comment.
2. **`/api/v1/cluster/files` with `database` filter is still O(N).** The DB-filtered paginated path calls `GetAllFiles`/`GetFilesByDatabase` then slices in-memory via `paginateSlice`, whereas the unfiltered path uses the lock-releasing `GetFileManifestPaginated`. So pagination only delivers its apply-path-latency benefit for the unfiltered walk (which is the catch-up use case, so the goal is met) — but a large DB-filtered API query still allocates the full filtered set. Fine for now; worth a comment so it's not mistaken for fully-paginated.
3. **"Newly-added entries may be missed" cursor semantics** are correctly documented in the `GetFilesPaginated` doc-comment, and the reliance on reactive FSM callbacks to cover the gap is the right argument — just flagging that this is a real (acceptable) eventual-consistency window during catch-up, not exact-snapshot semantics.

## Doc

The release-notes section is accurate and matches the code (lazily-built sorted-key cache, RLock released between pages, `?cursor`/`limit` backward-compatible, kill-switch retained).

Build/vet/gofmt/tests clean. The internal-review findings the PR lists (Restore invalidation, closure-capture safety, cursor-stability docs) are all genuinely addressed in the diff. Good change.

xe-nvdk:

PR #484, opened 2026-06-05
fix(wal): use io.ReadFull instead of f.Read for WAL header reads

### LINKED ISSUE #341: medium(wal): Partial WAL header uses f.Read instead of io.ReadFull
reader.go:109 — cascading misalignment. Fix: use io.ReadFull.

xe-nvdk:
Fixes #341. The WAL reader used f.Read to read fixed-size headers. f.Read may return fewer bytes than the buffer size without error, causing partial reads that cascade into misaligned subsequent reads. Both header reads now use io.ReadFull.

- WAL file header (7 bytes): f.Read → io.ReadFull
- WAL entry header (16 bytes): f.Read → io.ReadFull, io.ErrUnexpectedEOF mapped to io.EOF
- Payload read: = → := (n was previously declared by the removed header f.Read)

All 40 WAL tests pass. gofmt/go vet clean. No configuration-interaction risk.


xe-nvdk:

xe-nvdk:

xe-nvdk:

xe-nvdk:
> mapping `io.ErrUnexpectedEOF` to `io.EOF` when reading entry headers will cause truncated entry headers (e.g., from a crash mid-write) to be silently ignored instead of being flagged as corrupted.

Respectfully, this recommendation would reintroduce the cascading misalignment bug #341 describes. Here is the trace:

If `io.ErrUnexpectedEOF` propagates as an error instead of being mapped to `io.EOF`:

```
readEntry → fmt.Errorf("...: %w", io.ErrUnexpectedEOF)
caller:   err == io.EOF?  → NO (wrapped in fmt.Errorf)
caller:   err != nil → CorruptedEntries++ → continue
           ↑ file offset is 1–15 bytes into a partial header
           ↑ next readEntry parses garbage as payloadLen/timestamp/checksum
           ↑ CASCADING MISALIGNMENT — the exact bug #341 exists to fix
```

The file offset after a partial `io.ReadFull` on the entry header is unrecoverable. There is no way to skip forward — we do not know the payload length because the header is incomplete. The only correct action is to stop reading.

The `io.EOF` mapping matches the original code behavior: the original `f.Read` code checked `err == io.EOF` **before** the `n < WALEntryHeaderSize` check. A `f.Read` returning `(10, io.EOF)` would match `err == io.EOF` first and return `io.EOF` — treating a truncated header at EOF as clean end-of-file, not as a corrupted entry.

Corruption tracking (`CorruptedEntries++`) is meant for recoverable corruption within a valid entry (checksum mismatches), not for unrecoverable partial headers at crash points.

PR #487, opened 2026-06-06
fix(query): add RBAC permission checks to estimateQuery and listMeasurements

xe-nvdk:
## Summary

Adds RBAC read-permission checks to two query endpoints that were missing them:

- **`POST /api/v1/query/estimate`** (`estimateQuery`): now calls `checkQueryPermissions(c, sql, "read")` after SQL validation, matching the pattern already used by `executeQuery`.
- **`GET /api/v1/measurements`** (`listMeasurements`): now calls `checkMeasurementPermission(c, "*", "*", "read")` before the filesystem scan, matching the pattern used by SHOW DATABASES.

Closes #352.

## Changes

| File | Change |
|---|---|
| `internal/api/query.go` | +14 lines: two RBAC check insertions |
| `internal/api/query_rbac_test.go` | New: 8 tests + 7 benchmarks (~290 lines) |
| `RELEASE_NOTES_2026.06.2.md` | 1 paragraph documenting the fix |

## Configuration Matrix

| Configuration | Reaches new code? | Preconditions established? |
|---|---|---|
| OSS standalone (no RBAC) | Yes | `h.rbacManager == nil` → immediate return, 0 alloc, 1.8 ns |
| OSS + RBAC + token with read | Yes | `IsRBACEnabled() == true`, `tokenInfo != nil`, permission passes |
| OSS + RBAC + token without read | Yes | Permission denied → 403 before any DB/storage access |
| Cluster + RBAC | Yes | RBAC independent of cluster; `checkReplicationReady` runs first (no-op) |
| Enterprise + governance + RBAC | Yes | `rbacManager` independent of `governanceManager`/`licenseClient` |

## Benchmark Results (Apple M3 Max)

| Benchmark | Time | Allocs |
|---|---|---|
| `checkMeasurementPermission` — RBAC disabled | **1.8 ns** | 0 |
| `checkMeasurementPermission` — allowed | **51 ns** | 2 |
| `checkMeasurementPermission` — denied | **186 ns** | 7 |
| `checkQueryPermissions` — simple SQL | **6.1 µs** | 8 |
| `checkQueryPermissions` — complex SQL | **30.3 µs** | 19 |
| `checkQueryPermissions` — RBAC disabled | **2.4 ns** | 0 |

Overhead is negligible: the RBAC-disabled path (OSS) has zero measurable cost. The enabled path adds 50ns–30µs — orders of magnitude below query execution times.

## Test Plan

[...]

PR #489, opened 2026-06-06
fix(api): add SHOW command RBAC-gating to estimateQuery endpoint

### LINKED ISSUE #488: estimateQuery does not RBAC-gate SHOW DATABASES/TABLES (count-only metadata leak)
## Summary

`POST /api/v1/query/estimate` (`estimateQuery`) does not gate `SHOW DATABASES` / `SHOW TABLES` / `SHOW MEASUREMENTS` behind an RBAC permission check, unlike `POST /api/v1/query` (`executeQuery`). A token **without** `*.*:read` can learn the **count** of databases (or measurements in a database) by submitting a SHOW command to the estimate endpoint.

This is an information-leak and an authorization-parity gap, not a data-exfiltration bug — only a row count is returned, never database/measurement names or row data.

## Details

In `executeQuery` (internal/api/query.go), SHOW commands are special-cased and gated:

- `SHOW DATABASES` → `checkMeasurementPermission(c, "*", "*", "read")` (~line 1404)
- `SHOW TABLES|MEASUREMENTS [FROM db]` → `checkMeasurementPermission(c, db, "*", "read")` (~line 1418)

`estimateQuery` (~line 3165) has no equivalent handling. Flow for `{"sql": "SHOW DATABASES"}`:

1. `ValidateSQLRequest` passes (SHOW is not a dangerous pattern).
2. `hasCrossDatabaseSyntax` is false.
3. `checkQueryPermissions` extracts **zero** table references (no `FROM db.table`) → returns nil (allowed).
4. `getTransformedSQL` passes the SQL through unchanged (no FROM/JOIN).
5. It executes as `SELECT COUNT(*) FROM (SHOW DATABASES) AS t` (~line 3222), returning the count.

So a token scoped to a single measurement, with no `*.*:read`, can call `/api/v1/query/estimate` with `SHOW DATABASES` and get back the total number of databases — information `executeQuery` would deny.

## Impact

- Severity: Low (count-only metadata leak; requires a valid authenticated token; RBAC must be enabled).
- Scope: Arc Enterprise with RBAC enabled. OSS / RBAC-disabled deployments are unaffected (the check is skipped when RBAC is not configured).

## Suggested fix

Mirror the `executeQuery` SHOW handling in `estimateQuery`: match `showDatabasesPattern` / `showTablesPattern` and gate with the same `checkMeasurementPermission(c, "*"|db, "*", "read")` before estimating — or simply reject SHOW commands on the estimate endpoint (estimating a row count for a metadata command has little value).

## Context

Found by internal multi-agent review during PR #487 (which added the RBAC checks to `estimateQuery` and `listMeasurements`). The gap is pre-existing relative to that PR's scope — `estimateQuery` never handled SHOW — so it was tracked separately rather than expanding #487. A sibling finding in the same review (user `read_parquet()` bypass) was investigated and **refuted**: `ValidateSQLRequest` already rejects direct `read_parquet()` in user SQL, and `estimateQuery` calls it first.

xe-nvdk:
## Summary

Two query endpoints passed `SHOW DATABASES` / `SHOW TABLES` / `SHOW MEASUREMENTS` to DuckDB without an RBAC permission check. `checkQueryPermissions` extracts table references from `FROM`/`JOIN` clauses, and SHOW commands have none, so they bypassed authorization. This PR closes that gap on **both** the estimate and the main query endpoints.

### `estimateQuery` (`POST /api/v1/query/estimate`) — original scope (#488)
- `SHOW DATABASES` requires read on `*.*`; `SHOW TABLES`/`SHOW MEASUREMENTS` requires read on the target database.
- After the permission check passes, returns `400` directing callers to `/api/v1/query` — the estimate endpoint has no legitimate use for metadata commands.
- SHOW matching runs on **comment-stripped, trimmed** SQL so a comment (`/* x */ SHOW DATABASES`) cannot hide the command from the anchored regex.
- `SHOW TABLES` with no explicit `FROM db` checks permission against the `x-arc-database` header (the database the command implicitly targets), falling back to `"default"` only when neither is set.

### `executeQuery` (`POST /api/v1/query`) — expanded scope (pre-existing bugs found during review)
While verifying the estimate fix, the same SHOW gate in `executeQuery` was found to have **two pre-existing bugs**, both more serious there because `executeQuery` *executes* the listing (estimate only returns `400`):

1. **Comment-bypass (security, data-returning).** The SHOW gate matched **raw `req.SQL`**, so `/* x */ SHOW DATABASES` failed the anchored regex, fell through `checkQueryPermissions` (zero table refs → allowed), and DuckDB stripped the comment and executed it — **returning the database/table list** to a caller the gate would have denied. Now matched against comment-stripped SQL.
2. **Header-scoped database.** `SHOW TABLES` with no `FROM db` defaulted the permission check to `"default"`, ignoring the `x-arc-database` header that `handleShowTables` actually lists. Now resolves to the header DB before `"default"`.

## Configuration Matrix

|---|---|---|
| OSS standalone (no cluster, no compaction) | Yes — both handlers have no cluster/license gates | All compile-time: regexps, `scanSQLFeatures`/`stripSQLComments`, `checkMeasurementPermission` (guards nil `rbacManager`/`tokenInfo` internally), `metrics.Get()` |
| OSS + compaction | Same | Same |
| Cluster + no compaction | Same | Same |
| Cluster + compaction + server.tls_enabled | Same | Same |
| Cluster + compaction + cluster.tls_enabled | Same | Same |

No new pointer dereferences on optional subsystems.

## Test Plan

- [x] `gofmt -l` / `go vet ./internal/api/...` — clean
- [x] `go test ./internal/api/... -count=1` — full package passes
- [ ] Regression tests for the `executeQuery` comment-bypass and header-scope fixes are prepared and revert-verified (each fails — falls through to a nil-deref — when its fix is reverted, proving the request previously escaped the gate). They are committed to `main` in a follow-up so they sit alongside the authoritative `query_rbac_test.go` rather than this branch's stale copy.

## Notes

- Rebased onto `main` (post-#487) — resolves the earlier conflict; `query_rbac_test.go` is now main's version.
- The `executeQuery` comment-bypass is a live, data-returning RBAC bypass on `main` today; this PR is the fix.

Closes #488


xe-nvdk:
@gemini-code-assist review this please

xe-nvdk:
Addressed Gemini review: SQL is now normalized via `scanSQLFeatures` + `stripSQLComments` + `strings.TrimSpace` before the anchored SHOW regexes are matched (commit 21f3f14). This prevents the comment/whitespace bypass — a query like `/* x */ SHOW DATABASES` is now correctly caught.

xe-nvdk:
## Scope expanded + rebased — re-review requested

Heads-up on two changes since the last review:

**1. Expanded to fix the same bug class in `executeQuery`.** The original PR gated SHOW on the estimate endpoint (#488). While verifying that fix, I found the SHOW gate in `executeQuery` (`POST /api/v1/query`) had two pre-existing bugs — and worse there, because `executeQuery` *executes* the listing rather than returning `400`:

- **Comment-bypass (data-returning):** the gate matched raw `req.SQL`, so `/* x */ SHOW DATABASES` slipped past it, fell through `checkQueryPermissions` (zero table refs → allowed), and DuckDB stripped the comment and executed it — returning the database list to an unauthorized caller. Now matched against comment-stripped SQL (same normalisation `checkQueryPermissions` uses).
- **Header-scoped db:** `SHOW TABLES` with no `FROM db` checked permission against `"default"`, ignoring the `x-arc-database` header that `handleShowTables` actually lists. Now resolves to the header DB first.

This is a live, data-returning RBAC bypass on `main` today, so folding it in here rather than deferring.

**2. Rebased onto `main`** (post-#487) to clear the earlier conflict. `query_rbac_test.go` is now main's authoritative version.

**On tests:** regression tests for both `executeQuery` fixes are written and revert-verified (each fails with a nil-deref when its fix is reverted, proving the request previously escaped the gate). They'll be committed to `main` in a follow-up so they live alongside the authoritative `query_rbac_test.go` instead of this branch's stale copy — this keeps the diff focused on the production fix.

@gemini-code-assist please review.

xe-nvdk:
Fixed the critical: `showTablesPattern` now matches quoted database names (`"db"`, `'db'`, `` `db` ``), capturing the unquoted name so the RBAC check and `handleShowTables` use the real database. RE2 has no backreferences, so quotes are matched independently — safe here, since the goal is to *recognize* the SHOW so it's gated (over-recognizing is harmless; only a non-match would bypass).

Verified the bypass empirically (`SHOW TABLES FROM "secretdb"` extracted 0 refs before the fix) and extended `TestShowTablesPattern` with quoted-name cases (capturing the unquoted db) plus a trailing-junk negative.

@gemini-code-assist please review.

xe-nvdk:
Fixed the critical path-traversal and proactively closed a parity gap found by an internal multi-agent review:

**Path traversal (your finding):** `SHOW TABLES FROM ..` captured `..` (the regex allows dotted tokens) and, with RBAC disabled, would reach `handleShowTables` (which lists `database + "/"`). Now `validateIdentifier` validates the resolved database name — on **both** `executeQuery` and `estimateQuery` — before the RBAC check and any storage access. Verified end-to-end: `SHOW TABLES FROM ..` → 400 before storage; quoted valid names still pass.

**Arrow endpoint parity (proactive):** `POST /api/v1/query/arrow` (`executeQueryArrow`, build-tagged `duckdb_arrow`) had no SHOW gate, unlike the JSON and MsgPack paths. Not a data leak — the Arrow path never calls `handleShowTables`, so SHOW there only hits DuckDB's empty internal catalog — but it was an authorization-parity gap. Added the same comment-normalised SHOW gate + `validateIdentifier`, returning 400 directing to `/api/v1/query`.

A three-agent internal review (security / correctness / verification, run against the configuration matrix) otherwise found the gate sound: regex misses are **fail-safe** (DuckDB holds no Arc catalog, so a missed SHOW returns DuckDB's empty catalog, never Arc storage — only Arc's own `handleShow*` lists tenant data, and it's reached only via the gate), `validateIdentifier` is a strict superset of the DB-creation validator (no false-positive denials), and the two endpoints are faithfully mirrored.

Release notes updated to cover comment-normalisation, quoted names, header resolution, and traversal rejection across all four query endpoints.

@gemini-code-assist please review.

xe-nvdk:
Fixed — but centralised rather than per-endpoint.

I added the multi-statement rejection to **`ValidateSQLRequest`** (the shared validator called by `executeQuery`, `executeQueryMsgPack`, `executeQueryArrow`, and `estimateQuery`) rather than duplicating it in each SHOW gate. One guard covers all four endpoints, runs before any SHOW matching, reuses the SQL that's already normalised there (literals masked, comments stripped), and avoids needing a `sqlutil` import in `query_arrow.go`. A single trailing `;` is allowed; semicolons inside string literals/comments don't false-positive.

Scope correction on the severity: Arc databases are **virtual** — resolved to `read_parquet(...)` paths, never `ATTACH`ed to DuckDB (no `ATTACH`/`CREATE DATABASE` anywhere in `internal/database/`). So DuckDB's native `SHOW DATABASES`/`SHOW TABLES` returns DuckDB's own empty catalog, **not** Arc tenant data — the smuggled `SHOW` executes but discloses nothing tenant-scoped. The dangerous-keyword denylist also already blocks destructive second statements (`DROP`/`ATTACH`/`COPY`/`PRAGMA`/`SET`/`LOAD`/`CALL`/…). So this is defense-in-depth closing a smuggling surface, not a tenant-data RBAC bypass. Either way, multi-statement input has no legitimate use on these read endpoints, so it's now rejected outright.

Tests added to `TestValidateSQLRequest_BypassesAndFalsePositives` (smuggled SHOW rejected; trailing `;`, `;`-in-literal, `;`-in-comment allowed).

@gemini-code-assist please review.

xe-nvdk:
Fixed via a shared `normalizeSQLForShow(sql)` helper: it masks string literals **before** stripping comments, then unmasks — so a comment marker inside a quoted identifier (`SHOW TABLES FROM "my--db"`) is preserved and the regex captures the full db name, while real comments outside literals are still stripped. Applied at all three SHOW gates (executeQuery, estimateQuery, executeQueryArrow). The multi-statement guard in `ValidateSQLRequest` already mask-then-strips, so it was unaffected.

Verified the bypass empirically before the fix (raw strip → `SHOW TABLES FROM "my`, capturing db `my`) and added `TestNormalizeSQLForShow` (quoted-comment preserved + full-name capture; real comments stripped).

Tracking the broader pattern: I've opened #491 to harden `ValidateSQLRequest` as the single SQL-safety chokepoint and move SHOW/RBAC gating off raw-regex onto a normalized representation (+ fuzz coverage), since this PR's rounds have all been tokenizer edge cases against the regex approach. Not blocking this PR — the defensive posture holds (Arc DBs are virtual, denylist blocks destructive statements) — but it's the durable fix.

@gemini-code-assist please review.

xe-nvdk:
Fixed both backtick findings at the root — same class as round 7, one layer down: `MaskStringLiterals` only recognises `'` and `"`, not backticks.

- **HIGH** (`SHOW TABLES FROM \`my--db\``): the `--` inside the backtick name wasn't masked → comment-strip truncated it → gate authorised `my` while DuckDB lists `my--db`.
- **MEDIUM** (`SELECT * FROM \`my;db\``): the `;` inside the backtick name wasn't masked → false multi-statement rejection.

Fix is unified in one helper (`backticksToDoubleQuotes`) applied in both `normalizeSQLForShow` and `ValidateSQLRequest` before masking — so backtick-quoted content (comment markers, semicolons, keywords) is protected from all the safety scans. I deliberately did **not** change `MaskStringLiterals` itself: it's shared by the execution transform (mask→rewrite→unmask) and the partition pruner, where treating backtick identifiers as string literals could alter path rewriting — too broad a blast radius for this PR. The swap is scoped to the two read-only safety scans (their normalized output is only scanned, never reconstructed into executable SQL).

Tests added for both cases (full-name capture for the backtick comment marker; backtick semicolon not false-rejected; real multi-statement still rejected).

Note: I've opened #491 to move this gating off raw-regex onto a real tokenizer with fuzz coverage — that's the durable fix for this recurring edge-case class. This PR remains strictly safer than `main` (which still has the original executeQuery comment-bypass).

@gemini-code-assist please review.

Issue #491, opened 2026-06-06
Harden ValidateSQLRequest as the single SQL-safety chokepoint (move SHOW/RBAC gating off raw-regex)


xe-nvdk:
## Problem

The query API enforces its SQL-safety invariants — RBAC table extraction, SHOW-command gating, cross-database rejection, dangerous-keyword denylist, multi-statement rejection, `read_parquet`/`arc_partition_agg` blocking — with **regex and string-scanning over the raw SQL**, not a real parser. This is an inherently leaky abstraction: each invariant has to independently re-derive "what is a token / literal / comment / statement boundary," and every divergence from real SQL tokenization is a potential bypass.

PR #489 (a small SHOW-gating change) made this concrete. It took **7 Gemini rounds**, each surfacing a distinct tokenizer edge case in the same gate:

1. Comment hides the command — `/* x */ SHOW DATABASES` (regex on raw SQL).
2. Quoted db name not recognized — `SHOW TABLES FROM "db"` (char class too narrow).
3. Path traversal — `SHOW TABLES FROM ..` (char class too wide; no identifier validation).
4. Multi-statement smuggling — `SHOW DATABASES; SELECT 1` (anchored `$` assumes one statement).
5. Whitespace before paren in TVF detection — `generate_series  (...)`.
6. Endpoint parity — Arrow endpoint had no SHOW gate at all.
7. Comment marker inside a quoted identifier — `SHOW TABLES FROM "my--db"` (stripped comments before masking literals → truncation → wrong db authorized).

Each was individually fixed and is covered by a regression test, and the **defensive posture held throughout** (Arc databases are virtual / never `ATTACH`ed, so a missed SHOW returns DuckDB's empty catalog, not tenant data; the denylist blocks destructive statements). So these were defense-in-depth hardening, not live tenant-data leaks. But the *rate* of new edge cases in one small PR is the signal: the approach doesn't converge.

## Proposed direction

Treat **`ValidateSQLRequest` as the single SQL-safety chokepoint** and make the SQL handling robust rather than regex-by-feature:

1. **Normalize once, centrally.** All endpoints already call `ValidateSQLRequest`. Do the literal-masking + comment-stripping there once, and have it expose the normalized form (and/or a small set of derived facts: statement count, leading keyword, whether it's a SHOW/metadata command) so downstream gates don't each re-scan raw SQL. The `normalizeSQLForShow` helper added in #489 is a step toward this — generalize it.
[...]

PR #492, opened 2026-06-06
test(api): endpoint-level regression tests for SHOW-command RBAC gate

xe-nvdk:
## Summary

Adds two handler-level (not just regex) regression tests for the SHOW-command RBAC gate merged in #489. They were prepared during #489 but deferred, because that branch carried a stale copy of `query_rbac_test.go`; they now land alongside the authoritative file on `main`.

- **`TestExecuteQuery_ShowDatabases_CommentBypassDenied`** — a commented SHOW (`/* x */ SHOW DATABASES`, `-- x\nSHOW DATABASES`, trailing comment) must be denied with 403, not slip past the gate to DuckDB.
- **`TestExecuteQuery_ShowTables_HeaderDBScoped`** — `SHOW TABLES` with no `FROM db` must check permission against the `x-arc-database` header (the db `handleShowTables` actually lists), not literal `"default"`.

Both are **revert-verified**: with the corresponding fix removed from `query.go`, the request escapes the gate (nil-deref 500 instead of 403), proving the tests guard the behavior rather than passing trivially.

## Test plan

- [x] `go test ./internal/api/... -count=1` — passes
- [x] `go test -tags duckdb_arrow ./internal/api/... -count=1` — passes
- [x] `gofmt -l` / `go vet ./internal/api/...` — clean
- [x] Revert-verified both tests fail when their fix is removed

Test-only change; no production code touched. Part of the coverage tracked by #491.

### DIFF
--- a/internal/api/query_rbac_test.go
+++ b/internal/api/query_rbac_test.go
@@ -802,8 +802,9 @@ func TestExecuteQuery_ShowDatabases_CommentBypassDenied(t *testing.T) {
 		if err != nil {
 			t.Fatalf("%q: %v", sql, err)
 		}
+		bodyBytes, _ := io.ReadAll(resp.Body)
+		resp.Body.Close()
 		if resp.StatusCode != fiber.StatusForbidden {
-			bodyBytes, _ := io.ReadAll(resp.Body)
 			t.Errorf("%q: expected 403 (SHOW gate must deny), got %d: %s",
 				sql, resp.StatusCode, string(bodyBytes))
 		}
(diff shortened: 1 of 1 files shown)


xe-nvdk:
@gemini-code-assist please review — test-only PR adding revert-verified endpoint regression tests for the SHOW RBAC gate from #489.

xe-nvdk:
Fixed — both tests now close `resp.Body` (read+close per iteration in the loop, `defer` in the single-request test). @gemini-code-assist please review.

xe-nvdk:
Addressed the second review: replaced the `jsonQuote` helper with a table of pre-escaped JSON request bodies and removed the helper. Response bodies are closed (read+close in the loop, `defer` in the single test) — that was already in the prior commit. @gemini-code-assist please review.

Issue #2, opened 2025-10-09
Data loss prevention


notABot101010:
As mentioned in #1 , there is a possibility of data loss if the process / node crash when the data is buffered in memory.

It was mentioned that a Write Ahead Log (WAL) is planned.

I would also recommend considering SQLite which seems way easier to implement, and can easily be migrated to a WAL later if using the correct abstraction.

SQLite's performance allows to reduce batch interval to something like 1-5 ms and will allow creating parquet files with even better optimized size, reducing the need for further compaction.

PR #493, opened 2026-06-06
fix(compaction): stop time-type-mismatch wedging partitions; un-wedge existing

xe-nvdk:
## Summary

Fixes a regression where **compaction permanently wedges a partition** whose Parquet files disagree on the `time` column type (`INTERNAL Error: Failed to bind column reference "time": TIMESTAMP WITH TIME ZONE != VARCHAR`). The partition fails on every compaction cycle while raw files accumulate; queries still work, so it's silent unless you watch compaction logs/history.

## Root cause

**#411 (shipped in 26.05.1)** made the ingest buffer's schema-change signature *type-aware* (`name:type`). When a writer sent `time` as a non-integer type for some batches (e.g. a client encoding timestamps as strings) and integer epochs for others, the differing signatures routed those batches into **separate Parquet files** in the same partition. Compaction reads with `read_parquet(union_by_name=true)`, which reconciles column *names* but not conflicting *types* → bind failure → wedged.

Confirmed by binary bisect: every build back to April fails identically on the affected files, so the **compaction code never regressed** — #411's writer-side file fan-out is what turned a benign "client sends odd time" into un-compactable storage.

## Fix (three levels)

1. **`getColumnSignature` reverted to names-only** — undoes #411's type-awareness so a column whose type varies across batches stays in one buffer instead of fanning out into type-conflicting files. The `strings.Builder` allocation win from #411 is kept; **no added ingestion overhead** (in fact removes the per-column type switch).
2. **`time` forced to int64 microseconds** in `convertColumnsToTyped` (the single typing chokepoint). A string `time` is **rejected with a clear error at ingest** rather than silently written as `VARCHAR`. No extra data pass.
3. **Compaction normalizes `time`** via `COALESCE(TRY_CAST(time AS TIMESTAMP), make_timestamp(TRY_CAST(time AS BIGINT)))` so partitions that **already** contain mixed `TIMESTAMP`/`VARCHAR` `time` files (from an affected 26.05.1 deployment) compact on the next cycle — **no manual intervention**.

Also: `sustained_bench --backdate-hours` now applies to the lineprotocol path (was iot/msgpack only), used to reproduce/validate backfill compaction.

## Test plan

- [x] `go build ./cmd/... ./internal/...` and `-tags=duckdb_arrow` — clean
- [x] `go test ./internal/ingest/...` — pass (names-only signature + no-fan-out guard; `time` forced int64 / string rejected)
- [x] `go test ./internal/compaction/...` — pass
- [x] `gofmt -l` / `go vet` — clean
- [ ] **Dev validation before release:** re-ingest via the affected msgpack/Telegraf writer; confirm new data compacts and an existing mixed-type partition compacts on next cycle.

## Notes

- Targeted for **26.06.2**; release notes updated under Bug fixes.
- Follow-up worth filing: compaction failures are background-only — an operator metric/alert for repeated per-partition compaction failures would have surfaced this far sooner (it ran undetected since 26.05.1).


xe-nvdk:
@gemini-code-assist please review — targeted for 26.06.2 release tonight. Fixes a 26.05.1 regression (#411) that wedges compaction on time-type-mismatched partitions; see PR body for the three-level fix and root-cause bisect.

xe-nvdk:
All three addressed — these were excellent catches, especially the TIMESTAMPTZ one which would have re-broken compaction on the next cycle.

**CRITICAL (names-only → merge panic):** You're right, I threw out load-bearing behavior. Reverted to keep the **type-aware `name:type` signature** — it protects non-time columns from the `mergeBatches` type-assertion panic (verified: int64-then-float64 on a field does panic without it). The time-fan-out that #411 caused is instead prevented by **Fix #2 forcing `time`→int64** at the typing chokepoint, so `time`'s signature component is always `time:i64` and it can never fan out, while other columns stay panic-safe. The two compose.

**HIGH (TIMESTAMP vs TIMESTAMPTZ):** Confirmed Arc writes `Timestamp_us{TimeZone:"UTC"}` → `TIMESTAMP WITH TIME ZONE`. Naive `TIMESTAMP` normalize would strip the zone and recreate the mismatch (compacted-naive vs new-raw-tz) next cycle. Changed to `TRY_CAST(... AS TIMESTAMPTZ)` + `make_timestamptz`. **Verified end-to-end against the shipped DuckDB**: a TIMESTAMPTZ file + a VARCHAR-epoch file compact together and the output `time` type is `TIMESTAMP WITH TIME ZONE`.

**MEDIUM (null time):** Now rejected with a clear error at ingest (`groupByHour` reads the slice without validity → nil would silently route to the 1970 partition). Also removed the eager `valid`-slice allocation for `time` in the process.

Tests updated accordingly (type-aware signature + different-signature guard; null-time rejection).

@gemini-code-assist please re-review.

xe-nvdk:
Applied — added the `v.(int64)` fast path before the `toInt64` fallback in the time loop. Behavior unchanged (float/uint still go through toInt64, nil rejected, non-numeric errors); just skips the call + type switch for the common int64 case. @gemini-code-assist please re-review.

PR #494, opened 2026-06-07
fix(compaction): fix daily dedup time-bind (CTE-normalized window) + DuckDB regression test

xe-nvdk:
Follow-up to #493. The **dedup** compaction branch (measurements with tag metadata — which includes the **daily tier**) still failed on time-type-mixed partitions even after #493. Found in production: `daily compaction` on `production/cpu/2026/05/31` failing with `INTERNAL Error: Failed to bind column reference "time": TIMESTAMP WITH TIME ZONE != VARCHAR`.

## Root cause (two layers)

1. The original dedup form `SELECT *, ROW_NUMBER() OVER(...) AS rn FROM (… union_by_name …) ) WHERE rn=1` re-expands `*` over the raw union and mis-binds `time` as VARCHAR — even when every file's `time` is valid TIMESTAMPTZ.
2. A naive switch to **top-level** `SELECT * REPLACE(time…) … QUALIFY ROW_NUMBER() OVER(PARTITION BY …, "time")` is **wrong**: QUALIFY is evaluated *before* the SELECT projection, so the window runs over the **raw** `time`, not the normalized one. On a mixed-type partition the same `(tags, time)` renders as two different strings → two window partitions → **both rows survive** → silent under-dedup (duplicates written, sources deleted). This was caught by the deep reviewer, and reproduced empirically (2 rows instead of 1).

## Fix

Materialize the `time` normalization in a **CTE** the dedup window reads from, so the window/QUALIFY key on the already-normalized TIMESTAMPTZ `time`.

## Verification (against real shipped DuckDB)

- Real 134-file partition (`production/cpu/2026/05/31`) compacts with **no bind error**.
- A partition mixing a TIMESTAMPTZ-time file and a VARCHAR-time file dedups to **exactly 1 row** (top-level QUALIFY gave 2).
- Output `time` type is `TIMESTAMP WITH TIME ZONE`, instant preserved **exactly under a non-UTC session timezone** — UTC integrity confirmed (`make_timestamptz` is UTC-anchored; session tz only affects display).
- New **DuckDB-backed regression test** with real mixed-type parquet fixtures; it fails (2 rows) if reverted to top-level QUALIFY. Prior tests were string-only and could not catch binder behavior.

## Configuration matrix

| Config | Reaches dedup CTE branch? | Preconditions |
|---|---|---|
| OSS, compaction on, measurement WITH tags | Yes | tagColumns from readTagColumnsFromParquetFiles (error → non-dedup branch) |
| OSS, measurement WITHOUT tags | No — unchanged non-dedup branch (already normalizes via #493) | n/a |
| OSS + cluster (writer/compactor) | Yes — query is cluster-independent | manifest handoff downstream of COPY |
| Enterprise + tiering | Yes | output schema `Timestamp(isAdjustedToUTC=true)`, ingest-compatible |
| with SortKeysConfig | Yes — orderByClause applied after QUALIFY | distinct from window ORDER BY |

## Test plan
- [x] `go build` + `-tags=duckdb_arrow` clean; `go vet`/`gofmt` clean
- [x] `go test ./internal/compaction/...` (both tags) — pass
- [x] DuckDB-backed mixed-type dedup regression test (revert-verified)
- [x] UTC preserved under non-UTC session timezone

Targeted for 26.06.2 (release notes updated). Reviewed via CLAUDE.md config-matrix + deep reviewer; the reviewer caught the QUALIFY-precedence under-dedup before merge.


xe-nvdk:
@gemini-code-assist please review — follow-up to #493 fixing the daily/dedup compaction path. Deep review already caught and fixed a QUALIFY-precedence under-dedup bug; PR body has the root cause, the CTE fix, and the DuckDB-backed regression test. Targeted for 26.06.2.

xe-nvdk:
Fixed both — removed the unused `os` import and the leftover `_ = os.Remove` (t.TempDir() handles cleanup). @gemini-code-assist please re-review.

xe-nvdk:
Applied — wrapped the temp paths in `filepath.ToSlash` so the DuckDB SQL interpolation is correct on Windows too. @gemini-code-assist please re-review.

xe-nvdk:
Good catch — applied, with the `duckdb_arrow`-is-the-build-convention context in mind. Tagged the DuckDB-backed integration test with `//go:build duckdb_arrow` (matching Makefile `-tags=duckdb_arrow` and the existing query_msgpack_test.go precedent), and moved the string-only assertion into the untagged dedup_test.go so it still runs in both configs. Verified untagged build no longer pulls the CGO driver and tagged run still exercises the mixed-type regression. @gemini-code-assist please re-review.

xe-nvdk:
Applied to all five interpolation sites — wrapped the temp paths in `escapeSQLPath` (the same helper buildCompactionQuery uses), so the test SQL is robust to a single-quote in t.TempDir()'s path. @gemini-code-assist please re-review.

xe-nvdk:
Good cross-reference against our own release notes. Swapped to `typeof("time")` — verified it returns the identical `TIMESTAMP WITH TIME ZONE` and avoids the DESCRIBE-subquery form we flagged as fragile on the shipped DuckDB. @gemini-code-assist please re-review.

xe-nvdk:
Addressed the portability concern — but **not with the suggested `'-03:00'`**, which the shipped DuckDB rejects (`Not implemented Error: Unknown TimeZone '-03:00'`). Used `Etc/GMT+3` instead: a fixed POSIX-offset zone (UTC-03) built into DuckDB's ICU, so it needs no host tzdata (won't be flaky on alpine/minimal runners) while still exercising a non-UTC session. Verified it preserves the UTC instant exactly. @gemini-code-assist please re-review.

PR #495, opened 2026-06-10
fix(security): block DuckDB I/O function family in user SQL (GHSA-93cm-2v4m-c56c)

xe-nvdk:
## Summary

Completes the incomplete **CVE-2026-47735** fix (private advisory **GHSA-93cm-2v4m-c56c**, reporter @anir0y).

The 26.06.1 sandbox change blocked only the literal `read_parquet()` and `arc_partition_agg()` spellings. DuckDB exposes the same filesystem reads through the `parquet_scan()` alias and a wider family of path-taking table functions (`glob`, `read_blob`, `read_csv*`, `read_json*`, `read_ndjson*`, `read_text`, `parquet_metadata`/`parquet_schema`, `read_xlsx`, `delta_scan`, `iceberg_scan`, …). Because the DuckDB sandbox allowlists the **entire** storage root and the RBAC layer does not authorize function calls, any authenticated principal — including a least-privileged RBAC-scoped token — could read another database's Parquet data, file listing, and column statistics, e.g.:

```
SELECT * FROM parquet_scan('./data/arc/<other-db>/<m>/**/*.parquet')
```

**Reproduced live** on a running instance (auth + RBAC + license): a token scoped to `attackerdb:read` was denied `victimdb.secret` via the legitimate path (403) yet read the identical secret via `parquet_scan` (200).

## What changed

**1. `internal/api/query.go` — reject the I/O function family in user SQL.**
The validator now rejects the whole path-taking table-function family **by name, in any query position** — direct `FROM`, `JOIN`, comma cross-joins (`FROM a, parquet_scan(...)`), subqueries, `IN`-lists, and lateral joins. Whole-string name matching is used instead of FROM/JOIN-position anchoring: anchoring inherits the same comma-join blind spot the RBAC table-reference extractor has, which would leave `FROM a, parquet_scan(...)` open (and would be a regression vs the prior whole-string `read_parquet` denylist). Validation runs on user input **before** the query-transform layer — the only path that legitimately emits `read_parquet`. Pure non-I/O table functions (`generate_series`, `range`, `unnest`, …) and all scalar/aggregate functions are unaffected.

**2. `internal/auth/auth.go` + `auth_routes.go` — RBAC-only tokens.**
Secondary privilege-escalation-shaped bug: `CreateToken` / `CreateTokenWithValue` re-defaulted an empty permission string to `read,write`, so `POST /api/v1/auth/tokens` with `"permissions": []` (documented to create an RBAC-only token) was silently upgraded to a wildcard `read,write` token. A `PermissionsNone` sentinel — resolved once via `storePermissions()` — now persists a true no-platform-permissions token. The `""` => `read,write` default contract is unchanged for nil/omitted permissions. The cluster Raft path carries the resolved value and does not double-apply.

## Test plan

- [x] `go test -race ./internal/api/... ./internal/auth/...`
- [x] `go vet` + `gofmt` clean; full repo builds (`-tags=duckdb_arrow`)
- [x] New regression tests: I/O-family rejection across positions (comma-join, subquery, IN-list, lateral, scalar subquery); family-pinning test against the I/O list; false-positive guards (string literals, scalar funcs, pure table funcs, identifier superstrings); RBAC-only-token regression
- [x] **End-to-end in Docker** (auth + RBAC + license): all bypass vectors blocked; legit queries (`generate_series`, `range` in a comma-join, scalar `count`/`lower`, own-db reads) still work; RBAC denial intact; RBAC-only token created via API


xe-nvdk:
@gemini-code-assist please review — security fix completing the incomplete CVE-2026-47735 fix (private advisory GHSA-93cm-2v4m-c56c). Closes an RBAC/cross-tenant read bypass via the parquet_scan alias and the rest of DuckDB's file-I/O table-function family. Key area to scrutinize: completeness of the I/O-family denylist in internal/api/query.go (ioTableFunctionPattern) and whether any path-taking DuckDB function can still reach the executor through user SQL in any position.

xe-nvdk:
@gemini-code-assist Thanks for the review. I verified the `read_tsv` / `read_tsv_auto` finding against the DuckDB version Arc actually ships (**1.5.1**, pinned via `duckdb-go/v2 v2.10501.0`) and it does **not** hold — these functions do not exist in this version, so there is no bypass to close:

```
-- via Arc's query endpoint on the shipped binary (DuckDB v1.5.1):
SELECT function_name FROM duckdb_functions() WHERE function_name LIKE '%tsv%';
-- => 0 rows

SELECT * FROM read_tsv('/etc/hostname');
-- => Catalog Error: Table Function with name read_tsv does not exist!
```

DuckDB has no dedicated `read_tsv` table function; TSV files are read via `read_csv(..., delim='\t')`, and **`read_csv` is already in the denylist**, so the tab-delimited read path is covered. Adding non-existent function names would only mislead future maintainers about what the shipped catalog exposes — `TestIOTableFunctionPattern_Family` is intentionally pinned to functions that exist in the shipped DuckDB version.

Declining the three suggestions on that basis. If a future DuckDB release introduces `read_tsv`, the family-pinning test is where we'd catch and add it during the version bump. Everything else in the review — the denylist approach and the RBAC-only-token fix — you had no concerns with; happy to dig into any other area.

xe-nvdk:
@gemini-code-assist Confirmed and fixed — this one is a real bypass, good catch. Verified end-to-end on the running binary (DuckDB 1.5.1, auth+RBAC) with a least-privilege `attackerdb:read` token: `SELECT * FROM "parquet_scan"('…/victimdb/…')` returned the victim's rows **before** the fix, and is rejected **after**.

Root cause was exactly as you described: the I/O denylist matched against the shared `normalised` string, which masks double-quoted/backtick identifiers as string literals, so `"parquet_scan"` was hidden inside a `__STR__` placeholder.

Fix (commit `2dedeb8`): the I/O denylist now runs against a dedicated `ioDenylistNormalise(sql)` that strips identifier quoting (`"` and backtick) **before** masking — exposing quoted function names to the regex — while still masking single-quoted **string** literals so a value like `'parquet_scan ran'` does not false-positive, and stripping comments. I kept this separate from the general `normalised` form (used by the multi-statement and dangerous-keyword checks), which is unchanged.

Regression tests added for the double-quoted/backtick spellings (including a comma-join variant) plus negative guards for a single-quoted string containing a function name and a quoted **column** identifier (`"host"`). Post-fix validation: all quoted-identifier vectors blocked, unquoted still blocked, legit queries (own-db read, `generate_series`, quoted column, string literal) unaffected, RBAC denial intact.

PR #497, opened 2026-06-11
fix(storage): track storage metrics in S3/Azure backends + export read counters to Prometheus

### LINKED ISSUE #349: low(storage): S3/Azure backends dont track storage metrics
Local does. Fix: add metrics calls.

xe-nvdk:
Closes #349.

## Summary

- The local filesystem backend records storage counters via `metrics.Get()`, but the S3/MinIO and Azure Blob backends recorded nothing — on cloud-storage deployments `arc_storage_*` counters sat at zero. Both backends now mirror the local backend's instrumentation exactly: writes (`WriteReader`, S3 multipart), reads (`Read`, `ReadTo`, `ReadToAt`), byte counters, and `IncStorageErrors` on failure paths.
- Unknown-size streaming uploads (S3 multipart with `size <= 0`, Azure `UploadStream`) count the write but skip the byte counter; the caveat is documented at both sites.
- No double counting: `Write` delegates to `WriteReader` in both backends, and no caller-level instrumentation exists (verified repo-wide).
- **Related export gap fixed:** `arc_storage_reads_total` and `arc_storage_read_bytes_total` were tracked (and present in the JSON metrics snapshot) but missing from the Prometheus text export — read metrics were invisible to Prometheus even for local storage. Added the two stanzas plus `TestPrometheusFormat_StorageMetrics`, locking all five `arc_storage_*` counters into the export format.
- Release notes entry added to `RELEASE_NOTES_2026.06.2.md`.

## Test plan

- [x] `go vet` and `gofmt -l` clean on touched files
- [x] `go test ./internal/storage/ ./internal/metrics/ -count=1` passes
- [x] New `TestPrometheusFormat_StorageMetrics` asserts all five counter stanzas in the Prometheus output
- [x] Internal config-matrix + deep review completed (no Blocker/High findings; both Medium findings addressed: export regression test + declared-size caveat comments)


xe-nvdk:
@gemini-code-assist Both rounds of findings are addressed in the latest commit — partial-read bytes are now counted on mid-stream copy failures (extended to the local backend for parity), and the storage-error counter is now cancellation-aware via `recordStorageError(err)` using `errors.Is` on the returned error (matching the existing `context.Canceled` treatment in the query/puller paths) rather than a `ctx.Err()` check. Please take another look.

/gemini review

xe-nvdk:
@gemini-code-assist Round-3 findings addressed: `recordStorageError` now takes `ctx` and checks `ctx.Err()` in addition to `errors.Is`, catching client-disconnect failures that surface as plain EPIPE/connection-reset without wrapping the context error. Applied at all cloud-backend error sites and at the local backend's two streaming-copy sites. Please take another look.

xe-nvdk:
@gemini-code-assist Round-4 findings addressed: the buffered `Read` methods now count partially-transferred bytes on `io.ReadAll` failure, consistent with `ReadTo`/`ReadToAt` — applied to S3, Azure, and the local backend (`os.ReadFile` has the same partial-data-on-error contract). Please take another look.

/gemini review

Issue #300, opened 2026-03-26
high(mqtt): No database name validation on MQTT topic mapping


xe-nvdk:
subscriber.go:300-306 — path traversal risk. Fix: validate at creation and message time.

xe-nvdk:
Re-verified against `main` (`4a767b1`). **The code moved, the gap is real, but the severity is wrong — this is not path traversal.** Details below so the next person does not have to re-derive it.

## The cited location is stale

`subscriber.go:300-306` is now `onConnect`. The topic→database mapping lives at **`internal/mqtt/subscriber.go:346-351`**:

```go
database := s.config.Database
if s.config.TopicMapping != nil {
    if mapped, ok := s.config.TopicMapping[msg.Topic()]; ok {
        database = mapped
    }
}
```

## The validation gap is real

`Subscription.Validate()` (`internal/mqtt/subscription.go:185-213`) checks `Database != ""` and explicitly guards the three TLS **certificate paths** against `..` — but **never validates `TopicMapping` values at all**. Confirmed by grep: `TopicMapping` is only assigned (`manager.go:164,279`), serialised (`repository.go:243`), and read on the hot path. Nothing checks it.

The value flows unvalidated into `arrowBuffer.Write(ctx, database, records)` → `generateStoragePath()` (`internal/ingest/arrow_writer.go:3655-3669`), which interpolates it straight into a path:

```go
return fmt.Sprintf("%s/%s/%s/%s/%s/%s/%s_%s_%09d.parquet",
    database, measurement, year, month, day, hour, ...)
```

## But it does not escape the storage root — I tested it

`LocalBackend.Write` calls `validatePath` first (`internal/storage/local.go:86-90`), which sanitizes and then enforces containment with `filepath.Rel` against the base path (`:711-736`). I wrote a throwaway probe against a private temp root with a sibling root as the escape target:

```
absolute           err=<nil>          -> contained
plain dotdot       err=<nil>          -> contained
nul split dotdot   err=invalid path: path traversal detected: path escapes base directory
RESULT: nothing escaped into the sibling root
```

Worth noting `sanitizePath` alone is bypassable — it replaces `..` *before* stripping NUL bytes, so `.\x00./` sanitizes to `../`:

```
in=".\x00./../tmp/y" -> sanitized="../_/tmp/y"
```

[...]

PR #501, opened 2026-06-12
fix(tiering): escape paths in router read_parquet builders (#307)

### LINKED ISSUE #307: high(tiering): SQL injection in buildReadParquet via path interpolation
router.go:171-186 — single quotes not escaped. Fix: double single quotes.

xe-nvdk:
Closes #307.

## Summary

The `internal/tiering` router's `read_parquet()`-building helpers interpolated storage file paths into DuckDB SQL with `'%s'` and no single-quote escaping. They now route every path through `sqlutil.EscapeStringLiteral` (`internal/sql/mask.go`) — the same chokepoint the live query path already uses via `quotePath`.

## Important framing (found during review)

These helpers — `buildReadParquet`, `BuildReadParquetExpr`, `BuildMultiTierQuery` — are **not on the live query path**. The only router method production calls is `GetGlobPathsForQuery` (3 sites in `internal/api/query.go`), which returns raw glob strings; the query engine then builds tiered `read_parquet()` calls through its own `buildMultiTierReadParquet` → `quotePath` → `EscapeStringLiteral`, which was **already escaped**. The patched functions have zero production callers.

So this is **defense-in-depth / hardening of currently-unwired builders**, not a live-exposure fix. They appear to be built ahead of a tiered-query-routing path that isn't wired yet — escaping them now means a future caller inherits the protection. The release-notes entry and code comments are framed accordingly (no overstated "live SQL injection" claim).

Two reviewers (general correctness + security checklist) independently confirmed: single-quote doubling is the complete and sufficient escape for DuckDB's plain `'...'` literal in `read_parquet()` argument position (backslashes are literal; `E'...'`/dollar-quoting require a sigil the code never emits); no double-escaping (live callers escape downstream independently); no import cycle.

## Also addressed

- `BuildMultiTierQuery`'s `whereClause` is interpolated **raw** — documented as a separate injection sink a future caller must validate (or rework to structured predicates). Not escaped here because it's a SQL fragment, not a path literal.
- The real-world taint path — unvalidated MQTT topic-mapping database/measurement names — remains open and is tracked in **#300**. HTTP ingest already validates names (`isValidMeasurementName`), so #307 is not reachable from the HTTP path regardless.

## Test plan

- [x] `go build ./cmd/... ./internal/...`, `go vet ./internal/tiering/`, `gofmt -l` clean
- [x] `go test -race ./internal/tiering/ -count=1` passes
- [x] New `router_test.go` pins the escape contract (single/multi-path, quote-bearing, injection-shaped); verified to FAIL without the escape
- [x] Internal config-matrix + deep review + security-checklist review completed (no Blockers/High code findings; framing corrected per review)


xe-nvdk:
@gemini-code-assist Round-1 finding addressed in the latest commit: `buildFullPath` now returns a `local` (hot-tier) path unchanged instead of prefixing it with the cold backend URI, with a test pinning the guard.

For the record on exploitability: this is latent, not a production bug. The only caller of `buildFullPath` is the unwired `BuildMultiTierQuery`; the live tiered-query path is `buildMultiTierReadParquet` in `internal/api/query.go`, which constructs hot (`getStoragePath`) and cold (`storage.GetStoragePath`) paths separately and correctly and never calls the router builders. So hot+cold queries work today — the fix prevents the bug from surfacing if these builders are ever wired up. Please take another look.

xe-nvdk:
@gemini-code-assist Round-2 findings addressed: `BuildMultiTierQuery` and `buildFullPath` now guard against a nil manager. Accepted specifically because these are exported, unwired builders intended for a future caller who won't have NewRouter's manager-always-set invariant in hand — a nil-deref panic on a public method is a sharp edge worth removing even though production (via NewRouter) never hits it. Please take another look.

Issue #504, opened 2026-06-15
Cluster auth hardening follow-ups: HMAC domain separation + don't forward client auth headers to peers


xe-nvdk:
Two pre-existing cluster-auth defense-in-depth items surfaced during the GHSA-p378-jp5r-gpgw fix review (PR for that fix closes the reported vuln; these are follow-ups, not regressions it introduced).

## 1. HMAC lacks per-message-type domain separation

`security.ComputeHMAC`/`ValidateHMAC` (internal/cluster/security/auth.go) sign `nonce \x00 nodeID \x00 clusterName \x00 timestamp` with **no message-type prefix**. Three handlers use this identical MAC: join, leave, and (as of the GHSA-p378 fix) heartbeat. There is no nonce-replay cache on this path (only `ReplicateSync` uses `nonceCache`), so within the 5-minute timestamp tolerance a captured `(AuthNonce, AuthTimestamp, AuthHMAC)` triple from any one of the three is a **valid MAC for the other two**.

Consequence: an on-path attacker who captures a signed *heartbeat* from node X can replay those auth fields inside a **leave** message for X → `handleLeaveNotify` (on the leader) calls `RemoveServer`+`RemoveNode`+`Unregister`, evicting X. This is gated by `cluster.tls_enabled` (the in-code warning at coordinator.go:207-208 already tells operators TLS is required to keep MACs off the wire), so it requires a no-TLS cluster + on-path position.

**Fix:** add a per-message-type domain-separation prefix to the signed message (the cache-invalidate variant already binds a distinguishing literal). Cover join/leave/heartbeat together. Bump no wire-compat concern if done in a coordinated release (all nodes same version).

## 2. doForward forwards client `Authorization`/`x-api-key` headers to peers

`internal/cluster/router.go` `doForward` copies **all** client headers — including `Authorization` and `x-api-key` — to the peer it forwards to (router.go:~354). This was the third item in GHSA-p378 (we could NOT reproduce it: in standard writer/reader topology nodes service ops locally and don't forward to a freshly-joined node). The GHSA-p378 fix (fail-closed on empty secret) removes the rogue-peer precondition, so this isn't independently exploitable today, but blindly forwarding caller credentials to peers is worth hardening defensively:
- Option A: strip client auth headers on forward and re-authenticate the forwarded request as the cluster identity.
- Option B: only forward to peers whose identity has been verified.

## Severity
[...]

PR #505, opened 2026-06-15
fix(cluster): require shared secret for clustering, authenticate heartbeats (GHSA-p378-jp5r-gpgw)

xe-nvdk:
Addresses GHSA-p378-jp5r-gpgw (private advisory). Ships in 26.06.2.

## Summary

An Arc Enterprise cluster started with `cluster.enabled=true` but an empty `cluster.shared_secret` did not authenticate inter-node coordinator messages — the HMAC checks on join/leave were gated behind a non-empty secret, and heartbeats carried no auth fields at all. Reproduced on a licensed cluster: an unauthenticated join is accepted and the joined node is added as a Raft voter. (The credential-forwarding impact in the original report is topology-dependent and was not reproduced.)

## Fix

- **Fail closed.** A node with `cluster.enabled` and no `shared_secret` now refuses to start, mirroring the guard already in place for peer replication. This is only reachable by a clustering-licensed instance (single-node/OSS hit the standalone fallback above it), so it never affects non-cluster deployments. Because no secret-less cluster can start, every coordinator-protocol handler (join, heartbeat, leave, replicate, forward-apply) becomes unreachable in the vulnerable config.
- **Authenticate heartbeats.** Added `AuthNonce/Timestamp/HMAC` to the `Heartbeat` message, signed in `sendHeartbeats` and validated in `handleHeartbeat`, bringing heartbeat to parity with the already-authenticated join/leave path.

## Scope notes

- Pre-existing defense-in-depth follow-ups (per-message-type HMAC domain separation; `doForward` forwarding client auth headers to peers) are tracked in #504 — both are gated by `cluster.tls_enabled` and out of scope here.
- Rolling-upgrade behavior is documented in the release notes (upgraded nodes require signed heartbeats; complete the rollout in one window).

## Test plan

- [x] `go build ./cmd/... ./internal/...`, `go vet`, `gofmt -l` clean
- [x] `go test -race ./internal/cluster/... ./internal/config/...` — all pass
- [x] New `TestHandleHeartbeat_*` regression tests (reject missing/bad HMAC, accept valid); kill-switch verified (reject tests fail without the guard)
- [x] Internal config-matrix + Step 1b failure-semantics table + deep reviewer + security-checklist reviewer (no Blockers; 2 Medium → release note + #504)
- [x] **Binary smoke on the repro harness:** (a) empty-secret cluster refuses to start; (b) secured 4-node cluster forms healthy with 0 heartbeat rejections between legit nodes; (c) unauthenticated join and heartbeat are rejected

Reported by @sondt99 via coordinated disclosure.


xe-nvdk:
@gemini-code-assist Round-1 finding addressed: `sendHeartbeats` now logs at Error level when `GenerateNonce()` fails (heartbeat ships unsigned → peers reject → operator gets a breadcrumb instead of a silently-flapping node). Also applied the same fix to the pre-existing `leave` signing path, which had the identical silent swallow. Join already returns an error on nonce failure. Please take another look.

PR #506, opened 2026-06-15
fix(api): require admin auth on /api/v1/logs (GHSA-m3qr-fvp4-78xj)

xe-nvdk:
Addresses GHSA-m3qr-fvp4-78xj (private advisory). Ships in 26.06.2.

## Summary

`GET /api/v1/logs` returned buffered application logs **without authentication** even with auth enabled (the default). The route was registered in `Server.RegisterRoutes()` among the public base routes — which runs *before* the global auth middleware (`app.Use`) is installed. Fiber v2 matches routes in registration order and returns on first match before reaching a later middleware, so the route bypassed auth. Reproduced on a running binary (unauth 200 → after fix 401).

## Fix

- Removed `/api/v1/logs` from `RegisterRoutes`; added `Server.RegisterLogsRoute(authManager)`, called from `main.go` **after** `app.Use(auth.NewMiddleware(...))`. With auth enabled it wraps the handler in `auth.RequireAdmin`; with auth disabled (`authManager == nil`) it registers unguarded, preserving no-auth deployments.
- Explicitly whitelisted the `/api/v1/metrics` prefix. Those JSON metrics endpoints were *also* registered pre-middleware, but they expose only aggregate counters + Go runtime/host-arch stats (equivalent to the public Prometheus `/metrics` surface — no queries, DB/measurement names, paths, or credentials) and are intentionally public. Codifying it makes their public status an auditable decision rather than an accident of route ordering.

## Scope / severity

Exposure was operational log metadata only — component names, source locations, log messages, request outcomes. Arc does **not** write query text, database/measurement names, or credentials to this buffer, and the first-run admin-token banner uses stderr (not the logged buffer). Reconnaissance-grade info disclosure, not user-data or secret exposure.

Known limitation (recorded for the advisory, not changed here): the log buffer is global and `admin` is a global permission, so in a multi-tenant deployment an admin sees all orgs' log lines. Also pre-existing: logs access isn't captured by the audit middleware (registered later in the Fiber stack).

## Test plan

- [x] `go build -tags=duckdb_arrow ./cmd/... ./internal/...`, `go vet`, `gofmt -l` clean
- [x] `go test ./internal/api/...` — all pass
- [x] New `TestLogsEndpoint_RequiresAdmin` (unauth 401 / read 403 / admin 200; health + metrics stay public) + `TestLogsEndpoint_AuthDisabledStaysOpen` (no-auth deploys unaffected); kill-switch verified
- [x] Internal config-matrix + Step 1b + deep reviewer + security-checklist reviewer (no Blockers/High)
- [x] **Binary smoke:** logs unauth→401, admin→200+content, read→403, `/health`+`/api/v1/metrics`→200, auth-disabled logs→200
- [x] Docs updated (docs.basekick.net branch `docs/logs-endpoint-admin-auth`)

Reported by @sondt99.


xe-nvdk:
@gemini-code-assist Round-1 findings addressed: `RegisterLogsRoute` now uses the existing `withAdminAuth` helper (identical nil-handling, consistent with the rest of the package), and both `app.Test` responses `defer resp.Body.Close()`. Please take another look.

xe-nvdk:
@gemini-code-assist Round-2 finding addressed: both `app.Test` calls now use the default 1s timeout instead of `-1` (no infinite hang on a deadlock). Please take another look.

PR #507, opened 2026-06-15
fix(cluster): add per-message-type HMAC domain separation labels (#504)

### LINKED ISSUE #504: Cluster auth hardening follow-ups: HMAC domain separation + don't forward client auth headers to peers

xe-nvdk:
Closes #504.

## Summary

Adds per-message-type domain-separation labels to the join-family HMAC (`ComputeHMAC`/`ValidateHMAC`) to prevent cross-message-type replay attacks.

**Attack closed:** a heartbeat MAC captured by an attacker on the network could previously be replayed as a node-evicting leave within the 5-minute freshness window — the join/leave/heartbeat HMAC formats were identical. Each message type now binds a distinct label (`"join"` / `"leave"` / `"heartbeat"`) as the first NUL-delimited field of the signed payload, following the same domain-separation pattern `ComputeCacheInvalidateHMAC` already uses.

## Changes

- **Three exported constants** (`MsgTypeJoin`/`MsgTypeLeave`/`MsgTypeHeartbeat`) in `security/auth.go` guarantee sign/validate side agreement.
- **All 6 coordinator call sites** (3 sign + 3 validate) now thread the appropriate constant.
- Renamed `computeJoinHMACRaw` → `computeJoinFamilyHMACRaw` (handles all three message types, not just join).
- **`TestJoinFamilyHMAC_LabelBinding_NoCrossMessageTypeReplay`**: exhaustive cross-type regression test covering all 6 label combinations plus sanity checks.
- Updated determinism test pinned hash for the new message format.
- Release notes: clarified that join/leave HMAC format also changed, requiring lockstep cluster upgrade.

## Test plan

- [x] `go vet ./internal/cluster/...` — clean
- [x] `gofmt` — clean
- [x] All `internal/cluster/security` tests pass (including new cross-label regression test)
- [x] All cluster auth tests pass
- [x] Configuration matrix reviewed: OSS standalone never reaches this code; cluster-mode all 6 call sites correctly gated

## Rolling-upgrade note

Upgrade all cluster nodes in one window (lockstep). During the mixed-version window, join/leave/heartbeat HMAC validation fails across version boundaries — the old format (no label) and new format (labeled) produce different MACs. This clears automatically once all nodes are upgraded.


xe-nvdk:
Addressed Gemini feedback: defined a dedicated `MsgType` type (instead of untyped string constants) for compile-time enforcement that only valid message-type labels are passed to `ComputeHMAC`/`ValidateHMAC`.

All tests pass, build clean, vet clean.

xe-nvdk:
@gemini-code-assist review again please

PR #513, opened 2026-06-21
fix(security): S3 credentials into DuckDB secrets manager; escape read_parquet paths

xe-nvdk:
## Summary

Fixes three security findings surfaced (and the rest refuted) while verifying an external audit report.

- **H3 — S3 secret key was readable through the query API.** Credentials were configured via `SET GLOBAL s3_secret_access_key=…`; DuckDB exposes settings to plain SQL, so any authenticated principal could run `SELECT current_setting('s3_secret_access_key')` and exfiltrate the key. S3 credentials now go into DuckDB's **secrets manager** via `CREATE OR REPLACE SECRET arc_s3 (TYPE S3, …)` — the same mechanism Arc already used for Azure. The secret value is unreadable via `current_setting()` and redacted in `duckdb_secrets()` (only the non-secret access key ID remains visible). The secret is TEMPORARY (in-memory, pool-wide, never written to disk). Covers the startup path (`configureS3Access`) and the runtime tiering path (`ConfigureS3`) through a shared `buildS3SecretSQL` helper. Azure aligned to `CREATE OR REPLACE` for parity.
- **H2 / H10 — two `read_parquet('%s')` interpolations did not escape single quotes** (continuous-query aggregation runner, parallel partition executor). DuckDB can't parameterize `read_parquet()`; both sites now double embedded quotes, matching the rest of the query engine. H2 uses `sqlutil.EscapeStringLiteral` (the API-layer canonical helper); H10 keeps inline doubling consistent with `internal/query` (which has no `sqlutil` dependency).

The exposure for H2/H10 was narrow (paths derive from validated db/measurement names) — defense-in-depth per the documented convention. H3 is the substantive fix.

## Audit verification

Most of the report's critical/high findings did **not** hold up under code tracing:

| ID | Claim | Verdict |
|----|-------|---------|
| C1 | Cluster GET endpoints unauthenticated | **Refuted** — behind global auth middleware (registered before the handlers); not on the public whitelist |
| H1 | Scheduler GET no auth | **Refuted** — same |
| **H3** | S3 key via `current_setting` | **Confirmed — fixed here** |
| H4 | Azure key same | **Refuted** — already `CREATE SECRET` |
| H5 | HKDF outside FIPS | **Refuted** — stdlib `crypto/hkdf`, in-boundary |
| H8 | Default bind `0.0.0.0` | **Refuted** — default is empty (dual-stack), documented |
| H11 | Mixed TLS mode | **Refuted** — independent by design, startup warning |
| **H2/H10** | `read_parquet` not escaped | **Confirmed — fixed here** |

(C2/C3/H6/H7 were by-design or low/theoretical; details in the branch discussion.)

## Test plan

- [x] `gofmt -l` clean (changed files); `go vet` clean (affected packages)
- [x] `go test ./internal/database/... ./internal/api/... ./internal/query/...`
- [x] `TestBuildS3SecretSQL` — escaping + conditional-field unit test
- [x] `TestS3SecretNotReadable` — **real DuckDB**: asserts `current_setting('s3_secret_access_key')` returns NULL and `duckdb_secrets()` redacts the secret value
- [ ] (optional follow-up) MinIO round-trip confirming an `s3://` read still authenticates through the secret

## Notes

- `ConfigureS3` (runtime tiering reconfigure) currently has **no callers** — converted for correctness if ever wired; documented the empty-credential caveat.
- No code reads the now-unset `s3_*` global settings anywhere (verified by grep); the compaction subprocess configures its own credentials.


xe-nvdk:

Focus areas for this change:
- **H3** (`internal/database/duckdb.go`): S3 credentials moved from `SET GLOBAL s3_secret_access_key` to `CREATE OR REPLACE SECRET arc_s3 (TYPE S3, ...)` via the new `buildS3SecretSQL` helper. Please scrutinize: (a) is quote-doubling via `escapeSQLString` complete for DuckDB `CREATE SECRET` string literals, (b) is the single unscoped TEMPORARY secret visible across the `database/sql` connection pool and does it cover both `s3://` and custom-endpoint (MinIO) URLs, (c) any code path that still depends on the now-unset `s3_*` global settings.
- **H2/H10**: `read_parquet()` single-quote escaping in the CQ runner and parallel executor.

The original audit's C1/H1 ("unauthenticated cluster/scheduler GET endpoints") were refuted — those routes are behind the global auth middleware registered before the handlers. Flag if you disagree.

xe-nvdk:
Pushed 0b740c9 addressing the credential-chain finding — and going further than the suggested patch after an adversarial pass found two gaps in the naive version.

**What changed:**
- `buildS3SecretSQL` is now three-way instead of two-way:
  - both keys set → static-key secret (unchanged)
  - both empty → `PROVIDER CREDENTIAL_CHAIN` (your suggestion) — IAM roles / IRSA / env creds now work
  - **exactly one set → error**. The naive `accessKey != "" && secretKey != ""` condition would route a half-supplied pair (e.g. access key set, secret key forgotten) silently into the credential chain, discarding the provided key and authenticating as a *different* identity (host instance role) with no signal. This case is reachable today via the cold-tier path, whose gate checked only `AccessKey`. It's now a hard startup error.
- Verified against the bundled DuckDB that `CREATE SECRET (TYPE S3, PROVIDER CREDENTIAL_CHAIN, REGION …, ENDPOINT …, URL_STYLE 'path', USE_SSL false)` is accepted — so the MinIO-with-env-creds combo doesn't regress.
- Cold-tier S3 gate (`cmd/arc/main.go`) relaxed from `S3AccessKey != ""` to `Enabled && Backend == "s3"` so credential-chain cold tiers actually activate. Primary storage still requires explicit credentials (unchanged).
- Added tests for the credential-chain output and the asymmetric-key error; added your credential-chain test case in spirit.

Note on the refuted findings from the original audit (C1/H1 "unauthenticated cluster/scheduler GET endpoints"): those routes sit behind the global auth middleware, which is registered *before* the handlers — they are not public. Happy to be shown otherwise.

xe-nvdk:
Pushed 87731a9 for the `ReplaceAllString` `$`-expansion finding (continuous_query.go).

Used `ReplaceAllLiteralString` instead of escaping `$` → `$$`: the replacement is a fixed string, so the literal variant inserts it verbatim and eliminates the entire submatch-expansion class rather than papering over it. Verified via an adversarial pass that this is non-regressing — for any `$`-free input the two functions are byte-identical, and match count/positions are unchanged (the difference is *only* the Expand pass on the replacement).

Out of scope (noted, not fixed here): paths containing DuckDB glob metacharacters (`[`, `*`, `?`) would be glob-expanded by `read_parquet` — but that affects every `read_parquet` call site in Arc, predates this PR, and isn't introduced by it. Tracking separately.

Your earlier credential-chain finding (duckdb.go) and its test were already handled in 0b740c9 (with the asymmetric-key hardening on top).

xe-nvdk:
Pushed 2fa5609 for the critical cold-tier finding — good catch, this was real and reproducible.

Confirmed the chain: local primary storage → `configureS3Access` never ran → httpfs never loaded → runtime `ConfigureS3` → `CREATE SECRET (TYPE S3)` fails, and the sandbox lockdown blocks loading httpfs at runtime so it can't recover. My earlier relaxation of the cold-tier gate (`Enabled && Backend=="s3"`) actually widened this from "static-key cold tiers" to any S3 cold tier, so fixing it was necessary.

Fix: extracted `ensureHTTPFSLoaded` and call it at startup whenever `ColdS3Bucket` is set, even when primary is local/Azure (no primary secret created in that case). Added `TestColdTierS3SecretWithLocalPrimary` which drives the **real `New()` startup path** (local primary + cold bucket, sandbox locked down) and then calls `ConfigureS3` — it fails without the fix and passes with it.

Implementation note slightly different from the suggested one-line condition: rather than calling `configureS3Access` for the cold-only case (which would try to build a *primary* secret from empty keys), I load httpfs separately so the primary secret is only created when primary keys exist.

xe-nvdk:
Pushed fdfe4c4 for the Azure credential-chain dead-code finding.

Confirmed: the `PROVIDER CREDENTIAL_CHAIN` branch in `configureAzureAccess` was unreachable because the gate required both account name AND key. Relaxed it to `AzureAccountName != ""`, so account-name-only config now provisions a credential-chain Azure secret (managed identity / `az login` / env) — matching the S3 credential-chain behavior added earlier in this PR. Verified both Azure secret forms compile against the bundled DuckDB; added `TestAzureCredentialChainStartup` driving the real `New()` path with name-only config. Updated the `azure_enabled` log field and release notes too.

That clears all findings from this round. Summary of the PR now:
- H3: S3 creds → DuckDB secrets manager (not query-readable)
- H2/H10: read_parquet single-quote escaping; CQ also uses ReplaceAllLiteralString for `$`-safety
- S3 + Azure credential-chain support (IAM roles / IRSA / managed identity), with half-configured key pairs rejected
- httpfs pre-loaded for cold-tier S3 on local-primary deployments

xe-nvdk:
Pushed dd83b2a for the empty-endpoint finding. Gated the `ENDPOINT` clause on the *stripped* value so a malformed `"http://"`/whitespace endpoint is treated as "no endpoint" rather than emitting `ENDPOINT ''`. (Verified DuckDB actually accepts `ENDPOINT ''` rather than erroring — so this is semantic hardening, not a crash fix — but omitting it is still correct.) Added a unit test case.

xe-nvdk:
Pushed 9f0fb7e for the two latest findings.

**1. Primary S3 gate (duckdb.go:417) — fixed.** Changed `&&` → `||` so a half-configured primary credential pair reaches `buildS3SecretSQL` and is rejected at startup, instead of being silently skipped. Confirmed via analysis that no working deployment regresses: Arc only reads `ARC_*`-prefixed env into `cfg.S3{Access,Secret}Key`, and DuckDB's `CREDENTIAL_CHAIN` only fires when *both* are empty, so the "one explicit key + AWS env for the other" mode never functioned. `TestHalfConfiguredPrimaryS3Fails` drives the real `New()` path and asserts startup fails.

**2. CQ source rewrite (continuous_query.go:704) — refactored + $-safety kept, but respectfully declining the regex widening.** I extracted `wrapSourceMeasurement` and adversarially tested widening the pattern to `JOIN`/comma-joins/quoted identifiers as suggested. Widening *corrupts valid CQs*:
- comma lead-in (`,\s*`) rewrites column refs in scalar positions — `SELECT a, db.m AS x`, `GROUP BY a, db.m`, `foo(a, db.m)` → `read_parquet(...)` in a scalar slot → broken SQL;
- no trailing boundary mangles a three-part name — `FROM db.m.field` → `read_parquet(...).field`.

Key distinction: this is a **rewriter**, not a gate. A *miss* fails safe (unmatched → loud query-time "table not found"); an *over-match* silently corrupts. So the safe direction is to stay narrow. A correct broad rewrite needs a real SQL tokenizer (literal masking + genuine table-position detection + trailing boundary), which I'm tracking as a follow-up under #491 rather than bolting onto a security PR. I kept the `FROM`-only scope and added **negative-assertion tests** pinning that comma/GROUP BY/qualifier shapes are left unchanged, so a future widening can't silently regress.

xe-nvdk:
Pushed b9a9308 — one substantive change since your last (clean) review, please take a look.

**Scoped per-tier S3 secrets (fixes a clobber found in internal full-PR review).** The previous single shared `arc_s3` secret meant configuring an S3 cold tier overwrote the primary S3 secret (the two tiers can use different buckets/accounts). Now primary and cold use separate SCOPE-bound secrets — `arc_s3_primary` scoped to the primary bucket/prefix, `arc_s3_cold` to the cold bucket/prefix — and DuckDB resolves the right credentials per `read_parquet()` path. `buildS3SecretSQL` now takes `s3SecretParams{name, scope, ...}`; `ConfigureS3` gets the cold bucket/prefix; scope reuses the sandbox `s3PrefixURI` normalization so secret scope matches the allowlist entry. Verified against the bundled DuckDB (`TestScopedSecretsCoexist`: both secrets coexist, `which_secret` resolves each path to its own secret). Release notes updated.

Please focus on: the scope-build path (s3SecretScope empty-bucket → unscoped), and whether the cold ConfigureS3 still composes correctly with the startup httpfs-preload + sandbox lockdown ordering.

xe-nvdk:
Pushed 76b715a — added the defensive `s3cfg == nil` guard in `ConfigureS3` as suggested. (The sole caller always passes non-nil, but it's an exported method, so the guard is cheap insurance.)

Also ran the built binary (OSS standalone, local storage, no S3/Azure/cold) end-to-end to confirm the new branches don't fire spuriously and startup completes clean — DuckDB initialized with `s3_enabled:false`/`azure_enabled:false`, sandbox locked down, test query OK, 0 panics. The S3/cold/Azure-configured paths are covered by real-DuckDB tests (`TestScopedSecretsCoexist`, `TestColdTierS3SecretWithLocalPrimary`, `TestHalfConfiguredPrimaryS3Fails`, `TestAzureCredentialChainStartup`).

xe-nvdk:
Pushed 36e3c34 — implemented the Azure parity you suggested, with one correction to the framing.

You're right that Azure should get the same treatment, and I did it: primary and cold Azure now use distinct container-scoped secrets (`azure_secret_primary`, `azure_secret_cold`) via new `buildAzureSecretSQL` / `ConfigureAzure` helpers mirroring the S3 ones, plus a startup branch that loads the azure extension when only a cold Azure tier is configured.

One nuance from tracing the code: there was no *actual clobber* on Azure today, because (unlike S3's `ConfigureS3`) there was **no runtime cold-Azure credential path at all** — DuckDB only ever received the primary `azure_secret` at startup. So the real pre-existing gap was that **cold-tier Azure with a different container/account simply wasn't queryable by DuckDB**. This PR both closes the clobber risk (separate scoped secrets) and adds the missing cold-Azure query path.

Verified against the bundled DuckDB (`TestAzureScopedSecretsCoexist`: both secrets coexist, `which_secret` resolves each container to its own secret) + unit tests for connection-string vs credential-chain, missing-account-name error, and quote escaping. Release notes updated.

Issue #312, opened 2026-03-26
high(ingest): groupByHour wrong for pre-epoch timestamps


xe-nvdk:
arrow_writer.go:2842 — truncation-toward-zero division. Fix: floor division.

PR #514, opened 2026-06-22
perf(ingest): radix flush-sort + defer per-hour bucketing

xe-nvdk:
## Summary

Two flush-path ingestion optimizations found by profiling sustained MessagePack-columnar load (5M-row buffer / 100ms age, the README hot-path config):

- **Radix flush-sort.** The time-only sort that orders each flush before writing Parquet used a closure-based `sort.Slice`. Under concurrent producers the merged buffer is effectively unordered (~2.5M sorted runs in a 5M-row buffer), costing ~6.9% of ingest CPU. Replaced with an LSD radix sort over the int64 microsecond timestamps — with a sign-flip bias (`radixSortBias`) so pre-1970/negative timestamps (#312) order correctly, and per-pass skipping of the near-constant high bytes. ~5× faster (796ms → 157ms on 5M rows); sort CPU 6.9% → 2.5%. In-order single-producer ingest still short-circuits to an identity permutation.
- **Defer per-hour bucketing past the single-hour check.** `groupByHour` allocated a per-row index bucket on every flush, then discarded it whenever the data fell in one hour (the common live-ingest case) — ~201MB allocated-and-thrown-away per 5M-row flush. A cheap inline min/max scan now makes the single-vs-multi-hour decision; `groupByHour` runs only on the multi-hour backfill path. Single-hour decision: **201MB → 0 B/op, 16ms → 1.25ms.**

Both reclaim CPU/headroom on the Parquet-encode-bound flush path rather than raising the absolute ceiling.

### Measured (60s sustained IoT bench, MessagePack columnar)

| | before | after |
|---|---:|---:|
| Throughput | ~20.0M rec/s | **~20.9M rec/s** |
| peak→end decay | −17% | **−12%** (flatter) |
| p99 / p999 | 2.85 / 8.46ms | **2.67 / 6.80ms** |

README ingestion table re-measured across all four protocols (June 2026) and 26.06.2 release notes updated.

## Review

Ran the CLAUDE.md deep-review process (config matrix + single deep reviewer) **before** committing. It caught a real correctness bug — the radix sort mis-ordered pre-epoch (negative) timestamps, which Arc accepts (#312). Fixed with `radixSortBias` and guarded with permanent tests. Reviewer found 0 Blockers / 0 High after the fix.

## Test plan

- [x] `go test -race ./internal/ingest/...` — PASS
- [x] `permute_time_test.go`: radix correctness across data shapes (realistic interleaved, random, narrow-range, **pre-epoch-mixed**, **all-negative**, already-sorted) + reference cross-check spanning both radix and comparison paths, incl. negatives
- [x] `grouphour_alloc_test.go`: single-hour alloc regression guard (201MB → 0 B/op)
- [x] Edge cases: empty input, n=1 (positive + negative)
- [x] `go build ./cmd/... ./internal/...`, `gofmt -l`, `go vet` — clean
- [x] 60s sustained bench on real binary, 0 errors, all four protocols re-measured


xe-nvdk:

This is an ingestion-performance PR touching the flush hot path in `internal/ingest/arrow_writer.go`. Two changes:
1. Radix sort replacing `sort.Slice` for the time-only flush ordering (with `radixSortBias` for signed/pre-epoch timestamps).
2. Deferring `groupByHour` bucketing past the single-hour check to drop a ~201MB/flush allocate-and-discard.

Particular areas worth scrutiny: the radix sign-bit handling for the full int64 range, the per-pass skip-when-single-bucket optimization (src/dst pointer correctness), stability vs the old non-stable sort, and that the inline min/max scan is equivalent to the old `groupByHour` min/max for the single-vs-multi-hour decision.

xe-nvdk:

xe-nvdk:
Addressed the min/max scan finding in ba13ab1: the loop now ranges over `times[1:]` (the first element is already seeded into both min and max) and uses `else if` (a value strictly below the running min cannot also be strictly above the running max, since `globalMax >= globalMin` always holds). Verified equivalent to the original two-comparison scan across edge inputs (n=1, duplicates, negatives, MinInt64/MaxInt64) with a throwaway equivalence test, and the full `internal/ingest` suite still passes.

xe-nvdk:
Addressed the latest two findings in c02587e:

**`radixSortBias` hex literal — accepted.** Now `uint64(t) ^ 0x8000000000000000`. Pure clarity, no behavior change.

**min/max `else if` + `times[1:]` — declined, with measurements.** I'd applied this in ba13ab1, then benchmarked it: it's actually **~19–24% slower** on our hot-path data. Live ingest arrives monotonically ascending, so `t < globalMin` is reliably false and `t > globalMax` reliably true — the two unconditional comparisons are near-perfectly branch-predicted and run in parallel, whereas the `else if` makes the second comparison data-dependent on the first and stalls the predictor.

```
BenchmarkMinMaxScan_SingleHour (two-comparison)  ~1.30ms/op   <- production
BenchmarkMinMaxScan_ElseIf      (else-if+[1:])   ~1.60ms/op
(5M rows, -benchtime=50x -count=5)
```

So I reverted the loop to the two-comparison form (with a comment explaining why) and kept both variants as benchmarks in `grouphour_alloc_test.go` to document the gap. The `grouphour_alloc_test.go` benchmark now correctly matches production (the other half of that finding).

xe-nvdk:
Thanks — noting the narrowed suggestion now drops the `else if` (the part the last benchmark showed was the ~25% regression). On the remaining `times[1:]` part alone, I isolated it from the else-if to measure it cleanly (5M ascending, -benchtime=50x -count=5):

```
A  full range, two-comparison (production)   1.249ms   <- current
B  times[1:],  two-comparison                1.247ms
C  times[1:] + else-if                       1.563ms
```

`times[1:]` on its own (B) is a wash vs production (A) — 1.247 vs 1.249ms, within noise; it elides exactly one self-comparison out of 5M, which is unmeasurable. The whole regression was the `else if` (C). Since `times[1:]` buys nothing measurable and the two-comparison form is deliberate (commented in `flushPartitionedData`), I'll leave the loop as-is rather than add a churn commit for a neutral change. Holding the current code at c02587e.

Issue #490, opened 2026-06-06
Make duckdb_arrow the default build path (drop the tag) so default go test exercises shipped code


xe-nvdk:
## Problem

Every shipping build of Arc sets `-tags=duckdb_arrow` (Makefile `GOFLAGS`, `Dockerfile`, `.github/workflows/release-build.yml`), so production runs the Arrow code path. But a bare `go build ./...` / `go test ./...` from a developer shell — or any tooling that doesn't pass the tag — silently compiles the **non-Arrow** path instead. The default developer/test experience exercises code that is *not what ships*.

This is a real foot-gun, not hypothetical. In PR #489 (SHOW-command RBAC gating), the `executeQueryArrow` handler (`internal/api/query_arrow.go`, `//go:build duckdb_arrow`) was missing the SHOW gate that the JSON and MsgPack endpoints had. A default `go test ./...` could not catch it because the file isn't compiled without the tag — it was found only by an internal review that explicitly built with `-tags=duckdb_arrow`. The Arrow path can rot undetected.

Note: `make test` *does* pass the tag (`GOFLAGS=-v -tags=duckdb_arrow`), so the Makefile path is correct. The gap is everything that runs `go test` directly (contributors, IDE test runners, ad-hoc CI steps, review agents).

## Preferred fix: drop the build tag entirely

Per the prior decision (2026-05-12), the recommended approach is to **remove `//go:build duckdb_arrow` and make Arrow unconditional**, rather than teach every tool to pass the tag. Every production build already requires it; the opt-out path is effectively dead.

Files to change (verified present):
- `internal/database/duckdb_arrow.go` — strip `//go:build duckdb_arrow`
- `internal/api/query_arrow.go` — strip the tag
- `internal/api/query_arrow_json.go` — strip the tag
- `internal/api/query_arrow_json_test.go` — strip the tag
- `internal/api/query_msgpack_test.go` — strip the tag (also tag-gated)
- `internal/api/query_arrow_stub.go` (`//go:build !duckdb_arrow`, the no-op `executeQueryArrow`) — **delete**
- `internal/database/` — check for a `!duckdb_arrow` stub counterpart to `duckdb_arrow.go` and delete it too

Then `-tags=duckdb_arrow` in Makefile / Dockerfile / release-build.yml is harmless to leave or can be dropped.

The blocker that deferred this (PR #434, arcx loader) merged on 2026-05-13, so it's unblocked.

## Acceptance criteria

[...]

PR #516, opened 2026-06-22
ci: run the test suite with the shipping build tags

xe-nvdk:
## Summary

Adds a `CI` workflow that runs the Go test suite on every PR and push to `main` **with the same build tags the release uses**, so what we test is what we ship.

- **standard:** `go test -tags=duckdb_arrow ./...`
- **FIPS variant:** `go test -tags=duckdb_arrow,fips ./...` with `GOFIPS140=v1.0.0`
- **`go vet -tags=duckdb_arrow ./...`** (once, on the standard variant)

## Why

The only existing workflow (`release-build.yml`) builds and signs binaries but **runs no tests**. Worse, a bare `go test ./...` (no tag) does **not compile the Arrow query path at all** — the DuckDB driver (`duckdb-go/v2`) gates its Arrow CGO API behind `//go:build duckdb_arrow`:

```
$ go build ./internal/database/      # no tag
internal/database/duckdb_arrow.go:27:26: undefined: duckdb.NewArrowFromConn
```

So automated testing of the code that actually ships was missing entirely. This is the gap that let the PR #489 Arrow-handler RBAC bug through (caught only by a manual tagged build). This workflow makes **tested == shipped**.

## Design notes

- **No `-race`:** bcrypt-based auth tests hard-code 1000ms request timeouts that the race detector's CPU overhead exceeds. These are **not real races** (0 `DATA RACE` markers; the suite passes clean without `-race`). Race coverage can be added later behind a carve-out for the timeout-sensitive `internal/auth` / `internal/api` packages.
- **No gofmt gate:** Go 1.26's gofmt rewrites doc-comment *content* (e.g. mangles the `'it''s'` SQL-escape example in `internal/sql/mask.go` into a smart-quote), so enforcing it in CI would propagate comment corruption. Formatting is left to local tooling / review.
- Mirrors the `build-binaries` matrix in `release-build.yml` (runner, `setup-go@v6` / Go 1.26.4, `CGO_ENABLED=1`, `GOFIPS140`).

## Relation to #490

Issue #490 proposed *dropping* the `duckdb_arrow` tag so default `go test` exercises shipped code. That's **not feasible** — the tag is required by the DuckDB driver (above), and there's no native Go way to bake a default tag into a repo. #490 is **closed**; this PR delivers the underlying goal (test == ship) the correct way: run CI with the tag rather than removing it.

## Test plan

- [x] YAML validates (`yaml.safe_load`)
- [x] Reproduced both jobs locally:
  - standard `go test -tags=duckdb_arrow ./...` → 32 ok, 0 fail
  - FIPS `GOFIPS140=v1.0.0 go test -tags=duckdb_arrow,fips ./...` → 32 ok, 0 fail
  - `go vet -tags=duckdb_arrow ./...` → clean
- [ ] CI green on this PR (the workflow runs on itself)


xe-nvdk:
Closing. This PR surfaced that no CI runs the test suite, but the first run exposed a non-CI-portable test (TestScopedSecretsCoexist depends on ambient AWS config for DuckDB credential-chain validation). Not pursuing the CI workflow right now.

PR #517, opened 2026-06-24
fix(database): credential-chain auth for primary S3 storage (IRSA)

xe-nvdk:
## Summary

- An EKS Enterprise deployment authenticating to S3 via **IRSA** (IAM Roles for Service Accounts — no static keys) could **ingest** to S3 but **queries failed**: the DuckDB query path created an S3 secret only when static keys were set, so a `backend=s3`/`minio` primary store with empty keys created **no secret at all** → `s3://` reads failed with an opaque "Loading external extensions is disabled" error.
- This threads a `S3IsPrimaryBackend` signal into the DuckDB `Config` and widens the secret-creation gate so an S3-compatible primary backend **always** provisions a primary secret: static keys → `KEY_ID`/`SECRET`, empty keys → `PROVIDER CREDENTIAL_CHAIN` (IRSA / IAM role / env), scoped to the primary bucket/prefix so it coexists with a cold-tier secret. Reuses the existing `buildS3SecretSQL`/`configureS3Access` helpers unchanged; the write path already worked via the aws-sdk-go-v2 default chain.
- Hardening folded in from internal review: normalize `storage.backend`/`cold.backend` once at load; reject an S3 primary backend with empty `s3_bucket` (and Azure primary with empty `azure_container`) at config load — before any secret is built — instead of producing an unscoped secret + empty sandbox allowlist and failing at first query; `s3_enabled` startup log now reflects IRSA.

The cold tier already supported the credential chain (#513); this brings **primary storage** to parity.

## Test plan

- [x] `go build ./cmd/... ./internal/...` and `-tags=duckdb_arrow` — both clean
- [x] `gofmt -l` clean, `go vet` clean
- [x] `go test ./internal/config/... ./internal/database/...` pass (new `primary IRSA (no keys) -> scoped credential chain` subtest)
- [x] **Running binary vs MinIO** (`backend=s3` and `backend=minio`, no keys, env creds standing in for the chain): `duckdb_secrets()` → `arc_s3_primary / s3 / credential_chain`; queries return S3 data via **JSON, msgpack, and arrow** endpoints
- [x] Same config on `main`: no secret created, query fails — regression confirmed
- [x] Empty-bucket `backend=s3` aborts at config load (before DuckDB init)

## Known follow-ups (out of scope — pre-existing alias gaps)

- `backend=azblob` omitted from `isSharedBackend` (reconciliation mis-gating)
- cold-tier `backend=minio` omitted from the cold switch
- S3 keys/bucket set but `backend` omitted → silent local writes

A separate Helm PR will add IRSA support (gate cred env vars behind `useIRSA`, ServiceAccount `eks.amazonaws.com/role-arn` annotation).


xe-nvdk:
Addressed the Azure review finding in 193d643.

**Finding:** an Azure primary backend (`azure`/`azblob`) with an empty `azure_account_name` silently created no DuckDB secret (the gate is `AzureAccountName != ""`) and failed at query time.

**Fix:**
- Config-load guard now requires an account name **or** a connection string (plus a container) for an Azure primary backend — fail fast instead of failing at first query.
- Crucially, the account name is required only when **no connection string** is set. An unconditional requirement would falsely reject a valid connection-string deployment, whose Go write path (`internal/storage/azure_blob.go`, first auth case) authenticates from the connection string with the account name empty.
- Went one step further and closed the gap that finding exposed: `buildAzureSecretSQL` now consumes an operator-supplied connection string (emitted as `CONNECTION_STRING`), so a connection-string-only Azure primary can now **query** (`azure://` reads), not just write. Precedence: connection string > account name+key > account name only (`PROVIDER CREDENTIAL_CHAIN`).

Added connection-string test cases; both build tags compile, gofmt/vet/tests clean.

xe-nvdk:
Follow-up 78ede41: extended the connection-string DuckDB-secret support to the **cold tier** for parity with the primary path. `AzureConfig` / `ConfigureAzure` now carry a connection string (threaded from `tiered_storage.cold.azure_connection_string`), so a cold Azure tier authenticated via connection string (no account name) gets a queryable DuckDB secret instead of silently none.

Current PR head covers: primary S3 credential chain (IRSA), backend normalization, S3/Azure empty-config startup guards, and connection-string DuckDB secrets for both primary and cold-tier Azure. Both build tags compile; gofmt/vet/tests clean.

xe-nvdk:
Round 2 addressed.

**Cold-tier Azure connection string (high, duckdb.go:788):** already fixed in 78ede41 (before this review landed) — `AzureConfig` carries `ConnectionString`, `ConfigureAzure` passes it to `buildAzureSecretSQL`, and `cmd/arc/main.go` threads `cold.AzureConnectionString`. No further change needed.

**Cold-tier startup validation (medium, config.go:811):** added in ecf2cfa. When `tiered_storage.cold.enabled` is true, the backend + required fields are validated at config load (mirrors the primary guards): `s3` requires a bucket; `azure` requires a container plus an account name or connection string; any other backend is rejected. Validates exactly the backends the cold runtime switch handles (`s3`/`azure`).

Also added `TestLoad_StorageBackendValidation` covering the primary S3/Azure guards, backend normalization, the Azure connection-string carve-out, and the new cold-tier cases. Both build tags compile; gofmt/vet/tests clean.

xe-nvdk:
Round 3 addressed in 2ae1831.

**Whitespace trimming (medium):** the storage identifiers are now trimmed in-place at config load — S3 bucket/prefix and Azure container/account/connection-string, for both primary and cold tier — before the emptiness checks. This prevents copy-paste whitespace from passing validation and then producing opaque downstream secret-SCOPE / sandbox-allowlist / cloud-client failures. The S3 endpoint was already trimmed via `stripURLScheme`.

Took the pointer fix you flagged: `cold := &cfg.TieredStorage.Cold` so the trims persist to the real struct rather than a copy. Added `TestLoad_StorageValuesTrimmed` asserting the cold bucket is trimmed — which specifically exercises the pointer-mutation-persists path.

Both build tags compile; gofmt/vet/tests clean.

xe-nvdk:
Round 4 addressed in 3f759c1.

**Primary backend not validated against supported set (medium):** the two primary-storage `if` blocks are now a single `switch` with a `default` that rejects any unrecognized `storage.backend` at config load (supported: local/s3/minio/azure/azblob — matching main.go's backend switch). This unifies with the cold-tier validation pattern and folds the per-backend field checks in. main.go already fatal'd on an unknown backend, but this fails fast earlier and in one place.

**Test (medium):** added `invalid primary backend -> error` (gcs) and `local primary backend -> ok` cases.

Verified: the local binary still starts cleanly (backend=local); full config + database suites pass; both build tags compile; gofmt/vet clean.

xe-nvdk:
Ran a deep review over the **entire branch** (`main..HEAD`) to break the round-by-round cycle and fixed the findings together in daaee86. The common thread was **gate-vs-runtime divergence** — startup validation/secret gates that didn't match the conditions the runtime actually uses.

- **Cold-tier validation (was an OSS boot regression):** validated under `Cold.Enabled` alone, but the runtime enters the cold path only under `TieredStorage.Enabled`. A parent-disabled deploy with a leftover/default `cold.enabled=true` + incomplete cold backend would fail startup though the runtime ignores it. Now gated on `TieredStorage.Enabled && Cold.Enabled`; added a regression test for the parent-disabled case (the prior tests had pinned the bug).
- **Azure primary secret:** gated on field-presence regardless of backend; a stray `storage.azure_*` value on a non-Azure-primary deploy provisioned a spurious secret (and a malformed connection string would fail startup). Added `AzureIsPrimaryBackend` (mirrors `S3IsPrimaryBackend`) and gate on it.
- **Trimming completeness:** the comment claimed the S3 endpoint was trimmed via `stripURLScheme`, but that only runs inside `buildS3SecretSQL`; the Go client got the raw value. Now trim `S3Region`/`S3Endpoint`/`AzureEndpoint` (primary + cold) and fixed the comment.
- **Tests/docs:** added invalid/local primary-backend cases, the parent-disabled cold regression, and a connection-string quote-escaping test; release notes updated to the `tiered_storage.enabled` + `cold.enabled` condition.

Verified: both build tags compile, gofmt/vet/tests clean, and I ran the binary for the two highest-risk paths — parent-disabled + invalid cold config now boots; primary-S3 IRSA still creates `arc_s3_primary/credential_chain` and queries S3 after the Azure refactor.

PR #518, opened 2026-06-24
chore(helm): IRSA support for external S3 + bump appVersion to 26.06.1

xe-nvdk:
## Summary

Follow-up to #517 (the primary-S3 credential-chain code fix). This makes the **arc-enterprise Helm chart** able to authenticate to external S3 via the pod's IAM role (**IRSA** — IAM Roles for Service Accounts on EKS) instead of static keys.

- `storage.shared.credentials.useIRSA=true` → chart omits `ARC_STORAGE_S3_ACCESS_KEY`/`SECRET_KEY` env vars so Arc falls back to the AWS credential chain.
- `serviceAccount.{create,name,annotations}` → renders a `ServiceAccount` (new `serviceaccount.yaml`) with the operator's `eks.amazonaws.com/role-arn` annotation; `serviceAccountName` set on writer/reader/compactor pods.
- Validation: IRSA exempt from the static-credential requirement, but **fails fast** if `useIRSA=true` without the role-arn annotation.
- `values.schema.json`: declare `serviceAccount` + `credentials.useIRSA`.
- `Chart.yaml`: `appVersion` **26.05.1 → 26.06.1** (chart was pinned to a stale image); chart `version` 0.1.0 → 0.2.0.
- README: external-S3-with-IRSA usage section.

**Note on versions:** the credential chain on the primary-storage *query* path requires the Arc **26.06.2** binary (#517). On 26.06.1 the write path already uses the chain; bump `image.tag` to `26.06.2` once it ships for full primary-S3 IRSA query reads.

## Test plan

- [x] `helm lint` clean
- [x] `helm template` — IRSA on + role-arn: SA + annotation rendered, `serviceAccountName` on all pods, **no S3 key env vars**
- [x] `helm template` — IRSA on, no role-arn: **install fails** with clear message
- [x] `helm template` — default (bundled MinIO, IRSA off): cred env vars present, default SA — unchanged
- [x] **Live k3s (Rancher Desktop) install:** ServiceAccount created with role-arn annotation; writer/reader pods have `serviceAccountName: arc-arc-enterprise`; S3 key env vars omitted; Arc logs `Using default credential chain for S3 (... IAM role ...)`


xe-nvdk:

xe-nvdk:
Addressed both findings in 10eba1b.

**1 (validation too strict):** the `eks.amazonaws.com/role-arn` annotation is now required **only when `serviceAccount.create=true`** — the case where the values annotation is the role-attachment mechanism. With `create=false` (externally-managed SA, annotated via Terraform/eksctl) or EC2 instance-profile / env auth, the chart no longer blocks install.

**2 (nil-safety):** `serviceAccountName` helper and `serviceaccount.yaml` now defensively `default` the `serviceAccount` map to an empty dict, so a null/omitted `serviceAccount` override renders cleanly instead of nil-panicking.

Verified with `helm template`:
- `create=true` + no annotation → still fails (correct)
- `create=false` + external SA name → renders, `serviceAccountName: <name>`
- `serviceAccount: null` override → renders, defaults to `serviceAccountName: default`

README updated to document the create=true vs external-SA / instance-profile behavior.

xe-nvdk:
Round 2 addressed in bda8830.

**useIRSA with bundled MinIO (high):** confirmed the footgun — `useIRSA=true` omits the S3 cred env vars and Arc falls back to the AWS credential chain, which can't authenticate to the bundled MinIO. The combo previously passed validation silently because the IRSA checks lived inside the `external=true` branch. Added a fail-fast: `useIRSA=true` now requires `storage.shared.external=true`.

Verified with `helm template`:
- `useIRSA=true` + bundled MinIO (external=false) → fails with a clear message
- `useIRSA=true` + external + role-arn → renders, no cred env vars
- default bundled MinIO (no IRSA) → renders, cred env vars present (unchanged)

(The three earlier comments on this file were round 1, already addressed in 10eba1b.)

xe-nvdk:
Round 3 addressed in 80f796c.

**Nil-panic on nested storage.shared (high, both files):** confirmed — Go templates don't short-circuit `and`, so a bare `.Values.storage.shared.credentials.useIRSA` would dereference through nil when an operator's local-mode values file omits `storage.shared` (the bundled `values-local-storage.yaml` preset does exactly that; it renders today only because Helm coalesces `shared` from chart defaults — fragile to rely on).

- `_validation.tpl`: `storage.shared` / `.credentials` defaulted to empty dicts, and all credential checks wrapped in the `mode=="shared"` gate.
- `_helpers.tpl` storageEnv: `.credentials` defaulted to an empty dict (the block is already mode-gated, but explicit + consistent now).

Verified with `helm template`: local mode with `storage.shared: null` renders 3 statefulsets, no error; useIRSA+bundled-MinIO still fails; useIRSA+external and default-MinIO unchanged. Chart lints clean.

xe-nvdk:
Rebased onto current `main` (picks up the merged #517) and ran a deep review over the whole branch. The diff is now purely the Helm changes. Fixes in f57a279:

- **Version coupling (headline):** the chart's default image is 26.06.1, but primary-S3 *query* reads via the credential chain need the **26.06.2** binary (#517). The README claimed IRSA works for writes *and* queries — false on the default image. 26.06.2 isn't published yet, so bumping appVersion now would break `helm install` with ImagePullBackOff; instead the README IRSA section carries an explicit version caveat + an `image.tag: 26.06.2` example, and the appVersion bump is deferred until 26.06.2 ships.
- **Nil-safety completeness:** `objectStorageSecretName` and `useMinioCredKeys` still dereferenced `.storage.shared.credentials.existingSecret` unguarded, so `storage.shared.credentials: null` nil-panicked at render (same class as the earlier rounds, one layer deeper). Both helpers now default the nested maps; a null credentials block reaches the proper validation error instead of crashing.
- **Doc accuracy:** corrected "S3/Azure" → "S3/MinIO" in the shared-storage comments (the chart emits S3 config only).

Verified with `helm template`: local + `storage.shared:null` renders 3 statefulsets; `credentials:null` + external hits the creds-required validation (no panic); IRSA happy path emits no cred env; default MinIO unchanged. Chart lints clean.

The IRSA happy path and the credential-pair contract with the Arc binary were confirmed correct (useIRSA omits BOTH key env vars → no half-pair → credential chain engages).

PR #519, opened 2026-06-25
feat(license): offline (air-gapped) license activation from a signed file

xe-nvdk:
## Summary

Arc Enterprise targets aerospace/defense — production often runs in **GovCloud or fully air-gapped** networks that can't reach `enterprise.basekick.net`. Arc previously only activated **online** (HTTP at startup + every 4h), leaving those deployments unable to enable enterprise features.

This adds an **offline activation path**: point `license.file_path` (`ARC_LICENSE_FILE_PATH`) at a signed license file — the same `{license_file, license_signature}` detached pair the server already returns over HTTP, written to disk — and Arc verifies it **locally with no network** and activates.

The crypto already existed (RSA-2048 pinned public key + `VerifyLicenseFile`); this is file-I/O + startup wiring, reusing the exact verification path as online.

## Changes
- **`internal/license/offline.go`** (new): `OfflineLicenseFile` + `LoadOfflineLicense()` — bounded read (1 MiB cap), verify detached signature via `VerifyLicenseFile`, convert via `ToRuntimeLicense`, return an `offline` Client.
- **`internal/license/client.go`**: offline clients are network-free — `StartPeriodicValidation` no-ops; `Activate`/`Verify` hard-error; `ActivateOrVerify` returns the already-verified license.
- **`internal/config/config.go`**: `license.file_path` key, trimmed (a stray-whitespace path can't silently shadow a valid `license.key`).
- **`cmd/arc/main.go`**: startup switch — `file_path` precedence over `key` (warns when both set); shared `applyLicense()` enforces the core limit on both paths; offline errors **fail closed** (warn + OSS, non-fatal).

## Design decisions
- Offline files are **unbound site licenses** — machine fingerprint not enforced (verified safe: nothing downstream relies on it).
- **Expiry IS enforced** via `deriveStatus` (expired → status≠active → all gates closed).
- **Fail-closed** on every error path (missing / garbage / tampered / wrong-key / expired).

## Test plan
- [x] `offline_test.go`: golden path + reject (tampered, wrong key, missing field, non-JSON, missing file, oversized) + expired-gates-closed — reuses the existing `withTestPublicKey`/`signDetached` seam (real production crypto).
- [x] config test pins the `file_path` trim.
- [x] build + vet + tests **with and without `-tags=duckdb_arrow`**.
- [x] **binary run**: missing file → OSS fallback (non-fatal); both `file_path`+`key` set → "ignoring the key", **no online attempt**.
- [x] config matrix (all 7 cells) + **deep security review** — no bypass / fail-open path found.

## Out of scope (fast-follow)
- Server-side operator tool to **mint** the signed file (`enterprise_activation_server`) — being built next.
- Fingerprint-bound offline licenses.
- Air-gapped docs page (follows the server tool so the documented flow is complete).

### DIFF
--- a/cmd/arc/main.go
+++ b/cmd/arc/main.go
@@ -105,9 +105,68 @@ func main() {
 	}
 
 	// Validate Enterprise License early (before component initialization)
-	// This allows us to apply core limits to DuckDB and ingestion workers
+	// This allows us to apply core limits to DuckDB and ingestion workers.
 	var licenseClient *license.Client
-	if cfg.License.Key != "" {
+
+	// applyLicense logs the verified license and enforces its core limit on the
+	// Go runtime / DuckDB / ingest workers. Shared by the online and offline
+	// paths so both enforce identically.
+	applyLicense := func(lic *license.License) {
+		log.Info().
+			Str("tier", string(lic.Tier)).
+			Str("status", lic.Status).
+			Int("days_remaining", lic.DaysRemaining).
+			Int("max_cores", lic.MaxCores).
+			Time("expires_at", lic.ExpiresAt).
+			Strs("features", lic.Features).
+			Msg("Enterprise license verified successfully")
+
+		if lic.MaxCores > 0 {
+			machineCores := runtime.NumCPU()
+			if machineCores > lic.MaxCores {
+				previousGOMAXPROCS := runtime.GOMAXPROCS(lic.MaxCores)
+				log.Info().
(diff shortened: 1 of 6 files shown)


xe-nvdk:
Closing — built on an incorrect (detached) signature model. The activation server uses an embedded-signature license; offline activation will reuse that same model instead. Superseded by a simpler approach.

PR #524, opened 2026-06-30
content(scripts): PI/AF demo data generators for partner handoff

xe-nvdk:
## Summary

Self-contained scripts to generate synthetic sensor data and load it into Arc, for handing to a partner building a PI-style integration (a custom AF Data Reference) and for reproducing the anomaly-detection blog demo.

Lives in `scripts/pi-demo-data/`:

- **`generate_trend_data.py`** — 3 tags × 365 days, irregular timestamps, ~31.5M rows (`historian.sensors`). The PI-style historian shape for **PlotValues / trend** queries (the AF Data Reference case).
- **`generate_anomaly_data.py`** — 3 machines × 60 days @ 5s with 2 injected faults, ~3.1M rows (`factory.vibration`). The anomaly-detection dataset.
- **`arc_ingest.py`** — shared MessagePack columnar ingest helper (stdlib + `msgpack` + `duckdb`; no Arc client library needed).
- **`README.md`** — prerequisites, generate+load commands, the verified PlotValues query (with AF `PlotValues(...)` mapping), the anomaly queries, and the compaction perf tip.

Both datasets are shaped exactly as Arc stores data natively (columnar Parquet, `time`-partitioned), so query performance is representative.

## Test plan

- [x] `--parquet-only` path runs for both generators (3.1M / 31.5M rows produced)
- [x] Full load verified end-to-end against a live Arc instance (`ARC_AUTH_ENABLED=false`): anomaly dataset loaded 3.1M rows in ~2s via `/api/v1/write/msgpack`
- [x] Loaded data is queryable; per-machine stats show the injected pump-01 fault (avg 4.63 / max 14.37 vs healthy 4.50 / 13.31)
- [x] README's seasonal-residual detector flags exactly the faulty machine over the fault window (96 hours on pump-01, 0 on healthy machines)
- [x] PlotValues query verified separately against the trend dataset earlier (year trend → 1000 bins, ~118ms compacted)
- [x] Confirmed `ARC_AUTH_ENABLED` is the correct env var (viper `ARC` prefix + `.`→`_`)

## Notes

- Not product code — demo/reproduction tooling under `scripts/`. No changes to Arc itself.
- Synthetic data is dated 2025 (the demo reference year); generators have `TIMESTAMP` literals that can be adjusted.

### DIFF
--- a/scripts/pi-demo-data/README.md
+++ b/scripts/pi-demo-data/README.md
@@ -1,0 +1,155 @@
+# Arc PI / AF Data Reference demo data
+
+Self-contained scripts to generate synthetic sensor data and load it into an Arc
+instance, for building and benchmarking a PI-style integration (PlotValues,
+RecordedValues, Summaries) and for the anomaly-detection demo.
+
+Two datasets:
+
+| Script | Dataset | Use for |
+|---|---|---|
+| `generate_trend_data.py` | 3 tags x 365 days, irregular timestamps, ~31.5M rows (`historian.sensors`) | **PlotValues / trend** queries — the AF Data Reference case |
+| `generate_anomaly_data.py` | 3 machines x 60 days @ 5s, 2 injected faults, ~3.1M rows (`factory.vibration`) | Anomaly detection (z-score, MAD, seasonal residual) |
+
+Both shape the data exactly like Arc stores it natively (columnar Parquet,
+`time`-partitioned), so query performance is representative.
+
+## Prerequisites
+
+```bash
+pip install duckdb msgpack
+```
+
+A running Arc instance. Default target is `http://localhost:8000`. To run Arc
+locally with auth disabled (simplest for a demo):
+
+```bash
+docker run -d -p 8000:8000 -e ARC_AUTH_ENABLED=false \
+  ghcr.io/basekick-labs/arc:latest
+```
+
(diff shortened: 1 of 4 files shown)


xe-nvdk:
Thanks for the review. Addressed all findings in 99ede00:

1. **O(N²) LIMIT/OFFSET** → single `con.execute(...)` streamed with `fetchmany(chunk_rows)` (O(N)).
2. **Socket leak** → each `urlopen` wrapped in `with`.
3. **Lost error body** → catch `urllib.error.HTTPError` and surface Arc's response body (urlopen raises on non-2xx, so the prior `resp.status` check never saw error responses).
4. **Transpose** → `zip(*rows)` instead of the nested Python loop.
5. **`epoch_ms(TIMESTAMP)`** → switched both generators to the documented `(epoch(time) * 1000)::BIGINT`. Note: `epoch_ms(TIMESTAMP)` does return correct ms on DuckDB 1.3.2, but you're right that it's not its documented direction (BIGINT→TIMESTAMP) and shouldn't be relied on across versions, especially in scripts handed to a partner.

Re-verified end-to-end against a live Arc instance after the changes: 3.1M rows load at ~1.5M rows/sec, timestamps span the correct window, and the injected pump-01 fault is visible in the per-machine aggregates.

PR #525, opened 2026-07-01
feat(release): build macOS binaries for Homebrew tap

xe-nvdk:
## Summary

- Adds native **macOS binary builds** to the release workflow so `brew install` can work: `darwin-arm64` (macos-14 / Apple Silicon) and `darwin-amd64` (macos-13 / Intel), standard variant only.
- DuckDB links **statically** via CGO — the resulting Mach-O depends only on macOS system libraries (verified with `otool -L`), so the Homebrew formula needs no runtime deps.
- Wires the darwin binaries into the **SLSA sha256 subjects** and the release **`files:` block** (which previously listed only the Linux artifacts, so darwin would build but not attach).
- Drafts `RELEASE_NOTES_2026.06.3.md` with the `brew install basekick-labs/tap/arc` instructions.

This is the prerequisite for the `basekick-labs/homebrew-tap` (published separately). The tap formula points at the `arc-darwin-*` release assets this PR produces.

## Why standard-only (no FIPS darwin)

The FIPS build is Linux/CMVP-focused and distributed as a container image + Linux packages. Homebrew users on macOS want the standard build for local dev/eval. Adding a darwin-FIPS leg would be scope creep with no CMVP relevance.

## Test plan

- [x] `go build -tags=duckdb_arrow` builds + runs natively on macOS arm64 (verified locally; `otool -L` confirms static DuckDB, only system libs)
- [x] Workflow YAML validates; FIPS-only steps (`if: matrix.variant == 'fips'`) correctly skip for the standard darwin legs
- [x] No Linux-only commands in the shared build/sign/upload steps
- [ ] CI green on this PR (macOS runners build + sign the darwin binaries)
- [ ] Next release publishes `arc-darwin-arm64` / `arc-darwin-amd64` assets, then the tap formula checksums are filled

### DIFF
--- a/.github/workflows/release-build.yml
+++ b/.github/workflows/release-build.yml
@@ -340,6 +340,24 @@ jobs:
             artifact: arc-fips-linux-arm64
             build_tags: duckdb_arrow,fips
             gofips: "v1.0.0"
+          # macOS (darwin) binaries for the Homebrew tap (basekick-labs/homebrew-tap).
+          # Standard variant only: the FIPS story is Linux/CMVP-focused, and brew
+          # users want the standard build. DuckDB links statically via CGO, so the
+          # resulting Mach-O depends only on macOS system libs (verified with
+          # otool -L) — no runtime deps for the formula. macos-14 = arm64 (Apple
+          # Silicon), macos-13 = amd64 (Intel).
+          - runner: macos-14
+            suffix: darwin-arm64
+            variant: standard
+            artifact: arc-darwin-arm64
+            build_tags: duckdb_arrow
+            gofips: ""
+          - runner: macos-13
+            suffix: darwin-amd64
+            variant: standard
+            artifact: arc-darwin-amd64
+            build_tags: duckdb_arrow
+            gofips: ""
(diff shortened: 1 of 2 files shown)


xe-nvdk:

xe-nvdk:
Rebased on `main` (which now carries the #526 concurrency fix) and reconciled the shared `RELEASE_NOTES_2026.06.3.md` so 26.06.3 covers both stories: the reliability fix and this macOS/Homebrew packaging work. Only the release-notes file conflicted; the workflow change applied cleanly. Chart appVersion is 26.06.3 via the merged fix.

xe-nvdk:
Good catch on the cask collision — that's a real one. Fixed in `5950467`: the tap+install form now uses `brew install --formula arc` (the fully-qualified `basekick-labs/tap/arc` form was already unambiguous). Confirmed the `arc` cask exists (the Arc browser), so a bare `brew install arc` would indeed error with 'Formulae and casks both provide arc'.

Follow-up (separate repo): the `basekick-labs/homebrew-tap` README should get the same `--formula` correction.

PR #526, opened 2026-07-02
fix(ingest): clone request-derived database/measurement before async retention

xe-nvdk:
## Summary

- Fixes a concurrency data-routing issue: under high-throughput concurrent ingest, a write tagged with one database (via `x-arc-database` header or `db`/`bucket` query param) could be persisted under a **different** database. The misrouted name appeared truncated to the intended name's length (e.g. `energy_grid` → `vessels_tra`/`system_moni`).
- **Root cause:** Fiber's `c.Get`/`c.Query` return zero-copy strings aliasing the fasthttp request buffer, valid only for the handler's lifetime. Arc retains that string past the handler — the async Arrow-buffer flush uses it on a background goroutine as the storage-path directory that selects the database. A concurrent request reusing the fasthttp buffer overwrites the retained bytes → misrouted write. Length-fixed-at-read + bytes-overwritten-later explains the truncation signature.
- **Fix:** `strings.Clone` every request-derived `database`/`measurement` that outlives the handler, at the boundary. Sites: `msgpack.go`, `lineprotocol.go` (handleWrite convergence for all 3 LP entry paths), `tle.go` (db+measurement), `import.go` (LP+TLE import), `import_inprocess.go` (CSV/Parquet via `importPreamble`, db+measurement).
- **Already safe (unchanged):** MessagePack body-decoded values (decoder copies out of the request buffer synchronously); cluster write-forwarding (`BuildHTTPRequest` copies headers via `string(value)`, body `io.ReadAll`'d before forward; writer re-reads a fresh net/http request). Read/query endpoints never retain the header.
- No config, API, or on-disk format changes. Drop-in upgrade. Ships as 26.06.3.

## How it was found & verified

Surfaced while backfilling a demo: a 2-minute burst of ~132 large concurrent msgpack payloads scattered ~60% of `energy_grid` rows into other concurrently-written demos' databases. Diagnosed via storage-path logs (`buffer_key=energy_grid` but `storage_path=vessels_tra/...`), confirmed by a dedicated code-trace agent and an adversarial reviewer (which caught the `importPreamble` gap now included).

## Test plan

- [x] `gofmt -l` clean, `go vet ./internal/api/...` clean
- [x] New regression tests in `internal/api/database_header_aliasing_test.go` — the property test reproduces the **exact** truncation symptom (`got "vessels_tra"`) when the clone is removed, and passes with it; plus a 200-goroutine concurrent multi-DB isolation test under `-race`.
- [x] `go test -race ./internal/api/...` passes (one pre-existing flaky `TestLogsEndpoint_RequiresAdmin` timeout under `-race`, confirmed failing on `main` too, unrelated).
- [x] `go test ./internal/ingest/...` passes.

### DIFF
--- a/RELEASE_NOTES_2026.06.3.md
+++ b/RELEASE_NOTES_2026.06.3.md
@@ -1,0 +1,15 @@
+# Arc v2026.06.3 Release Notes
+
+> **Status:** In progress.
+
+> **This is a patch release.** A single reliability fix — no new features, no breaking API changes, no schema migrations. Drop-in upgrade from v2026.06.2.
+
+## Reliability fix
+
+**Improved database routing under heavy concurrent writes.** This release fixes an edge case where, under sustained high-throughput concurrent ingest, a write tagged with one database (via the `x-arc-database` header or the `db`/`bucket` query parameter) could occasionally be stored under a different database. The affected name showed up shortened to the length of the intended one — for example, writes meant for `energy_grid` (11 characters) landing under `vessels_tra` or `system_moni`, the leading 11 characters of other databases being written at the same moment.
+
+The cause was a subtle string-lifetime detail at the request boundary. Arc's write handlers read the target database (and, on the TLE and import paths, the measurement) using Fiber's `c.Get` / `c.Query`, which hand back lightweight strings that point into the request buffer and are only meant to be used while the handler runs. Arc's ingest path keeps that string a little longer — the background Arrow-buffer flush uses it to pick the storage directory for the database — and if another request reused the buffer in between, the value could change out from under it. Copying the value at the boundary resolves it completely.
+
+The behavior only showed up under concurrency with buffer reuse, so a steady or moderate write stream was very unlikely to see it, while a rapid burst of large payloads alongside other concurrent writers could hit it more often. It touched the MessagePack, line-protocol (native and InfluxDB-compatible), TLE, and bulk-import (CSV / Parquet / line-protocol / TLE) write paths. Reads and queries were never affected. There are no configuration, API, or on-disk format changes, and the fix is a drop-in upgrade.
+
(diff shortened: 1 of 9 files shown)


xe-nvdk:

xe-nvdk:
Thanks for the review. Addressed in `315cb20`.

**Applied (3 findings) — the static-default cases.** In `tle.go` (database + measurement) and `import.go` (measurement), the value was read via `c.Get(key, "<static-default>")`, so cloning the result cloned the static literal on every header-less request. Fixed to clone only the request-provided value and assign the literal directly when absent — this removes the needless allocation while keeping the aliasing protection intact (the request-provided value is still cloned before it escapes to the async flush).

**Declined (4 findings) — the empty-database cases** (`msgpack.go:300`, `import.go:124`/`:385`, `import_inprocess.go:417`/`:427`). These read `c.Get("x-arc-database")` / `c.Query(...)` with **no** default, so the value is either the aliased header (needs cloning) or `""`. `strings.Clone("")` is a zero-allocation no-op — Go's stdlib returns `""` directly without allocating. Benchmarked:

```
BenchmarkCloneEmpty-14      0.82 ns/op    0 B/op   0 allocs/op
BenchmarkCloneNonEmpty-14  13.43 ns/op   16 B/op   1 allocs/op
```

So guarding these with `if x != ""` saves 0 allocations and ~0.8 ns while adding a branch to a security-critical path. I've kept them as unconditional `strings.Clone` for simplicity and safety. Happy to revisit if you see a case I've missed.

xe-nvdk:
Holding on the 4 remaining empty-database suggestions: those sites read the header with no static default, so the value is either the aliased header (cloned) or `""`, and `strings.Clone("")` is a 0-allocation no-op (benchmarked 0.8 ns / 0 B / 0 allocs). Guarding them saves nothing and adds a branch to the security-critical path, so they stay as unconditional clones. No outstanding correctness findings — merging.

PR #530, opened 2026-07-03
fix(cluster): strip client forwarding headers + harden loop guard (CVE-2026-45045 class)

xe-nvdk:
## Summary

- Dependabot flagged [CVE-2026-45045 / GHSA-gcfq-8gqf-4876](https://github.com/advisories/GHSA-gcfq-8gqf-4876) in GoFiber (`X-Real-IP` spoofing via `Header.Add` in `BalancerForward`). **Arc is not affected by the CVE itself:** it does not import GoFiber's `middleware/proxy` (verified — not in the build graph), and nothing on the receive side trusts `X-Real-IP` / `X-Forwarded-*` (identity is socket-only via `c.IP()`; Fiber has no `EnableTrustedProxyCheck`/`ProxyHeader`). There is **no patched `fiber/v2` release** to bump to, so this closes the vulnerability *class* directly in Arc's own code as defense-in-depth.
- **Header stripping (defense-in-depth):** `BuildHTTPRequest` now strips client-controlled forwarding/identity headers before an inter-node forward — `X-Real-IP`, `X-Forwarded-For/Host/Proto/Port`, `Forwarded` (RFC 7239), `X-Arc-Forwarded-By`, `X-Arc-Original-Host`, `X-Arc-Shard-Routed`, and CDN client-IP headers `True-Client-IP` / `CF-Connecting-IP` / `X-Client-IP`. The forwarding node re-establishes the trusted values itself (`router.go`).
- **Loop-guard routing-integrity fix:** `decideForward()` makes the `X-Arc-Forwarded-By` loop guard **capability-first**. A capable node ignores the (client-settable) marker; a non-capable node carrying the marker returns **`508 Loop Detected`** instead of silently attempting a doomed local path. This closes an authenticated-only routing-integrity issue (CWE-290, bounded to routing behavior) reachable in role-separated clusters. The 508 path is unreachable via legitimate forwarding (peers only forward to capable targets).

**Single-node / OSS deployments are unaffected** — the forwarding path only executes with a cluster router.

## Not in scope

Adjacent to #504 item 2 (`doForward` forwarding client `Authorization`/`x-api-key` to peers) but **does not address it** — kept separate to avoid an RBAC regression (the receiving peer re-authenticates and RBAC-checks using the forwarded token). #504 reopened for that follow-up.

## Test plan

- [x] `gofmt -l` clean, `go vet ./internal/api/...` clean
- [x] `go test ./internal/api/... ./internal/cluster/...` pass (routing subset under `-race` too)
- [x] New tests: `TestForwardDecision_ClientCannotForceLocal` (capability-first ordering across reader/writer/compactor × marker), `TestBuildHTTPRequest_StripsClientForwardingHeaders` (all 12 stripped headers + legit header survives); updated `TestBuildHTTPRequest_PreservesMultiValueHeaders`
- [x] Ran the binary single-node: writes (204) + query (200) succeed with a spoofed `X-Arc-Forwarded-By` correctly ignored
- [x] Config matrix (6 rows) + failure-semantics table + deep reviewer + security-checklist reviewer + adversarial check — no Blockers/High/Medium
- [x] Release notes: `RELEASE_NOTES_2026.06.4.md`

Closes the remediation for Dependabot alert #12 (to be dismissed as not-affected + in-code hardening once merged).

### DIFF
--- a/RELEASE_NOTES_2026.06.4.md
+++ b/RELEASE_NOTES_2026.06.4.md
@@ -1,0 +1,53 @@
+# Arc v2026.06.4 Release Notes
+
+> **Status:** In development.
+
+> **This is a security-hardening patch release.** Defense-in-depth on the cluster request-forwarding path, plus a routing-integrity fix for clustered deployments. No breaking API changes, no schema migrations, no configuration changes. Drop-in upgrade from v2026.06.3. **Single-node (non-clustered) deployments are unaffected by every change here** — the forwarding path they harden never executes without a cluster router.
+
+## Security hardening
+
+### Strip client-controlled forwarding headers at the inter-node boundary (CVE-2026-45045 class)
+
+Dependabot flagged [CVE-2026-45045 / GHSA-gcfq-8gqf-4876](https://github.com/advisories/GHSA-gcfq-8gqf-4876) in GoFiber: the `BalancerForward` proxy helper injects `X-Real-IP` with `Header.Add()` instead of `Header.Set()`, appending the real client IP as a *second* header value so upstreams that read the first value trust an attacker-supplied IP.
+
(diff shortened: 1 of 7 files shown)


xe-nvdk:

Focus areas for this security-hardening PR:
- **Header stripping** in `BuildHTTPRequest` (internal/api/routing.go): is the canonical-key form correct for every entry (e.g. `X-Real-IP` → `X-Real-Ip`, `CF-Connecting-IP` → `Cf-Connecting-Ip`)? Are all values of a multi-value header stripped, in any casing?
- **Loop-guard reorder** in `decideForward`: capability checked before the client-settable `X-Arc-Forwarded-By` marker; non-capable + marker → 508. Confirm the 508 is unreachable via legitimate peer→peer forwarding (peers only forward to capable targets).
- **No forward path bypasses the strip** (cluster/router.go doForward, sharding/router.go both consume BuildHTTPRequest output and re-`Set` trusted values).
- **Go control flow**: the handlers switched from `if ShouldForwardX` to a `switch` with `goto localProcessing` from inside a case — legal and no skipped var decls?
- **Doc-vs-code**: RELEASE_NOTES_2026.06.4.md claims vs the code.

PR #534, opened 2026-07-14
fix(iceberg): warehouseRelKey must trim the storage root, not the warehouse

xe-nvdk:
Follow-up to #532, from a Gemini finding on the #533 tracking PR. Real bug, reproduced before fixing.

## Summary

- Storage backends' `Read`/`Write` take keys relative to the **storage root**, but `warehouseRelKey` trimmed `e.warehouse`. Those two bases are the same string only when `iceberg.warehouse` is the storage root — which is the default, and is why every test and every prior binary run missed this.
- Pointing `iceberg.warehouse` at a **subdirectory** (a supported config: the key exists and is passed straight through when set) silently dropped that subdirectory from every key. `version-hint.text` and the `v<N>.metadata.json` reader copies landed at the storage root instead of inside the warehouse, so directory-based readers (DuckDB, Spark) could not resolve the current snapshot.
- Fix: keep gating on the warehouse prefix (unchanged "is this ours?" semantics), but derive the key by trimming `DefaultWarehouse(e.backend)`.
- Also returns `ok=false` when the warehouse sits entirely outside the storage root — the backend cannot address it, and the suggested patch would have produced a key that silently escaped the root.

## Reproduced first

With `warehouse=<root>/warehouse`:

```
before: version-hint.text -> <root>/arc_mydb.db/cpu/metadata/version-hint.text          (leaked to root)
after:  version-hint.text -> <root>/warehouse/arc_mydb.db/cpu/metadata/version-hint.text (correct)
```

## Test plan

- [x] `TestCustomWarehouseSubdir` regression test: hint lands inside the configured warehouse, does **not** leak to the storage root, and the data file is still registered.
- [x] Full `internal/iceberg` suite green under `-race` (default-warehouse path unaffected).
- [x] `gofmt` / `go vet` / `go build -tags duckdb_arrow ./cmd/... ./internal/...` clean.
- [x] **Binary-verified** with `iceberg.warehouse` set to a subdirectory: warehouse writes under `data/warehouse/`, nothing leaks to the storage root, and `duckdb iceberg_scan` resolves via `version-hint.text` and reads 40 rows — the exact path that failed before the fix.

### DIFF
--- a/internal/iceberg/exporter.go
+++ b/internal/iceberg/exporter.go
@@ -541,14 +541,31 @@ func (e *Exporter) writeVersionHint(ctx context.Context, tbl *icetable.Table) {
 	}
 }
 
-// warehouseRelKey converts a full metadata URI to a STORAGE-RELATIVE key (backend Read/Write
-// operate on relative keys). Returns ok=false if the URI isn't under the warehouse root.
+// warehouseRelKey converts a full metadata URI to a STORAGE-RELATIVE key, since backend
+// Read/Write operate on keys relative to the STORAGE ROOT — not to the warehouse. Returns
+// ok=false if the URI isn't under the configured warehouse.
 //
-//	metaLoc="file:///wh/arc_db.db/cpu/metadata/00004-<uuid>.metadata.json", warehouse="file:///wh"
-//	-> "arc_db.db/cpu/metadata/00004-<uuid>.metadata.json"
+// The two bases coincide only when iceberg.warehouse is the storage root (the default), so
+// trimming the warehouse silently produced a key missing the warehouse's own path segment
+// whenever an operator pointed iceberg.warehouse at a SUBDIRECTORY — the version-hint and
+// v<N>.metadata.json copies then landed outside the warehouse, where directory-based readers
+// (DuckDB, Spark) could not resolve the current snapshot. Gate on the warehouse, trim the root.
+//
+//	storage root "file:///data", warehouse "file:///data/wh",
+//	metaLoc "file:///data/wh/arc_db.db/cpu/metadata/00004-<uuid>.metadata.json"
+//	-> "wh/arc_db.db/cpu/metadata/00004-<uuid>.metadata.json"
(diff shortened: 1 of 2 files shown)


xe-nvdk:

xe-nvdk:
Both findings taken in `ffd75e0` — the prefix one is a fair hit on code I had *just* written.

**Path-boundary matching.** Confirmed the bug concretely: with warehouse `file:///data/wh`, a bare `strings.HasPrefix` accepts `file:///data/wh-other/arc_db.db/cpu/metadata/v1.metadata.json` — a *different* warehouse's metadata treated as ours, yielding the key `wh-other/arc_db.db/...`. The same flaw applied to my `rel == metaLoc` root check (misses `/data` vs `/data-other`).

Being straight about severity: it is **not reachable today** — all three callers pass `tbl.MetadataLocation()` from our own catalog, so no foreign path ever reaches it. But this function is the "is this path ours?" boundary check and `pruneOldVersionFiles` **deletes** files under the key it derives, so a gate that can say yes to the wrong directory is not something to leave loaded. Fixed.

I extracted a helper rather than inlining the check at both sites (the trailing-slash normalization then lives in one place):

```go
func isUnderDir(p, dir string) bool {
	dir = strings.TrimSuffix(dir, "/")
	return p == dir || strings.HasPrefix(p, dir+"/")
}
```

`TestIsUnderDir` pins the cases: the dir itself, paths beneath it, a trailing-slash-configured dir, and the sibling prefixes (`wh-other`, `wharf`, `whx`) that must be rejected.

**`localFileURI` in the test.** Also correct — `"file://" + filepath.Join(...)` would emit `file://C:\...` on Windows and mismatch `DefaultWarehouse`. Switched to `localFileURI`, which is exactly what `DefaultWarehouse` uses.

Re-verified on the running binary after the change (the gate is in the live path, so tests alone are not enough): with `iceberg.warehouse` at a subdirectory, the hint lands under `data/warehouse/`, nothing leaks to the storage root, and `duckdb iceberg_scan` reads 40 rows.

@gemini-code-assist please review the follow-up.

Issue #556, opened 2026-07-28
medium(backup): partial-skip storm still reports success (skip ratio threshold)


xe-nvdk:
Split out of #322 / PR #554 (deliberately deferred there — this is a policy decision).

## Problem

PR #554 made backup failure classification positive: only a **source-read** failure is skippable, because the file may legitimately have been removed by compaction or retention between the listing and the copy. That race is real and skipping it is correct.

But `ReadTo` on a perfectly healthy source can also fail for reasons that are *not* a benign race:
- S3/Azure throttling (503 SlowDown) under load
- network timeouts / transient DNS failure
- expired or rotated credentials
- a storage outage partway through a long backup

In those cases every affected file is skipped individually, and the backup **completes successfully** with a subset of files. PR #554 added a guard for the total case — if *every* file is skipped the backup now fails — but a partial storm slips through:

- 10,000 files, S3 throttles 4,000 of them → backup "succeeds" with 60% of the data
- `skipped_files: 4000` is recorded in the manifest and logged as a warning, so it is *visible* — but nothing fails, and an operator with alerting on backup exit status sees green

## Suggested fix

Fail the backup when the skip ratio exceeds a threshold:

```go
const maxSkipRatio = 0.10 // configurable?

if len(files) > 0 && float64(skipped)/float64(len(files)) > maxSkipRatio {
    return fmt.Errorf("backup failed: %d/%d files skipped (>%.0f%%), source storage may be degraded",
        skipped, len(files), maxSkipRatio*100)
}
```

## Open questions (why this is deferred, not just implemented)

1. **What threshold?** 10% is a guess. Too low and heavy compaction churn fails legitimate backups; too high and it never fires.
2. **Config key or constant?** A `backup.max_skip_ratio` key needs `v.SetDefault` in `internal/config/config.go#setDefaults` and would need testing at a non-default value per CLAUDE.md.
3. **Ratio or absolute count?** On a 10-file database, 2 skipped files is 20% but probably fine. On 100,000 files, 1% is 1,000 missing files. Possibly needs both.
4. **Should it be fatal at all,** or a distinct terminal status (`completed_with_errors`) that alerting can key on? That may be more useful than a hard failure — the partial backup still has value.

Worth deciding deliberately rather than picking a number.

## Related

- #322 — the streaming change that introduced skip accounting
- PR #554 — added `skipped_files` to the manifest/progress and the all-skipped guard

Issue #329, opened 2026-03-26
medium(mqtt): Repository opens own SQLite connection


xe-nvdk:
repository.go:40 — extra writer. Fix: accept *sql.DB.

PR #567, opened 2026-07-30
fix(iceberg): address review blockers; merge main (#533)

xe-nvdk:
Targets the **`26.09.1`** branch so it lands inside PR #533 rather than racing it. Merges `main` (23 commits) and fixes everything the review of #533 found.

**The Iceberg subsystem itself is sound.** The config guardrails (refusing S3 backends and cold-tier tiering at load), the compaction-race handling (compaction uploads before deleting, and iceberg-go's commit-time footer read turns a vanished file into a failed commit that correctly retries), and the fingerprint cache all hold up under scrutiny. These are the gaps around it.

## Blockers

**1. `iceberg.catalog_db_path` at a non-default value silently excluded the catalog from backup.**

`main.go` hardcoded `cfg.Auth.DBPath` for the backup manager while Iceberg read `cfg.Iceberg.CatalogDBPath`. Both default to `./data/arc.db` — **the same file at the default, divergent the moment an operator sets the key.** The catalog holds every table's schema and snapshot pointers, so a restore brought back Parquet and warehouse metadata whose tables no longer resolved, contradicting the "backup-aware" claim.

This is the #534 shape, in the package #534 was filed against.

Fixed: backed up as `metadata/iceberg-catalog.db` when it is a separate file, and restored alongside. Older backups without it are skipped, not an error.

**2. A failed `version-hint.text` write was never retried.**

The scheduler short-circuits on an unchanged fingerprint *before* reaching `ReconcileMeasurement`, and cached the fingerprint regardless of whether the best-effort discovery-file writes succeeded. Once the file set went quiet — the steady state — the hint stayed stale **forever**, and directory-based readers (DuckDB `iceberg_scan`, Spark hadoop-format) resolved the wrong version indefinitely. The comment claiming "the next reconcile rewrites these" was false.

Fixed: publishing reports success, a converged pass republishes (that's the path a retry lands on), and the scheduler declines to cache when it failed.

**3. Merging `main` produced a silently broken backup path.**

`backup.go` **auto-merged with no conflict**, combining the PR's second `copyDataFiles` call with main's skip-ratio guard from #557. I reproduced both defects with tests before fixing:

```
after call 1: SkippedFiles=1
after call 2: SkippedFiles=0     <-- real skipped data file erased
3-file iceberg set, 1 missing => backup failed: 1 of 3 unreadable (>10%)
```

`SkippedFiles` used `Store` not `Add`, so the Iceberg copy erased the data-file skips and the manifest claimed a complete backup — defeating #556. And the ratio was per-call, so one stale metadata entry aborted a backup whose data files had all copied.

Fixed: skips accumulate; the ratio is evaluated once via `checkSkipRatio` over every file group.

## High / Medium / Low

- **`pruneOldVersionFiles` kept exactly `retain` copies**, so a reader that resolved `version-hint.text` just before a commit could find its version already deleted — continuous at `retain_snapshots=1`. Keeps `retain+1` now, bounding the race to one reconcile interval for a few KB.
- **`retain_snapshots=0` silently meant "keep forever"** — rejected at config load, consistent with the other Iceberg guardrails.
- **A warehouse outside the storage root** disabled discovery-file publishing permanently, logged at Debug. Now warns once per table, naming what degrades (directory readers) and what does not (catalog readers).
- **A seventh SQLite handle** on the shared file, reintroducing what #329/#562 consolidated. Borrows the auth manager's handle when the catalog is in the shared database; owns one only when the operator points the key elsewhere.

## Test plan

- [x] `TestBackup_SeparateIcebergCatalogIsIncluded` / `_SharedIcebergCatalogIsNotDuplicated` — **verified the first fails pre-fix**: `manifest does not record the separate Iceberg catalog`
- [x] `TestReconcile_HintFailureIsReported` / `_ConvergedPassRepublishesHint` — **verified both fail pre-fix**: `expected hintOK=false…` and `converged reconcile did not republish version-hint.text`
- [x] `TestCheckSkipRatio_SpansAllFileGroups` — accumulation across groups, and 1-of-33 not aborting the backup
- [x] `TestScheduler_DoesNotCacheAfterHintFailure` — skips cleanly when the catalog commit fails first (a different, already-fatal path)
- [x] Existing `TestExpireSnapshotsAndPruneVersions` updated for the `retain+1` bound, plus an assertion that pruning still *bounds* growth (8 writes must not leave 8 copies)
- [x] `go test` passes for `internal/iceberg`, `internal/backup`, `internal/config`, `cmd/arc`; build / vet / gofmt clean

**Binary run** with `catalog_db_path` *and* `retain_snapshots` both non-default:

| | result |
|---|---|
| Startup | healthy; `arc.db` and `iceberg-catalog.db` created as separate files |
| Reconcile | table created + reconciled |
| Backup | contains `metadata/iceberg-catalog.db`, manifest `has_iceberg_catalog: true`, `skipped_files: 0` |
| Discovery files | `version-hint.text` published |
| Errors / shutdown | **0** errors, `Graceful shutdown complete`, no `database is closed` from the borrowed handle |

## One note on the review process

My first attempt at the hint-failure test injected failure by wrapping the storage backend. It silently did nothing: `DefaultWarehouse` **type-switches on the concrete backend type**, so an embedding wrapper falls through to the `./data` default and `warehouseRelKey` rejects every metadata path. The test passed against buggy code. I rewrote it to use a real `LocalBackend` with an unwritable metadata directory. Worth knowing before anyone else writes a test against this package — the type switch makes backend wrappers useless as test doubles.


xe-nvdk:
Thanks — addressed the `isNotFoundErr` point in `ad3124b`.

**It was a real latent bug, not just a style concern.** I traced it: `"file not found"` is produced *only* by `LocalBackend` (local.go:235, 259, 308). S3 wraps the SDK error as `"failed to read from S3: %w"` and Azure as `"failed to read from Azure Blob Storage: %w"` — neither contains that substring.

So the match is correct today purely because `backupStorage` is always a `LocalBackend` (manager.go:53 is the sole assignment). If the backup destination ever becomes object storage, the match **inverts silently**: a genuine read failure would be classified as "no catalog in this backup", and the restore would report success having quietly skipped the catalog. That is the same silent-wrong-answer shape as the original blocker.

Replaced with `backupStorage.Exists`, which is part of the `Backend` interface and returns an explicit `(bool, error)` — every backend implements it correctly, and a real failure now surfaces as an error instead of being read as absence.

Also added the restore-side tests the change was missing:
- `TestRestore_MissingIcebergCatalogIsSkipped` — a backup taken before separate-catalog support restores cleanly, catalog skipped, nothing written
- `TestRestore_SeparateIcebergCatalogIsRestored` — a backup containing one restores it to the configured path with contents intact

Full suites still green: `internal/backup`, `internal/iceberg`, `internal/config`, `cmd/arc`; build / vet / gofmt clean.

On your point 2 — agreed, and worth restating for whoever merges: the `backup.go` fix has to travel with the merge. If #533 is rebased onto `main` instead of taking this merge commit, the `Store`-vs-`Add` defect comes back with no conflict and no compile error to catch it.

PR #585, opened 2026-08-12
fix(query): preserve whitespace before bare JOIN

schotime:
### Summary

  Fixes SQL rewriting for bare JOIN clauses following a table or alias.

  The join-matching regex could consume the whitespace before a bare JOIN, producing invalid SQL such as tJOIN read_parquet(...). INNER JOIN was unaffected because
  its preceding whitespace was not part of the match.

### Changes

  - Require a join modifier before consuming modifier-adjacent whitespace.
  - Add a regression test covering a bare JOIN transformation.

### DIFF
--- a/internal/api/query.go
+++ b/internal/api/query.go
@@ -65,10 +65,10 @@ var (
 	patternSimpleTable = regexp.MustCompile(`(?i)\bFROM\s+([a-zA-Z_][a-zA-Z0-9_]*)\b`)
 	// Pattern for database.table in JOIN clauses (e.g., JOIN mydb.mytable)
 	// Includes LATERAL JOIN support: "LATERAL JOIN", "JOIN LATERAL", "CROSS JOIN LATERAL"
-	patternJoinDBTable = regexp.MustCompile(`(?i)\b(?:(?:LEFT|RIGHT|INNER|OUTER|CROSS|NATURAL)?\s*)?(?:LATERAL\s+)?JOIN\s+(?:LATERAL\s+)?([a-zA-Z0-9_]+)\.([a-zA-Z0-9_]+)\b`)
+	patternJoinDBTable = regexp.MustCompile(`(?i)\b(?:(?:LEFT|RIGHT|INNER|OUTER|CROSS|NATURAL)\s+)?(?:LATERAL\s+)?JOIN\s+(?:LATERAL\s+)?([a-zA-Z0-9_]+)\.([a-zA-Z0-9_]+)\b`)
 	// Pattern for simple table in JOIN clauses (JOIN table_name)
 	// Includes LATERAL JOIN support: "LATERAL JOIN", "JOIN LATERAL", "CROSS JOIN LATERAL"
-	patternJoinSimpleTable = regexp.MustCompile(`(?i)\b(?:(?:LEFT|RIGHT|INNER|OUTER|CROSS|NATURAL)?\s*)?(?:LATERAL\s+)?JOIN\s+(?:LATERAL\s+)?([a-zA-Z_][a-zA-Z0-9_]*)\b`)
+	patternJoinSimpleTable = regexp.MustCompile(`(?i)\b(?:(?:LEFT|RIGHT|INNER|OUTER|CROSS|NATURAL)\s+)?(?:LATERAL\s+)?JOIN\s+(?:LATERAL\s+)?([a-zA-Z_][a-zA-Z0-9_]*)\b`)
 	// Pattern to extract CTE names from WITH clauses
 	// Matches: WITH name AS, WITH RECURSIVE name AS, and comma-separated CTEs
 	patternCTENames = regexp.MustCompile(`(?i)\bWITH\s+(?:RECURSIVE\s+)?(\w+)\s+AS\s*\(|,\s*(\w+)\s+AS\s*\(`)
--- a/internal/api/query_test.go
+++ b/internal/api/query_test.go
@@ -477,6 +477,22 @@ func TestJoinClausePatterns(t *testing.T) {
 	}
 }
 
+func TestConvertSQLToStoragePaths_BareJoinPreservesAliasSeparator(t *testing.T) {
+	h := &QueryHandler{
+		storage: &mockLocalBackend{basePath: "./data"},
+		pruner:  pruning.NewPartitionPruner(zerolog.Nop()),
+		logger:  zerolog.Nop(),
+	}
+
+	result := h.convertSQLToStoragePaths("SELECT * FROM a JOIN mydb.cpu ON 1=1")
+	if strings.Contains(result, "parquet'JOIN") {
+		t.Errorf("bare JOIN consumed the separator after the preceding table. Got: %s", result)
+	}
+	if !strings.Contains(result, " JOIN read_parquet('./data/mydb/cpu/**/*.parquet'") {
+		t.Errorf("bare JOIN was not rewritten correctly. Got: %s", result)
+	}
+}
+
 func TestValidateIdentifier(t *testing.T) {
 	tests := []struct {
 		name    string


xe-nvdk:
Thanks for this — the fix is correct and I've verified it end to end. One thing needs addressing before it can merge.

## The fix is right (and the bug is worse than the description)

The old pattern made the modifier optional *and* consumed adjacent whitespace, so with no modifier present `(?:(?:LEFT|...|NATURAL)?\s*)?` reduces to "optionally eat whitespace" — the match started at the space *before* `JOIN`, and since the whole match gets replaced, the separator disappeared.

Confirmed pre-fix output on `main`:

```
SELECT * FROM a JOIN mydb.cpu ON 1=1
  => SELECT * FROM read_parquet('./data/default/aJOIN/**/*.parquet', …) read_parquet('./data/mydb/cpu/…') ON 1=1
```

So it isn't only malformed SQL: the alias `a` and `JOIN` fuse into a fabricated measurement `aJOIN`, and the join operator is dropped entirely, silently turning it into a cross join.

Your change also fixes several cases beyond the one in the description — worth knowing the blast radius was larger:

- `LATERAL JOIN`
- DuckDB's `ASOF` / `SEMI` / `ANTI` / `POSITIONAL JOIN`
- **newline-separated joins** (`FROM a\nJOIN …`), which is how most real queries are formatted

`INNER JOIN` was unaffected exactly as you said, which is why the existing tests missed it — they only covered modifier forms.

I also checked whether this reached RBAC table extraction, since `extractTableReferences` shares these same patterns. It does not: the swallowed whitespace sits outside the capture group, so the extractor recovers the same identifier either way. No security impact.

## Blocker: the merged result doesn't compile

GitHub reports this as `MERGEABLE / CLEAN` and no CI has run on the branch, which is misleading here. The branch is based on `4ec4fb0`, and `main` has since changed the signature to `convertSQLToStoragePaths(ctx, sql)`. The changed lines don't overlap textually so the merge auto-resolves — and then fails to build:

```
vet: internal/api/query_test.go:725:77: not enough arguments in call to h.convertSQLToStoragePaths
	have (string)
	want (context.Context, string)
```

Could you sync your fork with the latest `main` and rebase? The only change needed is passing a context in the new test:

```go
result := h.convertSQLToStoragePaths(context.Background(), "SELECT * FROM a JOIN mydb.cpu ON 1=1")
```

(`context` is already imported in that file.)

## Verification after applying that one-line change

- `go vet ./internal/api/` clean, full `go build ./cmd/... ./internal/...` passes
- Full `internal/api` suite passes, including `-race`
- `internal/sql` and `internal/pruning` pass
- I reverted just the two regex lines and confirmed **your new test genuinely fails pre-fix**, then restored it — thanks for writing a test that actually pins the behavior

Rebase and I'll merge once CI is green.

---

One more thing your change surfaced, which is **not** yours to fix: `buildReadParquetExpr` is called with a hardcoded `"JOIN"` literal and discards whatever modifier the regex consumed, so `LEFT JOIN mydb.cpu` currently rewrites to a plain `JOIN` — silently converting outer joins to inner ones. That's pre-existing on `main` and orthogonal to this PR; I've filed it separately. Please leave it out of scope here.

xe-nvdk:
Heads up — I've opened #587, which includes your regex fix.

While fixing the related modifier-dropping bug (#586) in the same two patterns, it made sense to land both changes together rather than have them collide: #586's fix requires *capturing* the join prefix, which touches the exact characters your change corrects. Your fix is in there, extended to capture the prefix rather than just require the trailing whitespace.

Full credit to you in the release notes and PR description — the whitespace defect was your find, and the `aJOIN` fabricated-measurement case is called out explicitly.

Your analysis also turned out to be the thread that unravelled a third, unrelated defect: chasing an edge case in these patterns surfaced a stale-offset bug in the table-reference lookahead that made rewriting depend on table-name length (`JOIN cross` failed, `JOIN crossx` worked) and caused `JOIN generate_series(1, 10)` to be rewritten as if it were a measurement. That's fixed in #587 too.

I'll close this in favor of #587 once that merges — but if you'd rather land yours first, say so and I'll rebase mine on top instead. Thanks for the careful report and for writing a test that actually pins the behavior.

xe-nvdk:
@schotime — your fix shipped in #587. Quoting your original diagnosis, because it was exactly right:

> The join-matching regex could consume the whitespace before a bare JOIN, producing invalid SQL such as tJOIN read_parquet(...). INNER JOIN was unaffected because its preceding whitespace was not part of the match.

That is the whole bug. The modifier being optional *and* its trailing whitespace being optional meant the prefix collapsed to "optionally consume whitespace" for an unmodified `JOIN`, so the match began at the space before it — and replacing the match deleted the separator. Your one-line change, requiring whitespace after a modifier rather than letting it float, is the basis of what merged:

```go
// yours
(?:(?:LEFT|RIGHT|INNER|OUTER|CROSS|NATURAL)\s+)?(?:LATERAL\s+)?JOIN\s+...

// shipped
((?:(?:LEFT|RIGHT|FULL|INNER|OUTER|CROSS|NATURAL|SEMI|ANTI|ASOF|POSITIONAL)\s+)*(?:LATERAL\s+)?JOIN\s+(?:LATERAL\s+)?)...
```

The additions on top are a wrapping capture group (so the operator can be re-emitted rather than replaced with a hardcoded `"JOIN"` — that's #586), `?` → `*` for stacked modifiers like `RIGHT ANTI JOIN`, and five more DuckDB join keywords. The part that fixes *your* bug is unchanged from what you wrote.

I ran your two assertions verbatim against the merged code — the `parquet'JOIN` check and the ` JOIN read_parquet('./data/mydb/cpu/**/*.parquet'` check — and both pass. They're preserved in the shipped tests, folded into a table-driven `TestJoinModifierPreserved` with the context argument added.

You're credited by name in the release notes for 26.09.1, as finding *and* fixing it — not just reporting it.

Two things your report led to that were outside its scope:

1. **#586** — the same patterns discarded the join modifier entirely, replacing it with a hardcoded `"JOIN"`. `LEFT JOIN` silently became an inner join, dropping unmatched rows with no error. `NATURAL JOIN` failed at the DuckDB parser instead, since stripping `NATURAL` left no join condition.
2. **A stale-offset bug in the table-reference lookahead** — it recovered match positions from a copy of the SQL captured *before* earlier rewrite passes ran, so correctness depended on table-name length (`JOIN cross` was left unresolved while `JOIN crossx` worked), and `JOIN generate_series(1, 10)` was rewritten as if it were a measurement.

Neither was visible from your PR; both surfaced from pulling on the thread you started.

Closing this as superseded by #587 — but that's only because the two fixes touch the same characters and would have conflicted, not because anything was wrong with yours. Thanks for the careful diagnosis and for writing a test that actually pins the behavior; that's what made it quick to confirm.

xe-nvdk:
@schotime — to be unambiguous about attribution, since this PR will end up closed rather than merged and I don't want the button that gets pressed to be mistaken for the credit you're owed:

**This fix is yours.** You found the bug, diagnosed the root cause correctly, and wrote the change that shipped. #587 carries it forward — it does not replace it. The only reason your PR isn't the one merging is that #586 and a third defect touch the same characters in the same two regexes, so landing them separately would have conflicted. That's a mechanical consequence of overlapping lines, nothing more.

You're credited by name in three places, all pushed:

- **README → Contributors** — your line now reads "Data-time partitioning, compaction API triggers, UTC fixes, **bare-JOIN SQL rewriting fix**"
- **Release notes for 26.09.1** — "found **and fixed** by @schotime", describing your actual diagnosis rather than filing you under "reported by"
- **#587's description** — your PR is linked, and the folded-in change is attributed to you inline

For the record, the part of the shipped regex that fixes your bug is byte-for-byte what you wrote — `(?:MODIFIER\s+)` instead of `(?:MODIFIER)?\s*`. Everything layered on top (the capture group, `?` → `*`, the extra DuckDB keywords) exists to serve #586 and does not alter your fix. I also ran your two assertions verbatim against the merged code and both pass.

Genuinely useful report: you caught something that had been live for a long while and was invisible precisely because `INNER JOIN` round-tripped cleanly. Pulling on that thread turned up two further defects neither of us was looking for. Please do send more.

Closing this once #587 merges — as superseded, not rejected.

xe-nvdk:
Closing now as superseded by #587, which carries your fix forward — not rejected, and not a judgement on the work.

(My previous note said I'd close this once #587 merged; closing a touch earlier so the queue reflects reality. The attribution above stands regardless of what happens to #587: README, release notes, and #587's description all credit you by name, and those are already pushed.)

schotime:
No problems at all mate.
Just happy we found a bug and have a fix ;)

PR #622, opened 2026-08-20
fix: Reject concurrent backup operation requests

### LINKED ISSUE #299: high(auth): Backup race condition on operation-in-progress check
backup_routes.go:57-58 — TOCTOU between check and goroutine. Fix: use TryLock.

mvanhorn:
The backup API checks `Manager.GetProgress` before launching work in a goroutine, so concurrent POST requests can both observe an idle manager and both receive `202 Accepted`. The manager mutex added since the issue was opened prevents simultaneous execution, but it only serializes the accepted operations; the second request silently queues instead of receiving `409 Conflict`. The same check-then-spawn pattern exists in the restore endpoint and shares the manager's operation boundary. The remaining defect is therefore at the HTTP admission boundary, not in backup data processing.

## Test plan

- Send multiple concurrent valid backup POSTs while the accepted backup is blocked in storage discovery; assert exactly one response is `202 Accepted` and every other response is `409 Conflict`, with no second backup queued.
- While a backup request owns the admission guard, send a valid confirmed restore request and assert it receives `409 Conflict`, proving backup and restore share one operation slot.
- Let the accepted backup finish, then send another valid backup request and assert it receives `202 Accepted`, proving the guard is released on completion rather than leaving the API permanently busy.
- Force the admitted background backup to fail and then submit another request; assert the subsequent request is accepted, covering guard release on the error path.

## Summary

Add a single atomic operation-admission guard to `BackupHandler`, shared by backup and restore requests, and acquire it synchronously before either handler launches its goroutine. Treat failure to acquire as the existing conflict response, and release the guard from the operation goroutine on every completion, failure, or timeout path while retaining the manager mutex as the lower-level serialization safeguard. Keep request parsing and restore validation ahead of admission so invalid requests do not occupy the guard, and preserve the existing detached contexts, response bodies, and progress polling behavior.

Closes #299

### DIFF
--- a/internal/api/backup_routes.go
+++ b/internal/api/backup_routes.go
@@ -3,6 +3,7 @@ package api
 import (
 	"context"
 	"regexp"
+	"sync/atomic"
 	"time"
 
 	"github.com/basekick-labs/arc/internal/auth"
@@ -16,9 +17,10 @@ var validBackupID = regexp.MustCompile(`^backup-\d{8}-\d{6}-[a-f0-9]{8}$`)
 
 // BackupHandler handles backup and restore API operations.
 type BackupHandler struct {
-	manager     *backup.Manager
-	authManager *auth.AuthManager
-	logger      zerolog.Logger
+	manager         *backup.Manager
+	authManager     *auth.AuthManager
+	logger          zerolog.Logger
+	activeOperation atomic.Pointer[string]
 }
 
 // NewBackupHandler creates a new backup handler.
@@ -54,15 +56,6 @@ type CreateBackupRequest struct {
 // CreateBackup triggers a new backup.
 // POST /api/v1/backup
 func (h *BackupHandler) CreateBackup(c *fiber.Ctx) error {
-	// Check if a backup is already running
-	if p := h.manager.GetProgress(); p != nil && p.Status == "running" {
-		return c.Status(fiber.StatusConflict).JSON(fiber.Map{
-			"error":     "A backup or restore operation is already in progress",
-			"status":    p.Status,
-			"operation": p.Operation,
-		})
-	}
-
 	var req CreateBackupRequest
 	if err := c.BodyParser(&req); err != nil {
 		// Empty body is fine — use defaults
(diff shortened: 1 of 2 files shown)


xe-nvdk:
Hey @mvanhorn thank you for this PR. I will review this properly in the next couple of hours and provide feedback. Thank you so much for taking the time to contribute to Arc.

xe-nvdk:
Thanks @mvanhorn — this is a well-built fix. We ran it through our internal review process (configuration matrix + deep adversarial review + live verification) and it comes out clean. Summary of what we verified:

**Correctness**
- The ordering proof holds: inside the manager, defer LIFO guarantees *progress finalized → `m.mu` released → admission slot released* (the goroutine's `Store(nil)` defer is registered first, so it runs last). We traced every error return in `CreateBackup`/`RestoreBackup` — all set a terminal status before returning. There is no reachable state where two operations are both admitted, none where the API goes permanently busy, and none where an admitted request queues behind a finishing predecessor (the #299 complaint).
- No guard leak: nothing sits between `acquireOperation` and the goroutine spawn, and the release defer covers every path.
- The handler is the only caller of `CreateBackup`/`RestoreBackup` in the tree, so the guard's coverage is complete.

**Tests**
- Verified the concurrency test genuinely fails on `main`'s handler (`accepted=11 conflicts=5` out of 16) — it's a real regression pin, and it's deterministic rather than timing-dependent (the admitted backup provably holds the slot for the whole admission phase via the blocked `ListObjects`).
- Pass under `-race -count=2`, both at the PR head and after a test-merge with current `main` (merges clean). `gofmt`/`go vet` clean.

**Two non-blocking observations**
1. After a failed CAS, `activeOperation.Load()` can observe `nil` if the holder released in between, producing `"operation": ""` in the 409 body. Nanosecond window, self-correcting — fine to ship as-is.
2. `DeleteBackup` doesn't share the admission slot, so deleting a backup mid-restore can race. That's pre-existing on `main`, not introduced here — tracked separately in a follow-up issue.

Approving CI now; we'll add the 26.09.1 release-notes line on our side. Thanks again for the contribution!

Issue #639, opened 2026-08-20
low(iceberg): audit sweep — interval validation, silent cluster no-run, dropped-DB residue, per-pass hygiene


xe-nvdk:
Low-severity items from the 2026-08-20 adversarial audit of the Iceberg export subsystem, batched. None is urgent; each is cheap.

1. **`iceberg.reconcile_interval` ≤ 0 silently clamps to 5m** (`internal/iceberg/scheduler.go:77-79`), diverging from the documented 300s default and from `retain_snapshots`' explicit validation (`config.go:1232-1236`). An operator setting `0` hoping to disable gets a running reconciler. Validate at load like retain_snapshots.
2. **Cluster: no compactor-role node → Iceberg export silently never runs.** The reconciler shares the compaction cluster gate (`cmd/arc/main.go:3144-3147`); if no node holds the role, every node's gate is false with no log saying why. Log once at startup when iceberg is enabled but the gate can never pass.
3. **Dropped database leaves permanent catalog residue.** `handleDelete` deletes files then non-recursive best-effort `RemoveDirectory` (`internal/api/databases.go:383`, `local.go:596` fails on non-empty subtree). Surviving dirs → tables correctly emptied; but namespaces/tables/warehouse metadata persist forever, and an out-of-band `rm -rf` of the tree leaves tables referencing deleted files with no reconcile path ever visiting them again. Add catalog cleanup on database drop.
4. **One unreadable database dir aborts the whole reconcile pass** (`source.go:78-81` propagates; `scheduler.go` runPass gives up). Skip-and-log per database instead.
5. **`warehouseRelKey` compares URI strings, not resolved paths** (`exporter.go:673-687`): a symlinked/unclean rendering of the storage root silently lands in warn-once "hint unaddressable" mode despite being physically under the root. Resolve like `SharesDBPath` does.
6. **`-wal`/`-shm` sidecars not hardened to 0600** by `sharedSQLiteHandle` (`cmd/arc/main.go:3684-3713`) the way auth does (`auth.go:231-260`) — applies to a dedicated `iceberg.catalog_db_path` and to the shared DB when auth is disabled.
7. **iceberg-go writes warehouse metadata files at umask perms** (`io/local.go:34-45`) vs Arc's 0600 backend writes — metadata-only exposure (schema/paths), consistency nit.
[...]

Issue #632, opened 2026-08-20
high(iceberg): ExpireSnapshots misreads successful expiry as failure; PostCommit hook holds delete authority over Arc's data files


xe-nvdk:
Found during the 2026-08-20 adversarial audit of the Iceberg export subsystem. Independently live-repro'd twice against iceberg-go v0.6.0 (sqlite catalog, `WithRetainLast(2)+WithOlderThan(0)`, superseded files removed from disk Arc-style — the production order).

## Problem

`Transaction.Commit` for a snapshot-expiry runs orphan deletion as a **post-commit hook**: the catalog commit lands first, then iceberg-go `os.Remove`s the expired snapshots' manifest-lists, manifests, and **data files no longer referenced** (`table/updates.go:471-596`, `io/local.go:48-50`), joining every failure into `Commit`'s returned error with no `IsNotExist` tolerance.

In Arc, files leave the table precisely *because Arc already deleted them* (compaction, retention, delete API). So on the **default config** (`retain_snapshots=10`, compaction on), once history passes `retain`, every expiring pass produces ENOENT errors from a commit that actually succeeded. Repro output: error list **grows each pass** (expired manifests carry forward entries for every historically removed file), snapshot count correctly capped — proving the commit landed despite the reported failure.

Consequences at `internal/iceberg/exporter.go:498-507` → `416-418`:
1. Logs ERROR "ExpireSnapshots commit failed — snapshot history grows until it recovers" — wrong on both counts, fires forever after any churn, masks genuine expiry failures.
2. Returns the **pre-expire** table, so `writeVersionHint` publishes metadata still listing expired snapshots whose manifest-list files the hook successfully deleted → directory readers doing snapshot listing/time-travel hit missing files.
3. The hint write succeeds → scheduler caches the fingerprint (`scheduler.go:281-293`); a measurement that goes quiet right after keeps the stale, partially-dangling hint **indefinitely**.

## Tail risk

`postCommit=true` means iceberg-go holds physical delete authority over Arc's primary data files: any future listing bug that transiently omits live files for `retain` generations converts into silent deletion of primary data by the exporter. Arc owns data-file lifecycle; the exporter should never delete data files.

## Fix shape

[...]

PR #663, opened 2026-08-29
fix(iceberg): publish version hint after metadata copy

### LINKED ISSUE #636: medium(iceberg): version-hint.text is updated even when the v<N>.metadata.json copy failed — breaks working directory readers

## Problem

`writeVersionHint` (`internal/iceberg/exporter.go:611-631`): when the metadata read (line ~614) or the `v<N>.metadata.json` copy (line ~619) fails, `okAll` is set false but execution **falls through and writes `version-hint.text` anyway** (lines ~627-631). The hint then says `N` while `vN.metadata.json` doesn't exist — the exact "metadata file for version N missing" failure the function's own doc-comment says was empirically verified to break both DuckDB and Spark directory readers.

## Failure scenario

Before the pass, readers had a stale-but-working `N-1` hint/metadata pair. One transient failure — ENOSPC is the realistic one: the tiny hint write succeeds where the metadata-sized write fails — flips them to a broken pair for a full reconcile interval. The retry machinery (`hintOK=false` → fingerprint uncached → next-pass retry) bounds the damage but doesn't prevent it.

## Fix shape

Skip the hint update when the metadata copy failed — the publishing order exists precisely so the pair stays consistent. Keep `okAll=false` so the pass retries.

bferanmi806-sketch:
Fixes #636

`writeVersionHint` could advance `version-hint.text` even when reading or publishing the corresponding `v<N>.metadata.json` copy failed. That could leave directory-based Iceberg readers pointing at a metadata version that was not available.

Only advance the hint after the matching metadata copy succeeds, preserving the last known-good hint while the existing reconciliation retry path retries publication.

Adds a regression that keeps the hint writable while making the newer metadata unavailable, verifies the old hint is preserved, then verifies recovery publishes and advances to the new version.

Also updates the 2026.09.1 release notes.

### DIFF
--- a/RELEASE_NOTES_2026.09.1.md
+++ b/RELEASE_NOTES_2026.09.1.md
@@ -522,6 +522,16 @@ the binary and dependency scanners come up clean.
 
 ## Bug fixes
 
+### Iceberg version hints no longer advance before metadata copies publish ([#636](https://github.com/Basekick-Labs/arc/issues/636))
+
+Directory-based Iceberg readers fetch `version-hint.text` before opening the matching
+`v<N>.metadata.json` copy. A transient metadata read or copy failure could previously
+still advance the hint, leaving those readers pointed at an unavailable snapshot.
+
+Arc now publishes the hint only after the matching metadata copy succeeds. The previous
+hint remains valid during the failure, and the existing reconciliation retry path
+self-heals once storage recovers.
+
 ### Concurrent backup/restore requests are now rejected with 409 instead of silently queuing ([#299](https://github.com/Basekick-Labs/arc/issues/299))
 
 The backup API checked "is an operation running?" before launching the work in
--- a/RELEASE_NOTES_2026.09.1.md
+++ b/RELEASE_NOTES_2026.09.1.md
@@ -522,16 +522,6 @@ the binary and dependency scanners come up clean.
 
 ## Bug fixes
 
-### Iceberg version hints no longer advance before metadata copies publish ([#636](https://github.com/Basekick-Labs/arc/issues/636))
-
-Directory-based Iceberg readers fetch `version-hint.text` before opening the matching
-`v<N>.metadata.json` copy. A transient metadata read or copy failure could previously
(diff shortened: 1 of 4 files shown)


xe-nvdk (commented):
Thanks for this — it's a well-built fix, and welcome as a first-time contributor! 🎉

**Verified locally** (maintainer review):
- The fix matches #636's prescribed shape exactly: the hint now publishes only after the matching `v<N>.metadata.json` copy succeeds, `okAll=false` keeps the retry path armed, and the deferred `noteHintFailure` still fires on the early return.
- Bonus correctness: treating `warehouseRelKey` failure as a failure (instead of silently skipping the copy but still writing the hint) closes the same inconsistency from another angle. The permanent outside-warehouse configuration is unaffected — it's handled earlier by `parseVersionAndMetaDir` → warn-once → return true.
- **Regression test verified failing pre-fix** (reverted `exporter.go` to main, test fails with `hint changed after metadata failure: got "2", want "1"`; restored, passes). The recovery leg is a nice touch. Full `internal/iceberg` suite, `go vet`, and `gofmt` all clean.
- CI workflow run approved (first-time contributor gate).

**One change requested:** the release-notes entry needs to move from `RELEASE_NOTES_2026.09.1.md` to `RELEASE_NOTES_2026.09.2.md`. 26.09.1 is frozen and releases Monday from an already-cut branch, so this merge lands in the 26.09.2 patch (October) — where the rest of the iceberg audit issues (#632–#639) are targeted. The 26.09.2 file already exists on main; the same text works as-is under a `## Bug fixes` section there.

After that rebase/edit this looks ready to me — final approval/merge to the maintainer.

bferanmi806-sketch:
Thanks @xe-nvdk, really appreciate the detailed review and the welcome! I’ll move the release note to 26.09.2 and update the branch.

I enjoyed working on this. Once this is sorted, I’d be happy to pick up another issue from the Iceberg audit if there’s one you think would be useful to tackle next.

PR #664, opened 2026-08-29
fix(iceberg): validate reconcile interval at config load

bferanmi806-sketch:
Item 1 of #639.

- iceberg.reconcile_interval <= 0 is now rejected during config load when Iceberg is enabled instead of silently becoming the scheduler's 5-minute fallback.
- The normal 300-second default and scheduler defensive fallback remain unchanged.
- Regression coverage includes the default, valid positive values, zero, and negative values.
- Release notes target 26.09.2.
- Checks passed: go test ./internal/config, focused interval regression test, go vet ./internal/config, gofmt, and git diff --check.


xe-nvdk:
Reviewed and verified locally on the PR branch:

- `go test ./internal/config` passes, including the new `TestLoad_IcebergReconcileInterval` subtests (default, 1, 300, 3600, zero, negative).
- `go vet`, `gofmt -l`, and `git diff --check` are all clean.
- The validation sits inside the `cfg.Iceberg.Enabled` block next to the `retain_snapshots` check and follows the same error message style.
- The scheduler fallback in `internal/iceberg/scheduler.go` is untouched, as described, and still makes sense as a defensive default for programmatic `SchedulerConfig` construction.
- The "default" test case works correctly: viper treats an empty env var as unset, so it exercises the real 300s default.
- Rejecting 0 is the right call. The docs never documented 0 as a disable switch, and `iceberg.enabled=false` already covers that.

Matches item 1 of #639. LGTM.

PR #665, opened 2026-08-29
fix(iceberg): skip unreadable databases during reconciliation

bferanmi806-sketch:
Item 4 of #639.

StorageWalkSource now logs and skips failures listing measurements inside one database so other databases continue reconciling. Failures enumerating the top-level database list remain fatal for that pass.

Regression coverage uses a deterministic backend error injection and verifies both the skipped database log and top-level error behavior.

Checks: focused Measurements tests, go vet ./internal/iceberg, gofmt, and git diff --check passed. The full Iceberg package test was attempted but is environment-limited here by CGO_ENABLED=0/no C compiler for go-sqlite3 plus an existing Windows POSIX file-URL assertion.


xe-nvdk:
Reviewed and verified locally on the PR branch:

- Full `go test ./internal/iceberg` passes here with CGO available, covering the part your environment could not run. The two new tests pass, and so does the existing warehouse-dir regression.
- `go vet ./internal/iceberg` and `git diff --check` are clean. `gofmt -l` flags `scheduler_test.go`, but that already exists on main and this PR leaves the file untouched, so it is unrelated.
- Safety of the skip: a database that fails to list only drops out of the `seen` map, which evicts its fingerprint cache entries. The next successful pass re-fingerprints and re-reconciles it, so the cost of a transient failure is one extra full reconcile of that database. The reconcile-to-empty path only runs for measurements the walk actually returns, so a listing failure can never empty or delete an Iceberg table.
- Production wiring is correct: `runPass` is the only production caller of `Measurements` and it injects the logger via `WithContext`, so the skip log reaches the scheduler's component logger.

One observation, no change needed: `zerolog.Ctx` context-carried logging is new to this codebase, which elsewhere puts loggers in struct fields. A logger field on `StorageWalkSource` would also survive any future caller that forgets `WithContext`. Fine to keep as is for a single-caller type.

Matches item 4 of #639. LGTM.

xe-nvdk:
One change requested after all: please move the logging onto the struct so it matches how the rest of the codebase handles loggers.

Concretely:

- Add a `logger zerolog.Logger` field to `StorageWalkSource` and accept it in `NewStorageWalkSource`, the same way `NewLocalBackend` and the schedulers take theirs.
- Log the skip with `s.logger` in `Measurements`.
- Revert the `s.logger.WithContext(ctx)` change in `scheduler.go` and pass the logger at the construction site in `cmd/arc/main.go` instead.
- The tests can pass `zerolog.New(&logs)` directly to the constructor, which also simplifies them a bit.

This keeps a single logging pattern across the codebase. Everything else stands as reviewed.

PR #666, opened 2026-08-29
fix(sqlite): harden dedicated WAL and SHM permissions

bferanmi806-sketch:
Item 6 of #639.

The dedicated sharedSQLiteHandle path now tightens the SQLite database and its -wal/-shm sidecars to 0600, resolves symlinked database paths before hardening sidecars, tolerates missing sidecars, and leaves auth-owned handles untouched.

Regression coverage covers a fresh dedicated database and permissive sidecars behind a symlink.

Checks: gofmt and git diff --check passed. cmd/arc tests and vet were attempted but this Windows checkout cannot compile the existing DuckDB bindings (no Windows binding files).


xe-nvdk:
Reviewed and verified locally on the PR branch, on Linux with working DuckDB bindings:

- Full `go test ./cmd/arc` passes, covering what your Windows checkout could not compile. Both new tests pass, including the symlink one, and the three pre-existing `sharedSQLiteHandle` tests still pass.
- `go vet ./cmd/arc`, `gofmt -l`, and `git diff --check` are clean.
- The approach is sound. Pre-creating the sidecars 0600 before the first write is stronger than the auth.go pattern and fits this path, since the dedicated handle's callers run their schema writes later, after this function returns. The symlink resolution mirrors auth.go and runs after the DB file is pre-created, so `EvalSymlinks` cannot fail on a fresh install.
- The post-open `os.Chmod(dbPath, 0600)` also quietly fixes a pre-existing gap: a database file created 0644 by an older version stayed 0644 on this path, because the `O_CREATE` mode only applies to newly created files.
- Auth-owned handles return before any of this code, as described.

One change requested: this repo already has `cmd/arc/shared_sqlite_test.go` holding the `sharedSQLiteHandle` tests. Please move the two new tests and the helper into that file and drop `main_test.go`, so the function's tests stay in one place.

Optional simplification, your call: the post-open sidecar chmod loop repeats what the pre-create loop already guarantees, since that loop chmods unconditionally and covers pre-existing permissive sidecars. It can be dropped. The post-open chmod of the main DB file does real work and should stay.

Matches item 6 of #639.

PR #667, opened 2026-08-29
perf(iceberg): cache empty measurement negatives

bferanmi806-sketch:
Item 9 of #639.

Empty measurement directories with no Iceberg table now cache a confirmed missing-table result, avoiding repeated catalog lookups on unchanged passes. The cache does not create zero-column tables, is invalidated by re-ingest, preserves existing-table empty reconciliation, and is removed when a measurement disappears.

Other catalog errors are not cached and remain eligible for retry.

Checks: go vet ./internal/iceberg, gofmt, and git diff --check passed. Focused scheduler tests were attempted but this Windows checkout uses CGO_ENABLED=0, so go-sqlite3 is a stub.


xe-nvdk:
Reviewed and verified locally on the PR branch with CGO available:

- Full `go test ./internal/iceberg` passes, covering the scheduler tests your checkout could not run. Both new tests pass.
- `go vet`, `git diff --check`, and gofmt on the new code are clean. The one `gofmt -l` complaint in `scheduler_test.go` is the pre-existing alignment quirk on main, untouched by this PR.
- The passing `TestScheduler_EmptyDirNeverCreatesTable` also proves the important library assumption: iceberg-go's SQL catalog wraps `ErrNoSuchTable` in a way `errors.Is` matches. If it did not, the reconciler would error on every empty directory and the test would fail.
- Splitting `tableExists` into a `(bool, error)` form was the right move. Caching on top of the old error-swallowing bool would have cached a transient catalog outage as a permanent missing table.
- The cached negative short-circuits at the existing fingerprint check before any catalog access, so dead measurements now cost zero shared-DB queries per tick. Re-ingest invalidation, the existing-table empty path, and eviction on disappearance are all covered by the new test additions.

One change requested: the doc comment on the exported `TableExists` is now stale. It still says the reconciler uses it and explains the error-reads-as-false rationale, but the reconciler now calls `tableExists` and the exported form has only test callers left. Please move that explanation onto `tableExists` and reduce the `TableExists` comment to a note that it is the error-swallowing convenience form.

Matches item 9 of #639.

bferanmi806-sketch:
@xe-nvdk fixed this one too. I’ve now addressed the requested changes across #665, #666, and #667, so all the Arc PRs from the #639 batch should be updated on my side.

xe-nvdk:
Heads up on the two maintainer commits I pushed to this branch, since the second one is a nice case study.

To get this mergeable I merged current `main` into the branch and resolved the `RELEASE_NOTES_2026.09.2.md` conflict, same as on your other PRs (your entry now sits last under the shared `## Bug fixes` section, in merge order).

The interesting part: after that merge, the branch no longer compiled, even though git reported only the release-notes conflict. Your #665 added a `zerolog.Logger` parameter to `NewStorageWalkSource`, and this PR's new test at `scheduler_test.go` still called the two-argument form. Git merges text, not meaning: the two changes touched different lines, so the merge was "clean" while the combination was broken. This is called a semantic conflict, and it is why running the tests after every merge or rebase matters even when git shows no conflicts at all (`93fdc67` is the one-line fix, passing `zerolog.Nop()` like the sibling call sites).

A related habit worth stealing: my first run of the suite after the fix piped through `tail`, which made the shell report the pipeline as successful while the tests had actually failed. Exit codes come from the last command in a pipe, so either check the test output itself or use `set -o pipefail` when chaining.

Will merge once CI is green. Thanks for turning all four of these around so quickly.

PR #672, opened 2026-09-01
docs: add pre-26.06.1 token upgrade runbook

### LINKED ISSUE #457: Document: pre-26.06.1 token re-issue path for cluster upgrades
Phase A ([#451](https://github.com/Basekick-Labs/arc/pull/451)) introduced cluster-wide token replication. **Tokens created on a pre-26.06.1 node remain valid only on that node after upgrade** — they don't auto-migrate into the Raft FSM.

This is intentional (auto-migration adds a one-shot code path with its own failure modes — see Phase A planning doc) but it does mean operators upgrading a multi-node Arc Enterprise cluster need a clean re-issue workflow.

## Current state (as of v26.06.1 release)

- Release notes section "No migration of pre-existing tokens" calls out the policy.
- docs.basekick.net `configuration/authentication.md` covers the divergence-detection error log + the SQL cleanup for ID collisions.
- Both compose and Helm chart documentation point at the leader-only-banner behaviour.

## What's NOT yet documented as a single runbook

A step-by-step upgrade procedure:

1. Pre-upgrade: enumerate active tokens across all nodes (operator may need to grep each node's `auth.db` separately because they don't share state today).
2. Roll the cluster forward to v26.06.1.
3. Confirm `arc_cluster_auth_apply_create_total` counter is incrementing in sync on every node (proves the FSM replication path is alive).
4. Re-issue each previously-active token via `POST /api/v1/auth/tokens` from any node (the new tokens are cluster-wide).
5. Distribute the new token values to downstream consumers (CI secrets, SDKs, dashboards).
6. Revoke the old per-node tokens via `POST /api/v1/auth/tokens/:id/revoke` (the revoke also replicates cluster-wide).
7. Optional: drop the diverging pre-26.06.1 AUTOINCREMENT rows from local `auth.db` files to clean up the divergence-detection error log noise.

## Where to put this

Likely a new "Upgrading to v26.06.1" page under docs.basekick.net `operations/` or `installation/upgrades/`. Could also be a release-notes addendum once we have customer feedback on what they actually do.

## Out of scope

Auto-migration tooling — explicitly deferred. The risk/reward isn't worth it for what's likely a once-per-cluster operation.

Related: [Phase A memory note](memory/project_cluster_auth_phase_a.md).

mah1104ahm:
## Summary

- add a staged upgrade runbook for clusters with pre-26.06.1 API tokens
- cover inventory, backup, rolling upgrade, token re-issue, consumer rotation, and revocation
- include divergence recovery and explicit completion criteria

## Verification

- reviewed the documented endpoints and upgrade sequence against the current token routes
- `git diff --check`

Closes #457

### DIFF
--- a/RELEASE_NOTES_2026.06.1.md
+++ b/RELEASE_NOTES_2026.06.1.md
@@ -619,6 +619,21 @@ Before 26.06.1, every Arc Enterprise cluster node carried its **own** SQLite aut
 
 **No migration of pre-existing tokens.** Tokens created on a pre-26.06.1 node by the local-only API path remain valid **only on that node** after upgrade. Operators are expected to re-issue API tokens through the new replicated path so they take effect cluster-wide. Bootstrap tokens set via `ARC_AUTH_BOOTSTRAP_TOKEN` are unaffected if the same value was used on every node (which the previous workaround required).
 
+#### Upgrade runbook for pre-26.06.1 API tokens
+
+Use this procedure during the maintenance window for every Arc Enterprise cluster that has API-created tokens from before 26.06.1:
+
+1. **Inventory before the upgrade.** On every node, record each active token's name, permissions, owner, and downstream consumers. The local databases can differ, so do not treat any one node as authoritative. Do not copy bcrypt hashes or plaintext token values into the inventory.
+2. **Back up each local auth database.** Stop Arc on the node before copying the file configured by `auth.db_path`; include its SQLite `-wal` and `-shm` sidecars when present.
+3. **Upgrade every node to 26.06.1 or later** and wait for the Raft cluster to regain a stable leader and full membership.
+4. **Verify the replicated write path.** Create one short-lived validation token through `POST /api/v1/auth/tokens`, then confirm that `arc_cluster_auth_apply_create_total` advances by the same amount on every node and that the token authenticates against every node. Revoke the validation token when the check passes.
+5. **Re-issue each inventoried token.** Create a replacement through `POST /api/v1/auth/tokens` on any node, using the original scope and permissions. Capture the returned plaintext once; Arc cannot display it again.
+6. **Rotate downstream consumers.** Update CI secrets, SDK configuration, dashboards, and other clients to the replacement value. Verify each consumer against more than one cluster node before continuing.
+7. **Revoke the old per-node token** through `POST /api/v1/auth/tokens/:id/revoke`, then verify that it fails on every node. Repeat steps 5-7 one token at a time to keep rollback scope small.
+8. **Resolve divergence alerts before rejoining a node.** If `arc_cluster_auth_rejected_total` grows or the logs report an ID collision, stop the affected node, preserve its auth database backup, and remove only the conflicting legacy rows (or rebuild the local auth database from the FSM) before rejoining. Never delete an auth database while Arc is running.
+
+The migration is complete when every active consumer uses a replacement token, the apply counters converge across nodes, the rejected counter is stable, and every legacy token fails against every node.
+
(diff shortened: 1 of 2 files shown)


RELEASE_NOTES_2026.06.1.md
```diff
@@ -619,6 +619,21 @@ Before 26.06.1, every Arc Enterprise cluster node carried its **own** SQLite aut
...
+3. **Upgrade every node to 26.06.1 or later** and wait for the Raft cluster to regain a stable leader and full membership.
+4. **Verify the replicated write path.** Create one short-lived validation token through `POST /api/v1/auth/tokens`, then confirm that `arc_cluster_auth_apply_create_total` advances by the same amount on every node and that the token authenticates against every node. Revoke the validation token when the check passes.
+5. **Re-issue each inventoried token.** Create a replacement through `POST /api/v1/auth/tokens` on any node, using the original scope and permissions. Capture the returned plaintext once; Arc cannot display it again.
+6. **Rotate downstream consumers.** Update CI secrets, SDK configuration, dashboards, and other clients to the replacement value. Verify each consumer against more than one cluster node before continuing.
+7. **Revoke the old per-node token** through `POST /api/v1/auth/tokens/:id/revoke`, then verify that it fails on every node. Repeat steps 5-7 one token at a time to keep rollback scope small.
```
xe-nvdk: In cluster mode this endpoint proposes the revoke through Raft (`internal/auth/auth.go`, `RevokeToken`), and the FSM apply looks the ID up in replicated state (`internal/cluster/raft/fsm.go`, `applyRevokeToken`). Pre-26.06.1 tokens were never migrated into the FSM, which is the premise of this runbook, so this call has two failure modes for a legacy token:

- If no replicated token holds that ID, the apply is treated as idempotent log replay and no-ops without touching local SQLite. The API still returns success, and the legacy token keeps authenticating on its home node, so the "verify that it fails on every node" check can never pass.
- If a replicated token does exist at that ID (legacy `AUTOINCREMENT` IDs can collide with Raft-log-index IDs), the revoke disables the new replacement token cluster-wide, right after step 6 pointed CI and dashboards at it.

The reliable removal path for legacy tokens is the local cleanup the surrounding section already describes: stop the node that holds the legacy rows, delete them from the local auth DB (or rebuild it from the FSM), and restart. Suggest rewording step 7 around that, and noting that the API revoke applies only to tokens created after the upgrade, such as the step 4 validation token. The completion criterion in the closing paragraph depends on this fix too, since the API revoke as written can never make a legacy token fail on its home node.

To be clear, #457 itself says "the revoke also replicates cluster-wide", so this runbook transcribed the issue faithfully; the issue text needs the same correction.

xe-nvdk (changes requested):
Thanks for putting this together, and for tracking the shape of #457 so closely. I verified the endpoints, the metric names, `auth.db_path`, and the "on any node" claim (followers forward proposals to the leader) against the code, and they all check out. Two things before this ships:

1. Step 7 needs a different mechanism for the legacy tokens. Details in the inline comment.
2. Placement: this lands at the bottom of the 26.06.1 release notes, and operators upgrading months from now are unlikely to reopen those. #457 suggested a dedicated upgrade page under `operations/` or `installation/upgrades/` on docs.basekick.net, and I think that is the right home. I'd keep a short pointer in the release notes and move the runbook itself there so it stays discoverable.

mah1104ahm:
Thank you for the detailed review. I corrected the legacy-token
removal procedure and pushed bb31fa2 to this PR.

The runbook now explicitly forbids using the replicated revoke
endpoint for unmigrated legacy tokens, explains both the no-op and
ID-collision failure modes, and requires stopped-node cleanup with
backups and positive legacy-row identification. It now lives at
operations/pre-26.06.1-token-upgrade.md, with a pointer from the
release notes. I checked the revoke path against AuthManager and the
FSM implementation and verified the relative links. Publishing the
page on docs.basekick.net still needs your documentation-site
integration.

Thank you for catching the issue's misleading revoke instruction as well.

M Ahmed

xe-nvdk (approved):
Re-reviewed bb31fa2. The rework addresses the revoke concern fully: the runbook now forbids the replicated revoke endpoint for unmigrated legacy rows, documents both the silent no-op and the ID-collision failure mode, and moves cleanup to stopped-node local database maintenance with backups and positive row identification. The relocation to operations/pre-26.06.1-token-upgrade.md with a pointer from the 26.06.1 release notes is what I asked for, and the relative links resolve. I will handle the docs.basekick.net publication and a correction on #457 maintainer-side. Thanks for the careful rewrite.

PR #674, opened 2026-09-01
fix: bound the local directory cache

### LINKED ISSUE #318: medium(storage): LocalBackend.dirCache grows unboundedly
local.go:23 — Fix: add eviction.

mah1104ahm:
## Summary

- cap the local backend's directory cache at 1,024 entries
- evict a cached directory before inserting a new entry at capacity
- add a regression test that creates more directories than the bound

## Verification

- `gofmt` on changed Go files
- `go test ./internal/storage`
- `git diff --check`

Closes #318


xe-nvdk (changes requested):
Thanks, the fix is correct and nicely minimal. Evict-then-insert keeps the map at the 1,024 bound, the eviction runs under the write lock, and since a cache miss just repeats an idempotent `MkdirAll`, the arbitrary-entry eviction policy is a safe choice here (a miss costs one syscall, not correctness). The regression test covers the bound and `go test ./internal/storage` passes on the branch.

One ask before merging: please add a `## Fixed:` entry to `RELEASE_NOTES_2026.09.2.md`, matching the entries the sibling PRs in this series carry (see #663 through #667 and #670).

xe-nvdk (approved):
To move this along I pushed the remaining ask to your branch as a maintainer commit: merged current `main` and added the 26.09.2 release-notes entry under `## Bug fixes`, with a credit line naming you and the PR (the convention the file uses now). The code itself was already verified in the first review; re-verified after the merge that `go build ./...`, `go vet ./...`, and `go test ./internal/storage` are clean. Merging once CI is green.

PR #675, opened 2026-09-01
fix: prune expired tier cache entries

### LINKED ISSUE #343: medium(tiering): tierCache expired entries never evicted
metadata.go:29 — Fix: periodic sweep.

mah1104ahm:
## Summary

- prune expired tier-cache entries whenever a fresh cache result is stored
- keep the scan under the existing cache mutex
- add a regression test proving expired entries are removed while live entries remain

## Verification

- `gofmt` on changed Go files
- `go test ./internal/tiering`
- `git diff --check`

Closes #343

### DIFF
--- a/internal/tiering/metadata.go
+++ b/internal/tiering/metadata.go
@@ -16,6 +16,14 @@ type tierCacheEntry struct {
 	expiresAt time.Time
 }
 
+func (s *MetadataStore) pruneExpiredTierCache(now time.Time) {
+	for key, entry := range s.tierCache {
+		if !now.Before(entry.expiresAt) {
+			delete(s.tierCache, key)
+		}
+	}
+}
+
 // tierCacheTTL is how long tier lookups are cached (30 seconds)
 const tierCacheTTL = 30 * time.Second
 
@@ -545,6 +553,7 @@ func (s *MetadataStore) GetTiersForMeasurement(ctx context.Context, database, me
 
 	// Update cache
 	s.tierCacheMu.Lock()
+	s.pruneExpiredTierCache(time.Now())
 	s.tierCache[cacheKey] = &tierCacheEntry{
 		tiers:     tiers,
 		expiresAt: time.Now().Add(tierCacheTTL),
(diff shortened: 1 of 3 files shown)


internal/tiering/metadata.go
```diff
@@ -16,6 +16,14 @@ type tierCacheEntry struct {
 	expiresAt time.Time
 }
 
+func (s *MetadataStore) pruneExpiredTierCache(now time.Time) {
```
xe-nvdk, line 20: Non-blocking: a short doc comment noting that the caller must hold `tierCacheMu` would help the next reader, since the method touches `s.tierCache` without locking internally.

xe-nvdk (changes requested):
Thanks for this one too. #343 suggested a periodic sweep, and I think prune-on-store as implemented here is the better shape: no background goroutine to manage, every insert sweeps, so the cache stays bounded by the live entries plus one TTL window of stores. The expiry comparison (`!now.Before(expiresAt)`) matches the read path's validity check exactly, the scan stays under `tierCacheMu`, and the O(n) sweep only runs on a miss that is already paying for a SQLite query. Regression test is on point and `go test ./internal/tiering` passes on the branch.

One ask before merging: please add a `## Fixed:` entry to `RELEASE_NOTES_2026.09.2.md`, matching the entries the sibling PRs in this series carry (see #663 through #667 and #670).

mah1104ahm:
Thank you for the review. I pushed 9586dad after updating from main.

The release note is now under Bug fixes in RELEASE_NOTES_2026.09.2.md,
following the current contribution guide, and pruneExpiredTierCache
explicitly documents that its caller must hold tierCacheMu.

Validation: go test ./internal/tiering and go vet ./internal/tiering
both passed locally. Thank you for checking the original regression as
well.

M Ahmed

xe-nvdk (approved):
Re-reviewed 9586dad. The release notes entry is in place with the credit line, and pruneExpiredTierCache now documents the tierCacheMu precondition. Verified locally on the branch: go test -race ./internal/tiering passes, including the new pruning regression. Thanks.

Issue #635, opened 2026-08-20
high(backup): live shared-SQLite backup/restore is unsafe — torn copies on backup, stale-WAL corruption on restore


xe-nvdk:
Found during the 2026-08-20 adversarial audit (Iceberg lens, but the defect is in backup/restore of the shared SQLite DB — auth tokens, audit, tiering, MQTT, and the Iceberg catalog). Two halves, one fix domain.

## Half 1 — restore overwrites the live, open, WAL-mode DB

`internal/backup/restore.go:241` does a plain `os.WriteFile(destPath, data, 0600)` over the live database, on a **running** server (`POST /api/v1/backup/restore`, detached goroutine), with no writer pause and no removal of the stale `-wal`/`-shm` sidecars. The auth manager holds the file open in WAL mode (`internal/auth/auth.go:188`); with Iceberg enabled the reconciler is a periodic writer to the same handle every `reconcile_interval`.

Open connections hold page caches and the shm-index for the **old** content, and the pre-restore `-wal` (frames salted against the old DB) survives beside the restored file. The next write or checkpoint — an iceberg catalog commit mid-reconcile, an auth `last_used` update — mixes stale WAL frames / cached pages into the restored database, corrupting it. Docs never say to stop/restart the server for a metadata restore (only `restoreConfig` mentions restart).

## Half 2 — backup copies the live DB without a lock

`internal/backup/backup.go:440-477` checkpoints (`wal_checkpoint(TRUNCATE)`) on a private connection, closes it, then copies the file with plain `os.Open` + `ReadTo`. Between checkpoint and end-of-copy, live writers (auth, audit, iceberg catalog commits) keep committing; once the WAL passes the auto-checkpoint threshold (~1000 pages), a checkpoint writes pages into the main file **mid-copy** → silently corrupt `arc.db` / `iceberg-catalog.db` in the backup, discovered only at restore time.

## Fix shape

- Backup: use the SQLite Online Backup API or `VACUUM INTO` (or hold a read transaction across the copy).
- Restore: at minimum remove `destPath+"-wal"`/`-shm` after the write, log + document "restart required"; the real fix is refusing metadata restore while subsystems hold the DB, or restoring to a sidecar path applied at next boot.

PR #678, opened 2026-09-01
fix(backup): make live SQLite backup and restore safe

atirna:
## Summary

- snapshot live SQLite databases with `VACUUM INTO` before streaming them into a backup
- replace restored database files by rename and remove stale SQLite WAL/SHM sidecars
- mark metadata restores as requiring a server restart so open database connections are released

Refs #635

## Test plan

- `go test ./internal/backup -run 'Test(BackupSQLiteFile_SnapshotIsImmuneToMidCopyWriters|RestoreSQLiteFile_RemovesStaleSidecarsAndSnapshotsLiveState|RestoreSQLiteFile_RestoredFileIsImmuneToPreRestoreHandles)' -count=1 -v`

### DIFF
--- a/internal/backup/backup.go
+++ b/internal/backup/backup.go
@@ -433,33 +433,55 @@ func (m *Manager) backupSQLite(ctx context.Context, backupID string) error {
 	return m.backupSQLiteFile(ctx, backupID, m.sqliteDBPath, "arc.db")
 }
 
-// backupSQLiteFile copies one SQLite database file into the backup under
-// metadata/<destName>. It performs a WAL checkpoint first so the copy is
-// consistent, and streams via the storage backend rather than reading the whole
-// database into memory.
-func (m *Manager) backupSQLiteFile(ctx context.Context, backupID, dbPath, destName string) error {
-	// Checkpoint WAL to ensure all data is flushed to the main DB file.
+// snapshotSQLite writes a consistent snapshot of the live database at dbPath to
+// a fresh temporary file and returns its path. The caller must remove it.
+//
+// VACUUM INTO reads the database (including un-checkpointed WAL frames) under
+// one read transaction, so concurrently committing writers cannot interleave
+// pages into the snapshot — the failure mode of checkpoint-then-copy, where a
+// checkpoint between the copy's start and end rewrites the main file mid-read
+// once the WAL passes the auto-checkpoint threshold.
+func snapshotSQLite(ctx context.Context, dbPath string) (string, error) {
+	dir, err := os.MkdirTemp(filepath.Dir(dbPath), ".arc-snapshot-*")
+	if err != nil {
+		return "", fmt.Errorf("failed to create snapshot directory: %w", err)
+	}
+	snapshotPath := filepath.Join(dir, "snapshot.db")
+
 	db, err := sql.Open("sqlite3", dbPath)
 	if err != nil {
-		return fmt.Errorf("failed to open SQLite for checkpoint: %w", err)
+		os.RemoveAll(dir)
+		return "", fmt.Errorf("failed to open SQLite for snapshot: %w", err)
 	}
-	if _, err := db.ExecContext(ctx, "PRAGMA wal_checkpoint(TRUNCATE)"); err != nil {
-		db.Close()
-		return fmt.Errorf("WAL checkpoint failed: %w", err)
+	defer db.Close()
+	if _, err := db.ExecContext(ctx, "VACUUM INTO ?", snapshotPath); err != nil {
+		os.RemoveAll(dir)
+		return "", fmt.Errorf("SQLite snapshot failed: %w", err)
 	}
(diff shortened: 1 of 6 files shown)


internal/backup/backup.go
```diff
@@ -433,33 +433,55 @@ func (m *Manager) backupSQLite(ctx context.Context, backupID string) error {
...
 	}
-	if _, err := db.ExecContext(ctx, "PRAGMA wal_checkpoint(TRUNCATE)"); err != nil {
-		db.Close()
-		return fmt.Errorf("WAL checkpoint failed: %w", err)
+	defer db.Close()
+	if _, err := db.ExecContext(ctx, "VACUUM INTO ?", snapshotPath); err != nil {
+		os.RemoveAll(dir)
+		return "", fmt.Errorf("SQLite snapshot failed: %w", err)
+	}
+	return snapshotPath, nil
```
xe-nvdk, line 465: SQLite creates the `VACUUM INTO` target with its default file mode (0644 masked by umask). The 0700 temp directory shields it, and the copy streamed into backup storage gets the backend's permissions, so this is defense in depth rather than a hole: still, since this snapshot can contain the auth token database, an explicit `os.Chmod(snapshotPath, 0600)` after the vacuum would keep it consistent with the 0600 convention the surrounding code uses (and with the sidecar-permission hardening in flight in #666).

internal/backup/restore.go
```diff
@@ -225,24 +228,65 @@ func (m *Manager) restoreSQLiteFile(ctx context.Context, backupID, srcName, dest
...
 			}
+			os.RemoveAll(filepath.Dir(snapshotPath))
+		} else {
+			m.logger.Warn().Err(snapErr).Msg("Failed to snapshot database before restore")
 		}
 	}
 
-	if err := os.WriteFile(destPath, data, 0600); err != nil {
+	// Swap the file in by rename rather than overwriting the live inode: an
+	// open connection keeps reading and writing the old (now unlinked) inode,
```
xe-nvdk: One refinement to this comment's claim: it holds per raw SQLite connection, and subsystems here use `database/sql` pools. A pool that grows after the swap opens its new connections against the restored file by path, so post-restore writes can split across the old inode (existing connections) and the restored file (new pool connections). Each side stays internally consistent, so this does not reintroduce the torn-file corruption, and restart-required already covers it. I would just word the comment so it does not promise that all pre-restore writers stay off the restored file, because a growing pool will not.

xe-nvdk (changes requested):
Thanks, this is a strong PR on a sensitive code path. The approach matches what #635 asked for on both halves: `VACUUM INTO` freezes the backup source under one read transaction, the restore swaps by rename and removes the stale `-wal`/`-shm` sidecars, and the `.before-restore` safety copy is now itself a consistent snapshot instead of a torn file read. The tests are the best part: they reproduce the mid-copy checkpoint race and the stale-sidecar hazard deterministically rather than asserting around them. I ran the full `./internal/backup` suite on the branch and it passes, and I verified the `VACUUM INTO ?` bound parameter works against the vendored driver via those tests.

Three items:

1. Please add a `## Fixed:` entry to `RELEASE_NOTES_2026.09.2.md`, matching the entries the sibling fix PRs carry (see #663 through #667 and #670). This one deserves a visible entry more than most.
2. The PR says `Fixes #635`, and the issue names a deeper restore fix (refusing metadata restore while subsystems hold the DB, or staging the restore to apply at next boot). What lands here is the issue's minimum plus the rename swap and the `restart_required` signal, which I think is the right scope for one PR. I'd change `Fixes` to `Refs` and let a maintainer decide whether to keep #635 open for the apply-at-boot design, or file a follow-up issue for it.
3. Two smaller notes inline.

internal/backup/backup.go
atirna: The snapshot is now chmodded to 0600 immediately after VACUUM INTO and before it is read into backup storage.

internal/backup/restore.go
atirna: Updated this to cover pool growth: new connections can open the restored path, so restart is required before further writes.

atirna:
thx for the review

- added the `## Fixed:` release-note entry
- switched the PR link to `Refs #635`; the restart-required signal is unchanged, so #635 stays open for the apply-at-boot design

xe-nvdk (approved):
All four review items verified on the branch: the snapshot is chmodded to 0600 right after `VACUUM INTO`, the restore comment now describes pool growth honestly, the PR references `Refs #635` so the apply-at-boot design stays tracked, and the release-notes entry is in. Full `./internal/backup` suite passes locally on the updated head.

One maintainer-side touch pushed to your branch: moved the release-notes entry from a standalone `## Fixed:` heading into the shared `## Bug fixes` section and added the contributor credit line, which is the convention this file settled on today. Merging once CI is green.

PR #679, opened 2026-09-01
fix(compaction): avoid retaining evicted job history entries (#315)

### LINKED ISSUE #315: medium(compaction): jobHistory trimming retains old entries
manager.go:305 — slow memory leak. Fix: copy to new array.

bferanmi806-sketch:
Closes #315.

Trimming now copies the newest 100 job-history entries into fresh slice storage instead of reslicing the old backing array. Adds a deterministic regression test that asserts backing-array separation while preserving ordering and map references. Includes the 26.09.2 release-note entry.

Local tagged/race validation is blocked by the Windows DuckDB/cgo environment; Linux CI is the authoritative full validation.


xe-nvdk (approved):
Verified locally on Linux (build, vet, full `./internal/compaction` suite with `-count=1`): the copy-on-trim matches #315's prescription, and the regression test's address comparison with the cap-101 setup is a genuinely deterministic way to prove backing-array separation. A side benefit worth noting: `Stats` hands out `recent_jobs` as a reslice of the live history, and with the trim now bounding the backing array at ~100 entries, a caller holding that snapshot can no longer retain an unbounded array either.

Pushed one maintainer commit to the branch: merged current `main` and moved your release-notes entry under the shared `## Bug fixes` section with the contributor credit line, per the convention this file settled on today. Merging once CI is green.

PR #689, opened 2026-09-02
fix(backup): report restart_required for config-only restores

atirna:
## Summary

Two leftovers from the review of #678:

- `POST /api/v1/backup/restore` returned `restart_required: true` only for metadata restores. A config-only restore needs the same restart — `restoreConfig` already logs "(restart required)" because the restored `arc.toml` only takes effect on reload — but the response didn't say so. The flag now covers both, and a data-only restore still omits it.
- `TestRestoreSQLiteFile_RestoredFileIsImmuneToPreRestoreHandles` skipped when the fresh connection's query over the restored database errored. That readability is the behavior under test, so it fails now instead of skipping.

## Test plan

- `TestRestoreBackupReportsRestartRequired` drives the route through the existing rig: metadata restore → flag true; config-only restore → flag true (fails on current `main`, which returns no flag); data-only restore → flag absent.
- `go test -tags=duckdb_arrow -race ./internal/backup/... ./internal/api/...` — ok; `go vet -tags=duckdb_arrow ./...` — clean.

### DIFF
--- a/RELEASE_NOTES_2026.09.2.md
+++ b/RELEASE_NOTES_2026.09.2.md
@@ -55,18 +55,6 @@ literals (e.g. a log-search `LIKE '%INTERVAL 5 MINUTE%'`) and identifiers such a
 
 ## Bug fixes
 
-### Config-only restores now report `restart_required` (follow-up to [#678](https://github.com/Basekick-Labs/arc/pull/678))
-
-`POST /api/v1/backup/restore` returned `restart_required: true` only when
-metadata (SQLite databases) was restored. A config-only restore needs the same
-restart — the restored `arc.toml` only takes effect on reload, which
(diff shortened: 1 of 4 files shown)


xe-nvdk (approved):
Verified on the branch: both suites pass with the race and duckdb flags, the config-only case is exactly right (`restoreConfig` was already logging restart-required with no API signal), and converting the immunity test's skip into a hard failure is the correct call — the readability of the restored file IS the property under test, and it held over repeated runs here. Also appreciated: the release-notes entry with credit line arrived pre-formatted, which is exactly the flow CONTRIBUTING describes. One maintainer commit pushed to the branch moving the entry to the file's merge-order position. Merging once CI is green.

PR #696, opened 2026-09-03
fix(wal): chunk oversized payloads so wide-row ingest stays durable (#677)

### LINKED ISSUE #677: WAL payload byte cap is irreconcilable with the row-bounded ingest buffer on wide rows: WAL silently records nothing
Split out from #676 (the "related" note at the bottom).

With very wide rows, the ingest buffer's bound and the WAL's bound cannot be reconciled by configuration, and the result is that the WAL silently records nothing while reporting healthy ingest.

### The mismatch

- The ingest buffer is bounded by **row count** (100,000 rows).
- A WAL record is bounded by **bytes** (`WAL payload exceeds maximum allowed size`, cap 104,857,600 bytes).

At tens of KB of line protocol per row — wide industrial/telemetry schemas with hundreds of fields — a buffer flush of even ~1,500 rows produces a 700+ MB WAL payload, several times the cap. Every flush of that measurement fails the size check, so its data never reaches the WAL at all. There is no cap value that fixes it: any byte cap large enough for 100k wide rows is absurd, and any row bound small enough to fit the cap throttles narrow-row ingest.

### Observed behavior

During sustained bulk ingest of wide rows, each writer's WAL directory holds a single 7-byte file (header only, no records) for hours of live traffic, while the size-check failure fires roughly once per import request for the widest measurement. Operationally this is worse than a loud failure: writes are accepted and durable only in Parquet flushes, and the operator believes the WAL is protecting the window in between.

### Proposed fixes (either resolves it)

- [ ] **Chunk WAL record writes:** split an oversized flush into multiple WAL records under the cap instead of rejecting it wholesale.
- [ ] **Bound the ingest buffer by bytes as well as rows:** flush whichever bound is hit first, sized so a full flush always fits one WAL record.

And in either case:

- [ ] **Make the failure loud:** a WAL write rejected for size should surface in metrics/health, not just a log line — a WAL that records nothing while ingest runs should never look healthy.

atirna:
## Description

Wide ingest requests (a single client body with tens of KB per row) can exceed
the WAL's 100MB single-entry payload cap while staying well under the server's
1GB request limit. Before this change every such write was rejected wholesale
by the WAL: ingest kept reporting healthy while each writer's WAL directory
held a 7-byte header-only file for hours, and the only trace of the rejection
was a log line. There was no cap value that fixed it, any byte cap large
enough for a full row-bounded buffer is absurd, and any row bound small enough
to fit the cap throttles narrow-row ingest.

The WAL now splits an oversized payload into multiple entries at msgpack
element boundaries, so the same request's data reaches the WAL as N
size-valid entries:

- row-format payloads (a top-level msgpack array of records) split between
  records, each chunk a shorter array of the same records
- columnar payloads (the zero-copy path's top-level map with a `columns`
  sub-map) split by row range; `m` carries over and each column array is
  sliced to the same `[lo, hi)` rows

Chunks are self-contained msgpack values, so the reader replays them with no
format change: concatenating the chunks' decoded rows reproduces the original
request. A payload that still cannot be split (a single record or single row
larger than the cap) keeps the loud `ErrPayloadTooLarge` rejection, and that
rejection is now counted in the `arc_wal_oversized_payloads_total` metric
instead of living only in a log line.

Fixes #677

## Why

The WAL is the durability window between ingest accepting a write and the
Parquet flush landing. An operator watching a bulk load of wide rows believed
that window was protected; it was not, and nothing on the metrics endpoint
said so. Chunking removes the mismatch at the boundary that owns it (the WAL
entry size), which resolves the issue for both proposed fix shapes without a
config change and without throttling narrow-row ingest.

## Verification

- before, on current `main`: a 117.2 MiB columnar msgpack request fed through
  the real decode and buffer write path with a real WAL writer attached is
  accepted by ingest while the WAL directory holds a single 7-byte
  header-only file, and the rejection appears only as a `WAL write failed -
  data may be lost on crash` log line
- after: the same request produces multiple WAL entries each under the cap;
  reading the file back with the WAL reader returns entries whose decoded row
  union equals the request, and replaying them through the recovery path
  (`WriteColumnarDirectNoWAL`) applies every row
- new regression tests fail on the pre-fix tree and pass after:
  `go test ./internal/wal/ -run 'ChunksOversized|SingleOversized|PayloadAtLimitSingleEntry'`
  and `go test ./internal/ingest/ -run 'TestWideRequest'`
- full focused suites green: `go test ./internal/wal/ ./internal/metrics/
  ./internal/ingest/ ./internal/cluster/... -count=1`, plus the new tests
  under `-race`; `gofmt -l` empty and `go vet` clean on the touched packages

## Notes

- chunk target is 32MB, sized so every chunk clears the 100MB single-entry
  cap after container headers, the WAL envelope, and the JSON + base64
  envelope replication wraps around a payload
- replication streams each chunk as its own `ReplicateEntry` under the
  protocol's own 100MB message cap

### DIFF
--- a/RELEASE_NOTES_2026.09.2.md
+++ b/RELEASE_NOTES_2026.09.2.md
@@ -55,6 +55,21 @@ literals (e.g. a log-search `LIKE '%INTERVAL 5 MINUTE%'`) and identifiers such a
 
 ## Bug fixes
 
+### Wide-row ingest requests no longer silently bypass the WAL ([#677](https://github.com/Basekick-Labs/arc/issues/677))
+
+A single ingest request larger than the WAL's 100MB per-entry payload cap
+(wide rows make this easy to hit well under the 1GB request limit) was
+rejected wholesale by the WAL, so ingest kept reporting healthy while the
+WAL directory held only a 7-byte header file: writes were durable in
+Parquet flushes alone and nothing on the metrics endpoint said so. The WAL
+now splits an oversized payload into multiple entries at msgpack element
+boundaries (row-format requests between records, columnar requests by row
+range), each under the cap and independently replayable with no format
+change. A payload that still cannot be split, a single record or row above
+the cap, keeps the loud rejection, now also counted in the new
+`arc_wal_oversized_payloads_total` metric so a WAL that records nothing
+never looks healthy again.
+
 ### S3 query paths now follow calendar days across DST ([#321](https://github.com/Basekick-Labs/arc/issues/321))
 
 S3 time-range path generation now builds one inclusive partition path per UTC
--- a/RELEASE_NOTES_2026.09.2.md
+++ b/RELEASE_NOTES_2026.09.2.md
@@ -264,3 +264,5 @@ warehouse metadata directory, after the files are removed, so a racing
 reconcile pass can only empty tables rather than recreate them. Cleanup
 failures are logged and re-running the DELETE retries them, including when the
 files are already gone.
+
+Contributed by [@atirna](https://github.com/atirna) in [#696](https://github.com/Basekick-Labs/arc/pull/696).
--- a/internal/ingest/wal_wide_payload_test.go
+++ b/internal/ingest/wal_wide_payload_test.go
@@ -1,0 +1,98 @@
+package ingest
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/Basekick-Labs/msgpack/v6"
+	"github.com/basekick-labs/arc/internal/wal"
+	"github.com/rs/zerolog"
+)
+
+// TestWideRequestReachesWALAndReplays is the #677 repro end to end: a wide
+// columnar msgpack request above the WAL single-entry cap goes through the
+// real decode -> ArrowBuffer.Write path with a real WAL writer attached, the
+// WAL holds the request's rows as chunked entries instead of a header-only
+// file, and those entries replay through the recovery callback the same way
+// boot recovery would apply them.
+func TestWideRequestReachesWALAndReplays(t *testing.T) {
+	tmp := t.TempDir()
+	writer, err := wal.NewWriter(&wal.WriterConfig{
+		WALDir: filepath.Join(tmp, "wal"), SyncMode: wal.SyncModeAsync,
+		MaxSizeBytes: 1024 * 1024 * 1024, BufferSize: 100000, Logger: zerolog.Nop(),
+	})
+	if err != nil {
+		t.Fatalf("wal.NewWriter: %v", err)
+	}
+	buf := newReplayTestBuffer(t, filepath.Join(tmp, "data")) // package-standard buffer+storage wiring
+	t.Cleanup(func() { buf.Close() })
+	buf.SetWAL(writer)
+
+	// The reporter's payload: time + 10 fields of ~10 KB per row, above the
+	// WAL single-entry cap.
+	const rows = 1200
+	wide := strings.Repeat("w", 10*1024)
+	timeCol := make([]interface{}, rows)
+	for i := range timeCol {
+		timeCol[i] = int64(1_700_000_000_000_000 + i)
+	}
+	wideRow := make([]interface{}, rows)
+	for i := range wideRow {
+		wideRow[i] = wide
+	}
+	columns := map[string]interface{}{"time": timeCol}
+	for f := 0; f < 10; f++ {
+		columns[fmt.Sprintf("field_%d", f)] = wideRow
+	}
+	payload, err := msgpack.Marshal(map[string]interface{}{"m": "cpu", "columns": columns})
+	if err != nil || len(payload) <= wal.MaxWALPayloadSize {
+		t.Fatalf("marshal (%v) or payload %d not above cap", err, len(payload))
+	}
+
+	// The live path: decode the request, write the decoded records.
+	decoded, err := NewMessagePackDecoder(zerolog.Nop()).Decode(payload)
+	if err != nil {
+		t.Fatalf("decode: %v", err)
+	}
+	if err := buf.Write(context.Background(), "default", decoded); err != nil {
+		t.Fatalf("write: %v", err)
+	}
+	walPath := writer.CurrentFile()
+	if err := writer.Close(); err != nil {
+		t.Fatalf("wal close: %v", err)
+	}
+	// The WAL must not be header-only, and its entries must cover every row.
+	if info, err := os.Stat(walPath); err != nil {
+		t.Fatalf("stat: %v", err)
+	} else if info.Size() <= 7 {
+		t.Fatalf("WAL is header-only (%d bytes): the wide request never reached the WAL", info.Size())
+	}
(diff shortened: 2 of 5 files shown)


xe-nvdk (changes requested):
Thorough work on a delicate path, and the analysis in the description matches what I measured. Verified locally on the branch:

- the #677 repro behaves exactly as described: on current main, the wide columnar request leaves a 7-byte header-only WAL file; on this branch it lands as chunked entries whose replayed row union equals the request
- both new regression tests fail on the pre-fix tree and pass here; `go test ./internal/wal/ ./internal/metrics/ ./internal/ingest/ -count=1` green, the new tests green under `-race`, `go vet` and `gofmt -l` clean
- type fidelity through the columnar re-encode holds in an adversarial probe (int64, max uint64, float32, NaN, nil, bool, bin, msgpack ext timestamps), row alignment across columns preserved, every chunk under the cap
- envelope-marked payloads (0x01, the replication receiver path) fall through the splitter untouched and keep the loud rejection, and `RawMessage` passthrough in the msgpack fork is verbatim, so the interleaved encoder writes in `encodeColumnarRange` are sound

Three changes requested:

1. `encodeColumnarRange` panics on ragged columns. A columnar payload whose arrays differ in length crashes with `slice bounds out of range` at `arr[lo:hi]`: the row count is taken from one map entry and every other column is sliced to it. Today's ingest decoders validate lengths before the WAL sees the payload, so this is unreachable from client input, and that is exactly why it needs the guard here: the splitter runs on the request path inside the WAL layer, and its safety currently depends on a validation invariant that lives two packages away. Validate the decoded column lengths in `splitOversizedPayload` and turn a mismatch into the loud rejection path.

2. The columnar split's memory cost. Decoding every column to `[]interface{}` and re-encoding boxes each cell: I measured a 154MB narrow-row columnar payload (two columns, 9M rows) costing 4.0GB of total allocations with ~900MB held during the split, a ~26x amplification on the ingest request path. Bulk loads are the workload that hits this code, and several concurrent oversized requests would make this a real OOM risk. The array path already avoids this by carrying raw element bytes; the columnar path can do the same: walk each column's elements with `DecodeRaw` to record boundaries, then emit each chunk's column arrays by writing a new array header plus the raw byte range. That also makes chunks byte-faithful instead of re-encoded.

3. The release notes credit line landed at the end of the file, under the Iceberg database-drop entry, which is not yours. Move it into the #677 entry. The branch also predates today's merges of #673 and #675, which added entries at the same spot in the bug fixes section, so this file will need a rebase either way.

One smaller ask, fold into whichever change touches it: a `splitOversizedPayload` failure (msgpack parse error now, the ragged rejection after item 1) returns an error that bumps neither `arc_wal_oversized_payloads_total` nor `arc_wal_failed_writes_total`; only the per-buffer `total_wal_errors` counter sees it. Since the point of the metric is that a WAL recording nothing must not look healthy, every oversized-payload rejection path should count in `arc_wal_oversized_payloads_total`.

Two observations, no action needed in this PR: a chunk can legally reach ~100MB when a single element lands on a 32MB boundary, and anything above roughly 75MB then fails replication's `WriteMessage` after the JSON plus base64 envelope; that exposure predates this PR (any unsplit 75-100MB entry has it) so I would rather track it as a follow-up issue than grow this one. And a crash mid-sequence replays a prefix of the request's chunks, which is fine since rows are independent, but worth a sentence in the `splitOversizedPayload` doc comment.

atirna:
thx for the review.

pushed the rework:
- ragged column arrays reject before slicing, and every oversized rejection increments `arc_wal_oversized_payloads_total`; `arc_wal_failed_writes_total` stays tied to physical WAL write failures.
- column chunks now retain raw msgpack element bytes rather than decoding values to `[]interface{}`.
- moved the #696 credit into the #677 release-note entry and merged current `main`.

Focused WAL, ingest, metrics, and cluster checks, the WAL race test, vet, and gofmt are green.

xe-nvdk (changes requested):
Re-reviewed the rework. Three of the four asks are confirmed fixed, verified locally on the branch:

- ragged columns now reject deterministically with the column named in the error (`column "b" has 10 values, want 3000` in my probe), no panic, and the rejection bumps `arc_wal_oversized_payloads_total`; your regression test covers it
- every oversized rejection path, parse errors included, now routes through `oversizedPayloadError`, so the metric holds the #677 "never silently healthy" property
- credit line sits in the #677 entry and the rebase over today's merges is clean; the wal, metrics, and full ingest suites pass, the chunking and wide-request tests pass under `-race`, vet and gofmt clean; my type-fidelity probe still passes and chunks are now byte-faithful as a bonus

The memory ask is the one that did not land. Re-measuring the same 154MB narrow-row columnar payload (two columns, 9M rows): total allocations went from 4.0GB to 3.8GB, and the memory held during the split went up, from ~0.9GB to ~1.5GB. The cause is that `DecodeRaw` copies each element into its own allocation (`d.rec = make(...)` per call), so the per-element boxing was traded for 18M small copies plus the `[]msgpack.RawMessage` slice headers.

The shape that actually removes the cost: record offsets instead of bytes. The decoder reads a `bytes.Reader` unbuffered (it satisfies the fork's `bufReader` interface), so after any `Skip()` the absolute position is exactly `len(raw) - reader.Len()`. Per column: `DecodeArrayLen`, note the position as the first boundary, then `Skip()` through `rowsPerChunk` elements at a time, noting the position at each chunk boundary — that is O(chunks) bookkeeping, not O(rows). Emitting a chunk is then an `EncodeArrayLen(hi-lo)` followed by one `Write` of the original payload's byte range per column. Held memory drops to roughly the produced chunks, and the per-element allocations disappear entirely. The ragged check falls out for free by comparing each column's element count during the same walk.

Everything else is done; this last item is worth one more pass because the splitter runs on the ingest request path and the bulk loads that trigger it are exactly the workloads that run several oversized requests at once.

atirna:
pushed the offset-based column split. It now records only per-chunk offsets with `DecodeArrayLen` and `Skip`, then writes the original element spans, so the splitter no longer allocates one `RawMessage` per cell.

xe-nvdk (approved):
Approved. The offset-based split closes the last item, verified locally on c714a85:

- re-measuring the same 154MB narrow-row columnar payload (two columns, 9M rows): total allocations fell from 3.8GB to 386MB and held memory from ~1.5GB to ~322MB, which is the output chunks plus buffer growth and nothing else; the split also runs about twice as fast
- the position bookkeeping is sound: bytes.Reader satisfies the msgpack fork's unbuffered reader interface, so `len(payload) - reader.Len()` after `Skip()` is an exact absolute offset, and the O(chunks) boundary arrays are the only per-column state
- the ragged check moved into the same walk and still rejects cleanly with the column named, bumping `arc_wal_oversized_payloads_total`; the nil-column guard on `DecodeArrayLen` returning -1 is a nice touch
- chunks are byte-verbatim element spans now, and my fidelity probe (int64, max uint64, float32, NaN, nil, bool, bin, ext timestamps) plus row-alignment and cap checks all pass
- full wal, metrics, and ingest suites green; the chunking, ragged, and wide-request tests green under `-race`; vet and gofmt clean; CI green

Thanks for the careful iteration on this one. The WAL is better for it.

PR #697, opened 2026-09-03
fix(cluster): add periodic file replication reconciliation (#393)

### LINKED ISSUE #393: Periodic reconciliation for file replication
## Problem
The catch-up walker runs once at startup. If a reactive FSM callback is missed (dropped from the bounded queue), the file stays missing until the next restart. No periodic re-check exists.

## Proposed Solution
Add a periodic tick (e.g. every 5 minutes) that re-runs the catch-up walker against the manifest. ~20-line addition alongside the existing startup walker in `coordinator.go:1875`.

## Context
Deferred from Phase 3. Comment at `coordinator.go:1875`: "If Phase 5 adds periodic reconciliation, it will live alongside this startup-only walker."

## Effort
Small

bferanmi806-sketch:
## Summary

- Adds periodic peer-replication reconciliation to recover files when FSM callbacks are missed.
- Reuses the existing paginated pull, deduplication, and retry path.
- Keeps startup catch-up one-shot: periodic failures never reopen readiness, while successful reconciliation may heal previously recorded startup failure/drop state.
- Gates each tick on current registry membership, role, and health.
- Adds a configurable 5-minute default interval and separate replication_recheck_* observability.
- Adds missed-callback regression coverage.
- Leaves #412 shared scheduling/status intentionally out of scope.

## Testing

- Passed: go test ./internal/cluster/filereplication -count=1
- Passed: focused periodic reconciliation tests in ./internal/cluster/filereplication
- Passed: go test ./internal/config -count=1
- Passed: go vet ./internal/cluster/filereplication
- Full coordinator tests, including the focused membership coverage, are blocked locally on Windows with CGO_ENABLED=0 because DuckDB requires CGO.

Fixes #393

### DIFF
--- a/RELEASE_NOTES_2026.09.1.md
+++ b/RELEASE_NOTES_2026.09.1.md
@@ -2,21 +2,6 @@
 
 > **Status:** Released 2026-08-31.
 
-## New: periodic peer file replication reconciliation
-
-Cluster nodes now re-walk the Raft file manifest every five minutes by default
-after the startup catch-up attempt. The interval is configurable with
-`cluster.replication_reconciliation_interval_seconds`. The pass uses the
-existing paginated manifest walk, pull queue, deduplication, retries, checksum
-verification, and local-file checks, so a missed reactive callback can be
-repaired without restarting the node. Healthy writer, reader, and compactor
-nodes may run the pass; joining, leaving, unhealthy, and standalone nodes are
-gated. Startup catch-up remains a one-shot operation and keeps its existing
-`catchup_*` readiness and query-gate semantics. The existing
-`cluster.replication_catchup_enabled` switch disables both startup and
-periodic reconciliation. Recheck outcomes are exposed under
-`replication_catchup_status` as `replication_recheck_*` counters.
-
 ## New: Apache Iceberg export (opt-in)
 
 Arc can now publish its data as **Apache Iceberg tables** so any Iceberg-aware engine — Spark, Trino, DuckDB, Snowflake, PyIceberg — can query Arc's data directly, without going through Arc's API. This is the strongest form of the "your data, no lock-in" promise: Arc already writes open Parquet files you own; now those same files are also a standard Iceberg lakehouse table.
--- a/RELEASE_NOTES_2026.09.1.md
+++ b/RELEASE_NOTES_2026.09.1.md
@@ -2,6 +2,17 @@
 
 > **Status:** Released 2026-08-31.
 
+## New: periodic peer file replication reconciliation
+
+Cluster nodes now re-walk the Raft file manifest every five minutes after the
+startup catch-up attempt. The pass uses the existing paginated manifest walk,
+pull queue, deduplication, retries, checksum verification, and local-file
+checks, so a missed reactive callback can be repaired without restarting the
+node. Startup catch-up remains a one-shot operation and keeps its existing
+`catchup_*` readiness and query-gate semantics. The existing
+`cluster.replication_catchup_enabled` switch disables both startup and
+periodic reconciliation.
+
 ## New: Apache Iceberg export (opt-in)
 
 Arc can now publish its data as **Apache Iceberg tables** so any Iceberg-aware engine — Spark, Trino, DuckDB, Snowflake, PyIceberg — can query Arc's data directly, without going through Arc's API. This is the strongest form of the "your data, no lock-in" promise: Arc already writes open Parquet files you own; now those same files are also a standard Iceberg lakehouse table.
--- a/RELEASE_NOTES_2026.09.1.md
+++ b/RELEASE_NOTES_2026.09.1.md
@@ -4,14 +4,18 @@
 
 ## New: periodic peer file replication reconciliation
 
-Cluster nodes now re-walk the Raft file manifest every five minutes after the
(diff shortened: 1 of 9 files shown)


xe-nvdk (changes requested):
This is careful work in a part of the codebase where careful matters. Verified locally on the branch:

- full `go test ./internal/cluster/... ./internal/config/` green (covering the coordinator suite you could not run on Windows), `./internal/cluster/filereplication/` green under `-race -count=2`, `go vet` clean; the one `gofmt -l` hit (`puller_test.go`) is pre-existing on main and untouched here
- the `TryRLock` reasoning in `canRunFileReconciliation` checks out: `Coordinator.Stop` holds `c.mu` while `puller.Stop()` waits for the reconciliation goroutine, so a blocking read lock in the gate would deadlock shutdown; your test pins it
- the `fsm` capture at the `StartPeriodicReconciliation` call site is the nil-checked local that also feeds the FSM callbacks, so the fetch closure cannot see a nil FSM
- defer ordering in `RunCatchUp` releases `reconciliationMu` before closing `catchupFinished`, so the scheduler cannot wake into a held walk lock; the no-overlap, dedup, and no-reopen tests all pass and test real races

Two changes requested:

1. A periodic pull that succeeds must heal `catchupFailed` and `catchupDropped`. The scenario: startup catch-up drops path X on a full queue, so `catchupDropped` goes to 1 and `FullyCaughtUp` reports false (with `query_gate_on_catchup` on, the node 503s). No reactive callback for X will ever come, its FSM apply already fired, and a missed callback is this PR's own premise. The periodic pass then pulls X successfully and the file is complete on disk, but `finishEntry` skips the clears for `enqueueSourceReconciliation`, so the gate stays red until a process restart, which is the outcome this feature exists to remove. The "never reopen startup readiness" rule is right for the failure-recording direction and for tag removal, and wrong for the success direction: `clearCatchUpFailureLocked` and `clearCatchUpDropLocked` only move the gate toward ready and are no-ops for paths startup never recorded. Let reconciliation-sourced success run the two clears, and flip the assertion in `TestPeriodicReconciliationDoesNotModifyCatchUpBookkeeping` from "did not heal" to "healed without touching the other startup counters".

2. The release notes entry is in RELEASE_NOTES_2026.09.1.md, which shipped on 2026-08-31 (its status line says Released). New behavior lands in the planned release file, RELEASE_NOTES_2026.09.2.md, alongside the other pending entries, and needs the standard credit line: `Contributed by [@bferanmi806-sketch](https://github.com/bferanmi806-sketch) in [#697](https://github.com/Basekick-Labs/arc/pull/697).`

Two smaller notes, neither blocks on its own:

- `runCatchUpOnce` is guarded by the coordinator-lifetime `sync.Once`, while `catchupFinished` belongs to the puller. A coordinator that goes through an in-process Stop/Start cycle builds a fresh puller whose `catchupFinished` nobody will ever close, so the periodic loop silently never runs there. Production wiring today only starts the coordinator once per process, so this is latent, and closing the channel on `runCatchUpOnce`'s early-return paths (or keying the once per puller start) would remove the trap cheaply.
- `walkManifest` evaluates `reconciliationGateRejected` per entry, and the gate does a registry lookup that clones a node each call: a million-entry manifest means a million registry hits per pass. Checking the gate once per page keeps the same responsiveness at page granularity and drops the overhead three orders of magnitude.

One question rather than an ask: `ReplicationReconciliationIntervalSeconds` at 0 or below silently becomes the 5-minute default, so the only way to turn the periodic pass off is `replication_catchup_enabled=false`, which also kills startup catch-up. If an operator should be able to keep startup catch-up while opting out of the periodic pass, 0 meaning disabled would be the natural knob; happy to leave that for a follow-up if you would rather not grow this PR.

bferanmi806-sketch:

bferanmi806-sketch:
recheck

xe-nvdk (approved):
The healing rework in cbdf3d2 lands exactly where it needs to. `finishEntry` now lets a successful pull from any source clear path-scoped `catchupFailed`/`catchupDropped` state, reconciliation failures still cannot record into startup bookkeeping, and startup tags remain untouched by reconciliation workers. I traced the tag lifecycle to confirm the tag guard is purely defensive: a drop unmarks before recording (catchup.go), so a tag only exists while a startup entry is queued or in flight, and queue dedup keeps a reconciliation pull from ever finishing against a tagged path. The reworked `TestPeriodicReconciliationHealsCatchUpBookkeeping` pins the healing behavior with tightened preconditions, and `TestPeriodicReconciliationDoesNotReopenStartupReadiness` still covers the isolation side.

The release notes entry is now in RELEASE_NOTES_2026.09.2.md with the credit line, and the updated wording (periodic failures never reopen readiness, successful reconciliation pulls may heal recorded failure or drop state) matches the implementation.

Verified locally on the branch: build, vet, and the full `internal/cluster/...` suites under `-race`, all green.

One thing before merge: the branch conflicts with main in RELEASE_NOTES_2026.09.2.md (the Bug fixes section gained entries from #673, #675, #696, and #698 since you cut the branch). Please rebase on main and keep your entry at the top of the Bug fixes section. CI cannot run on the PR until the conflict is resolved, so the rebase also unblocks the checks.

Issue #676, opened 2026-09-01
WAL fills the default 10Gi PVC during bulk ingest: no backpressure, boot deadlock on full volume, liveness kills recovery


xe-nvdk:
During sustained bulk ingest (snapshot backfill at ~100 MB/s), the write-ahead log can grow faster than it flushes. On a Kubernetes deployment with an S3/object-storage backend, the local PVC holds only the WAL and cache, and the chart's default `persistence.size: 10Gi` (`helm/arc/values.yaml`) fills in minutes at that rate. The resulting failure chain is worse than a full disk:

1. **WAL fills the volume with no backpressure.** Writes keep being accepted while the WAL grows unboundedly until the volume is full.
2. **Boot deadlock.** After the writer goes down, it cannot start again: creating the initial WAL file on startup requires free space on the already-full volume, so the pod can never come back without manual intervention (growing the PVC or deleting WAL segments by hand).
3. **Liveness kills recovery.** Once space is available, WAL replay after an unclean stop can exceed the liveness budget in `helm/arc/templates/deployment.yaml` (`periodSeconds: 10`, `failureThreshold: 3`, no `startupProbe`), so the pod is killed mid-recovery and crash-loops repeatedly before it manages to drain.

Observed in the field during a TB-scale migration; the operator's workaround was `ARC_WAL_ENABLED=false` for the migration window (throughput was unchanged, WAL stayed flat), which works but gives up durability for live writes during that window.

### Proposed fixes

- [ ] **Backpressure (core):** when WAL disk usage crosses a high-water mark, reject writes (503, or a typed backpressure error clients can retry) instead of growing until the volume is full. Reserve enough headroom that the writer can always create its initial WAL file on boot, so the deadlock in (2) becomes impossible.
- [ ] **`startupProbe` (chart):** add a `startupProbe` with a generous `failureThreshold × periodSeconds` window (minutes, not seconds) so WAL recovery is never killed by the liveness probe; keep the current livenessProbe for steady state.
- [ ] **Sizing guidance and defaults (chart/docs):** document WAL volume sizing for object-storage backends (roughly: peak ingest rate × worst-case flush lag, plus cache), call out that `10Gi` is a dev-scale default, and consider raising it or adding a loud comment in `values.yaml`.

### Related (probably a separate issue)

[...]

PR #700, opened 2026-09-04
fix(helm): protect WAL recovery with a startup probe

atirna:
## Summary

- Add a five-minute startup probe before the OSS chart's liveness probe starts, so WAL replay can finish without a crash loop.
- Document the 10Gi PVC as development-scale and describe the WAL sizing inputs for object-storage deployments.
- Check the installed Deployment's startup-probe contract in the existing Helm integration job.

Refs #676

## Test plan

- `helm lint ./helm/arc`
- `helm template arc ./helm/arc` checks the default `/health` startup probe and the unchanged liveness/readiness probes.
- `helm template arc ./helm/arc --set startupProbeFailureThreshold=45` checks the grace-period override.
- The startup-probe assertion fails against the pre-fix chart and passes after restoring the chart change.

### DIFF
--- a/.github/workflows/release-build.yml
+++ b/.github/workflows/release-build.yml
@@ -1491,6 +1491,8 @@ jobs:
             --set persistence.enabled=false \
             --wait --timeout=2m
 
+          bash scripts/test_startup_probe.sh helm/arc
+
           # Check if deployment is ready
           kubectl rollout status deployment/arc-test --timeout=2m
 
--- a/RELEASE_NOTES_2026.09.2.md
+++ b/RELEASE_NOTES_2026.09.2.md
@@ -73,7 +73,7 @@ The default 10Gi PVC is documented as development-scale; object-storage
 deployments should size it for peak ingest, WAL flush lag, cache, and recovery
 headroom.
 
-Contributed by [@atirna](https://github.com/atirna) in [#676](https://github.com/Basekick-Labs/arc/issues/676).
+Contributed by [@atirna](https://github.com/atirna) in [#700](https://github.com/Basekick-Labs/arc/pull/700).
 
 ### Periodic peer file replication reconciliation repairs missed FSM callbacks ([#393](https://github.com/Basekick-Labs/arc/issues/393))
 
--- a/RELEASE_NOTES_2026.09.2.md
+++ b/RELEASE_NOTES_2026.09.2.md
@@ -65,6 +65,16 @@ The ingest, API, query, and Iceberg test suites were verified against 0.24.0.
 
 ## Bug fixes
 
+### Helm recovery no longer crash-loops during long WAL replay ([#676](https://github.com/Basekick-Labs/arc/issues/676))
+
+The OSS Helm chart now uses a five-minute startup probe before liveness starts,
+so a pod replaying its WAL can recover without Kubernetes repeatedly killing it.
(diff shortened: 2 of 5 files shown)


xe-nvdk (changes requested):
This lands the right fix for the liveness prong of #676, and the design checks out at every level I could test. The code itself is ready; two administrative items block the merge, listed at the end.

What I verified:

**Source ordering.** The probe comment's claim is accurate: WAL recovery runs synchronously in `main()` (RecoverWithOptions, cmd/arc/main.go) well before `api.NewServer` is constructed and Fiber binds the listener, so `/health` cannot answer until replay finishes. The startup probe therefore measures exactly the window that needs protecting.

**Static checks.** `helm lint` clean, `helm template` renders the probe correctly at the default and with `--set startupProbeFailureThreshold=45`, `scripts/test_startup_probe.sh` passes on this branch and fails against the pre-fix chart (regression coverage confirmed), and shellcheck reports nothing.

**Live cluster A/B.** I installed both charts side by side on a local k3s cluster with the container command wrapped in a 120-second sleep before exec, simulating a long replay ahead of the listener bind:
- Pre-fix chart: liveness killed the container on a ~70s cycle ("Container arc failed liveness probe, will be restarted"), 4 restarts in 5 minutes, never Ready. This reproduces the #676 crash loop exactly.
- This branch: the startup probe absorbed 12 failed checks against its budget of 30, zero restarts, pod Ready at 2m32s.

Optional polish, fine to fold into the rebase or skip:
- `startupProbe` inherits `timeoutSeconds: 1` while liveness uses 5. During replay the port refuses instantly so it rarely matters, and a slow first `/health` right after the listener binds would only cost one 10s period. Setting 5 for consistency is cheap.
- `check_probe 30` passes `--set` with the default value, so the values.yaml default itself is never exercised; one run without `--set` would pin it.
- The assertions exit silently on failure under `set -e`; an error echo would make a CI failure readable.
- The CI call templates `helm/arc` while the job installs `dist/arc-*.tgz`; pointing the script at the tgz would test the exact artifact being released.

1. **CLA.** The CLA check is failing; signing is now required on every PR (#699). Please sign per the bot's instructions in this thread and comment `recheck`.
2. **Rebase.** RELEASE_NOTES_2026.09.2.md conflicts with main (the Bug fixes section gained the #697 entry this morning; that is the only conflicting file). Please rebase and keep your entry at the top of the section.

atirna:
rebased onto main and kept the #676 entry at the top of Bug fixes.

atirna:
thx for the review. I pushed the test follow-up: the default chart render is now checked without an override, the 45 override stays covered, and assertion failures name the missing probe field. `helm lint ./helm/arc` and the startup-probe script pass.

kept `timeoutSeconds` and the packaged-chart check as-is for this chart-only test follow-up.

atirna:

xe-nvdk (approved):
CLA is green and the rebase is clean: the chart, workflow, and release-notes content match what was verified in the first round (including the live cluster A/B), with your entry at the top of the Bug fixes section. The new test commit improves on two of the optional notes: the default case now templates without --set, so the values.yaml default is actually exercised, and the assertions report what failed instead of exiting silently. I re-ran the script against a sabotaged threshold to confirm the failure message fires. Merging.

PR #701, opened 2026-09-04
fix(query): apply queryTimeout to queryMeasurement endpoint

### LINKED ISSUE #308: high(query): queryMeasurement endpoint has no timeout
query.go:2997-3006 — uses context.Background(). Fix: apply queryTimeout.

MrBeldum:
## Summary
- `GET /api/v1/query/:measurement` executed against `context.Background()` and `db.Query()` with no deadline, so configured `queryTimeout` never applied (#308).
- Wrap `c.UserContext()` with `queryTimeout` (same pattern as `POST /api/v1/query`), pass the cancel func into the Arrow dispatch, and use `QueryContext` on the database/sql fallback.
- Timeout now returns 504 `"Query timed out"` and increments timeout metrics. Stream and error paths cancel the context.

Fixes #308

## Test plan
- [x] `go test -tags=duckdb_arrow ./internal/api/` (full package, including new `TestQueryMeasurementAppliesQueryTimeout`)
- [x] Confirm a configured `query_timeout` actually 504s a slow `GET /api/v1/query/:measurement` in a running server (verified in review: 504 at 1.5s with query.timeout=1, timeout metric incremented; pre-fix build ran 9.5s past the timeout)


xe-nvdk (changes requested):
Thanks for taking #308. This is the right fix, and it verified out at every level I tested. Two administrative items block the merge, listed at the end.

What I verified:

**Cancel ownership.** The contract with `executeArrowJSONQuery` holds exactly as your comment states: every `handled=true` path releases the cancel func (immediate error paths directly, the success path inside the stream callback), and the `handled=false` driver-fallback path returns without cancelling, so the database/sql fallback receives a live context. On the fallback itself, every return path after the context is created releases cancel: the QueryContext error, the Columns error, and the stream writer callback. The ColumnTypes error deliberately falls through, so nothing leaks there either.

**Pattern match.** The wiring matches the established handlers: UserContext as the base so client disconnects propagate, `ctx.Err() == context.DeadlineExceeded` for the 504 classification, cancel called in the stream callback rather than deferred because SetBodyStreamWriter runs after the handler returns.

**Tests.** Full `internal/api` suite green under `-race`, with and without `-tags=duckdb_arrow`. `TestQueryMeasurementAppliesQueryTimeout` fails against the pre-fix handler, so the regression is pinned.

**Live server.** I ran the confirmation your test plan left unchecked. With `query.timeout = 1` on a `duckdb_arrow` build, an expensive `where` subquery against `GET /api/v1/query/cpu` returned 504 `"Query timed out"` at 1.5s and bumped `arc_query_timeouts_total` from 0 to 1, through the Arrow path. The same query on a pre-fix build from main ran 9.5 seconds past the configured timeout and returned 200 with the counter untouched. Fast queries still return 200 in single-digit milliseconds under the timeout-aware code.

One optional note on the second test: `TestQueryMeasurementNoTimeoutKeepsUserContext` also passes against the pre-fix handler, because a nil-deadline context cannot be told apart from `context.Background()`. If you want it to pin UserContext propagation, inject a marked context in middleware (`app.Use(func(c *fiber.Ctx) error { c.SetUserContext(ctxWithTestValue); return c.Next() })`) and assert the value comes through. Fine to skip.

1. **CLA.** The CLA check is failing; signing is required on every PR. Please sign per the bot's instructions in this thread and comment `recheck`.
2. **Release notes.** Community fixes get an entry in the current planned release file. Please add a short entry near the top of the Bug fixes section in `RELEASE_NOTES_2026.09.2.md` describing the fix, ending with the standard credit line you can copy from any neighboring entry.

xe-nvdk:
@MrBeldum thank you for this fix — the review verified it end to end, including a live 504 A/B against a pre-fix build, and it is ready to merge from our side.

To move it along I have covered the release-notes item from my review myself: your branch now carries the 26.09.2 entry with your credit line, plus a merge with current main so everything is up to date. The only remaining step is yours: please sign the CLA per the bot's instructions in this thread and comment `recheck`. As soon as that check flips green, this merges (and it unblocks #705, which builds on your context wiring).

MrBeldum:

MrBeldum:
recheck

MrBeldum:
@xe-nvdk CLA is signed and the CLA check is green. CI was also green on the tip.

One remaining blocker on our side: the branch is currently conflicting with `main` again (merge state dirty). Working on a clean rebase next — will ping when conflict-free.

MrBeldum:
Refreshed this branch onto current `main` (merge commit) and resolved the `RELEASE_NOTES_2026.09.2.md` conflict.

Kept:
- the `queryTimeout` fix for `/api/v1/query/:measurement` and the release-notes credit for this PR
- main’s `expires_at` UTC storage note

Ready for another look when CI is green.

MrBeldum:
@xe-nvdk Conflicts with `main` are resolved (release notes keep upstream entries plus the #308 measurement-timeout credit). CLA was already green — should be mergeable again once CI finishes.

MrBeldum:
@xe-nvdk One extra fork-CI wrinkle: the latest `CI` run is `action_required` (https://github.com/Basekick-Labs/arc/actions/runs/34612481373) — it needs a maintainer approval to execute from the fork. CLA is green and the branch is mergeable with `main` again.

MrBeldum:
@xe-nvdk Update: fork CI is approved and green now (Build & Test + CLA). Branch is `CLEAN` / mergeable with `main`. The earlier changes-requested items (CLA + release notes) are done — ready to merge from our side whenever you are.

xe-nvdk (approved):
CLA is green and the refresh is clean. Thanks for pushing this through, and for the clear running commentary on the branch state.

What I checked on the updated branch:

**The conflict resolution did not cost anything.** I diffed the PR head against current main: the change is exactly three files, and the release-notes diff is a pure addition. Every heading main carries is still present, with your #308 entry added at the top of Bug fixes, so the entries that landed while this sat (including the `expires_at` UTC note) are intact.

**The fix itself is untouched.** `internal/api/query.go` and the test file are byte-identical to the commit I reviewed on the 4th, so the earlier verification carries over unchanged: cancel released on every return path, the ownership handoff into the Arrow dispatch correct, and the live result of 504 `"Query timed out"` at 1.5s with `arc_query_timeouts_total` incrementing, against a pre-fix build that ran 9.5 seconds past the same configured timeout. Nothing on that execution path moved in main since, so that result still stands.

**Suite health on the merged state.** Full `internal/api` under `-race` with `-tags=duckdb_arrow`, plus build, vet, and gofmt across the repo, all clean.

Merging.

PR #705, opened 2026-09-04
fix(query): enforce query governance on every user-SQL endpoint

### LINKED ISSUE #702: queryMeasurement bypasses query governance (rate limits, quotas, max rows, per-token timeout)
Found while reviewing #701. Enterprise query governance is enforced only in the POST `/api/v1/query` handler: `CheckRateLimit`, `CheckQuota`, and the resulting `governanceMaxRows` / `governanceTimeout` are all derived there (internal/api/query.go, the `localProcessing` block) and passed into the execution paths.

`GET /api/v1/query/:measurement` never consults the governance manager. It enforces RBAC via `checkMeasurementPermission`, and after #701 it applies the global `query.timeout`, but a token with governance limits can still:

- issue unlimited requests through this endpoint (no rate limit or quota accounting),
- pull up to the endpoint's own `limit` cap regardless of the policy's max rows,
- run up to the global timeout even when the policy's `MaxDuration` is shorter.

The endpoint accepts a user-supplied `where` fragment, so these are real queries with real cost, not just point lookups. Authorization is not affected; this is a limits and accounting gap.

Fix shape: hoist the governance block shared by the POST handler into a helper and call it from `queryMeasurement` (and any other user-SQL endpoint that lacks it), wiring `MaxRows` into the SQL limit clamp and `MaxDuration` into the effective timeout the same way POST does.

xe-nvdk:
## Summary

Enterprise query governance (rate limits, quotas, `max_rows_per_query`, `max_scan_duration_sec`) was enforced only on `POST /api/v1/query`. This PR hoists that block into a shared `checkQueryGovernance` helper and enforces it on the three endpoints that were open: `GET /api/v1/query/:measurement`, `POST /api/v1/query/arrow`, and `POST /api/v1/query/estimate` (its `COUNT(*)` wrapper executes the full user subquery, so it carries real scan cost).

- The helper records metrics/logging and sets `Retry-After`; each endpoint renders the 429 in its own error shape (`EstimateResponse` keeps `warning_level` as the severity channel, the arrow endpoint keeps its `fiber.Map` shape, POST keeps msgpack adaptation via `respondError`).
- `max_rows_per_query` caps streamed rows on every row-returning path: Arrow JSON, the database/sql fallback, and the Arrow IPC stream via a batch slice mirroring `drainArrowBatches`, with an early break so DuckDB stops materializing batches at the cap.
- `max_scan_duration_sec` overrides the global `query.timeout` with the same `effectiveTimeout` semantics as POST on all four endpoints.
- The license gate sits behind a package-level test seam (`queryGovernanceLicensed`), same style as the existing `arrowJSONQueryFunc` seam, since `license.Client` carries no injectable license.

**Stacked on #701** (it builds on that PR's timeout context wiring in `queryMeasurement`); this PR sits until #701 merges, then rebases to a single commit. Fixes #702.

## Notes for review

- POST deliberately keeps its pre-existing charge-before-parse ordering; the other three charge after validation and RBAC, so malformed or forbidden requests do not burn quota there (documented on the helper).
- Pre-existing semantics kept deliberately: a policy `MaxDuration` larger than the global timeout raises the effective timeout for that token (matches POST today); worth a separate decision if we ever want min() semantics.
- The experimental arcx Arrow serve path (arcx_engine builds only) is not row-capped yet; documented inline. It does get the policy timeout via ctx.

## Test plan

- [x] 6 new handler tests: rate-limit 429 + Retry-After, quota 429, MaxRows + MaxDuration flow-through, tokenless no-op, and rejection tests for the estimate and arrow endpoints (budget pre-consumed through the manager so no DB is needed)
- [x] `go test -tags=duckdb_arrow -race ./internal/api/ ./internal/governance/` green; vet/gofmt clean
- [x] Live E2E on an Enterprise-licensed dev server: arrow returned exactly 10 of 50 rows under a 10-row policy (verified with an Arrow IPC reader); arrow and estimate both 504 at a 1s policy timeout against a 300s global; one shared rate budget across all four endpoints (burst arithmetic matches `arc_governance_rate_limited_total` exactly); POST still 429s through the shared helper
- [x] Two adversarial review rounds; round 1 caught the arrow/estimate coverage gap, round 2 verified the IPC slice ownership on every exit path and found nothing above LOW

### DIFF
--- a/RELEASE_NOTES_2026.09.2.md
+++ b/RELEASE_NOTES_2026.09.2.md
@@ -182,6 +182,21 @@ is published. Responsibly reported by **[@rexpository](https://github.com/rexpos
 
 ## Bug fixes
 
+### Every query endpoint now enforces query governance ([#702](https://github.com/Basekick-Labs/arc/issues/702))
+
+Enterprise query governance was enforced only on `POST /api/v1/query`, so
+a token with governance limits could bypass rate limits, quotas, the
+per-query row cap, and the per-token execution timeout through the three
+other user-SQL endpoints: `GET /api/v1/query/:measurement`,
+`POST /api/v1/query/arrow`, and `POST /api/v1/query/estimate` (whose
+`COUNT(*)` wrapper executes the full subquery, so it carries real scan
+cost). All four now share one enforcement path: rejected requests get 429
+(with `Retry-After` for rate limits), the policy's `max_rows_per_query`
+caps streamed rows on the row-returning endpoints (JSON, Arrow IPC, and
+the database/sql fallback), and `max_scan_duration_sec` overrides the
+global `query.timeout` everywhere. RBAC was never affected; this closes a
+limits and accounting gap, not an authorization hole.
+
 ### The measurement endpoint now honors the configured query timeout ([#308](https://github.com/Basekick-Labs/arc/issues/308))
 
 `GET /api/v1/query/:measurement` executed against `context.Background()` with
--- a/internal/api/query.go
+++ b/internal/api/query.go
@@ -121,6 +121,15 @@ var (
 // Returns (rowCount, handled). If handled is false, the caller falls back to database/sql.
 var arrowJSONQueryFunc func(h *QueryHandler, c *fiber.Ctx, ctx context.Context, cancel context.CancelFunc, convertedSQL string, profileMode bool, governanceMaxRows int, start time.Time, timestamp string, onComplete func(int), onFail func(string), onTimeout func()) (int, bool)
 
+// queryGovernanceLicensed reports whether the Enterprise query-governance
+// gate is open for this handler. Package-level indirection (same seam style
+// as arrowJSONQueryFunc) because license.Client carries no injectable
+// license, so handler tests could otherwise never exercise the licensed
+// path of enforceQueryGovernance.
+var queryGovernanceLicensed = func(h *QueryHandler) bool {
+	return h.governanceManager != nil && h.licenseClient != nil && h.licenseClient.CanUseQueryGovernance()
+}
+
 // arrowMsgPackQueryFunc is set by query_msgpack.go init() when compiled with duckdb_arrow tag.
 // It executes a query via DuckDB's native Arrow API and streams a MessagePack response.
 // Returns (rowCount, handled). If handled is false, the caller MUST treat the request as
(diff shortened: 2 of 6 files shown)


xe-nvdk:
> Thanks for the contribution!
> 
> Before this can be merged, please sign the [Contributor License Agreement](https://github.com/Basekick-Labs/arc/blob/main/CLA.md). Arc is AGPL-3.0 and Basekick Labs also ships commercially licensed builds; the CLA is what lets your contribution be included in both. You keep full ownership of your work.
> 
> To sign, reply to this thread with exactly:
> 
> I have read the CLA Document and I hereby sign the CLA
> 
> You can retrigger this bot by commenting **recheck** in this Pull Request. Posted by the **CLA Assistant Lite bot**.

xe-nvdk:
recheck

xe-nvdk:
Rebased onto current main now that #701 has merged. This also explains the CLA failure: the check was never asking the PR author to sign (xe-nvdk is on the workflow allowlist). It was blocking because the branch still carried MrBeldum's commit 2cd71e321, the pre-squash version of #701, which was unsigned when that check last ran. The rebase drops that commit (its content is in main via d5b38ca6e), so the PR is now a single commit from an allowlisted author.

Two conflicts came out of the rebase, both resolved:

- `internal/api/query_arrow.go`: #706 extracted the IPC batch loop into `streamArrowIPC`, so instead of keeping my inline batch slice I moved the governance row cap into that helper as a `maxRows` parameter. That is a better home for it: the cap now sits beside the ownership logic it has to respect, and the helper's existing leak-checked tests cover it. The four existing call sites in `query_arrow_test.go` pass 0 (no cap). Slice release is wired on all three exit paths (cast error, dictionary-encode error, and after Write, before the write and flush error checks), plus the early break at the cap so DuckDB stops materializing batches.
- `RELEASE_NOTES_2026.09.2.md`: kept both sides. I verified every heading main carries is still present, with the #702 entry added.

Added two tests to the `streamArrowIPC` family: `TestStreamArrowIPCGovernanceRowCap` (cap mid-batch, asserts the third batch is never consumed and decodes the emitted stream back to confirm the row count) and `TestStreamArrowIPCGovernanceCapExactBoundary` (cap landing exactly on a batch boundary). I mutation-tested the new release path by deleting it, and the checked allocator fails with the expected leak.

Re-verified live on an Enterprise-licensed server, since the cap changed hands: with a 10-row policy against 50 stored rows, the Arrow IPC endpoint returns exactly 10 on the plain, dictionary-encoded, and zstd-compressed paths, and the JSON measurement endpoint returns 10. The 1-second policy `MaxDuration` still overrides the 300-second global on both the arrow and estimate endpoints (504 at 1.5s), and the shared rate budget still trips across endpoints with matching metrics.

Full `internal/api` suite green under `-race` with `-tags=duckdb_arrow`; build, vet, and gofmt clean.

xe-nvdk:
Correction to my previous comment, and a push that fixes it.

An adversarial review pass caught that the two row-cap tests I described were never actually in the commit. I appended them to the test file after finishing the rebase, then pushed, and a push carries the commit rather than the working tree, so they stayed as an unstaged local modification. The claim that the rehomed cap had test coverage was wrong: as pushed, deleting the slice, its release, or the early break all passed the committed suite. That is fixed in 643f9b2, where the tests are actually committed (the test file diff is 138 lines now, not 8).

While correcting that I also acted on the rest of the pass:

- Both new tests now release the reader and the IPC reader through `defer` rather than after the assertions. Previously a failing assertion would `Goexit` before the release, so the deferred allocator check fired on the way out and buried the real diagnostic under a leak report and an `AssertSize` panic. A cap regression now reports as a cap regression.
- The row-cap test decodes the emitted stream and checks every value rather than only the row count, so a slice-offset bug cannot pass.
- The exact-boundary test now asserts `reader.idx` and the emitted batch count. Its comment previously claimed it pinned "must not slice", which it did not and should not: a full-length slice at the boundary emits identical bytes and is released on the same path, making it a spare allocation rather than a behavior worth asserting. The comment now says what the test actually pins.
- Marked the top-of-loop `remaining <= 0` guard as defensive, since the post-flush break makes it unreachable today and it is the fallback if that break is ever removed.
- Added the missing blank line between the #702 and #308 release-notes sections.

Re-verified by mutation against the committed tests: removing the slice release, removing the early break, and flipping the cap comparison off by one are each caught now.

Everything else in the pass held up, including record ownership on all exit paths (no double-release, since the cast and dictionary transforms always return fresh records), `NewSlice` parent semantics for the cdata-imported DuckDB records, cap arithmetic across the edge cases, the loop tail and caller cleanup on the early break, call-site plumbing, and rebase integrity against #701 and #706.

One pre-existing issue surfaced that is not from this change and is not in scope here: a panic inside the IPC stream writer is swallowed by the closure's recover, so `reader.Release()`, `conn.Close()` and `cancel()` never run, leaking the DuckDB result, the pooled connection and the timer. It is identical on main. I will file it separately.

Issue #546, opened 2026-07-21
low: value time.Time API fields — omitempty is a no-op and some render in local time, not UTC (codebase-wide)


xe-nvdk:
## Context

Two latent defects share the same set of struct fields — value `time.Time` fields on JSON API responses:

**(1) `omitempty` is a no-op on a value `time.Time`.** `encoding/json` never omits a `time.Time` struct for `omitempty` (a struct is not an "empty" kind, and `time.Time` implements `json.Marshaler`), so the zero value always marshals as `"0001-01-01T00:00:00Z"` instead of being omitted. The tag misleadingly implies the field is optional.

**(2) Some are stamped in local time, not UTC.** Arc's convention is UTC everywhere. The MQTT stats timestamps are set with a bare `time.Now()` (and reconstructed via `time.Unix`, which returns local), so `connected_since` / `last_message_at` render with the server's local offset — inconsistent with every other timestamp in the API.

Surfaced during #328 (MQTT `lastMessageAt` mutex→atomic). Left out of scope there; investigated codebase-wide here.

## The uniform rule (matches the already-correct sites)

- A value `time.Time` field **never** carries `omitempty` (it does nothing).
- A genuinely-**optional** timestamp (a meaningful "unset" state) is `*time.Time`, set only when a real value exists — then `omitempty` actually omits it.
- Every Arc-stamped timestamp is **UTC** (`time.Now().UTC()`, and reconstruct with `.UTC()`).

Reference for the pattern: `internal/queryregistry/registry.go`, `internal/tiering/tier.go` (×4), `internal/backup/manifest.go`, and auth's own `TokenInfo.ExpiresAt` — all already `*time.Time`.

## Sites + treatment

| Site | Fix |
|---|---|
| `internal/mqtt/subscription.go:135` `LastMessageAt` | → `*time.Time`, **UTC** (set only when a message was received) |
| `internal/mqtt/subscription.go:136` `ConnectedSince` | → `*time.Time`, **UTC** (set only when connected) |
| `internal/auth/auth.go:67` `LastUsedAt` | → `*time.Time` (already UTC from SQLite; matches sibling `ExpiresAt`) |
| `internal/license/client.go:71` `VerifyResponse.ExpiresAt` | drop `,omitempty` (decode-side/debug-only; authoritative signed struct already uses plain tag) |
| `internal/license/client.go:131` `ActivateResponse.ExpiresAt` | drop `,omitempty` (same) |
[...]

Issue #460, opened 2026-05-25
audit: normalise timestamps to UTC across auth + cluster (follow-up to PR #459)


xe-nvdk:
## Background

PR #459 (Phase A.2 Item 2) fixed a UTC drift in the RBAC FSM apply path: `nextProposerTimestamp()` on the proposer side and every `time.Unix(0, entry.{Created,Updated}AtUnixNano)` in `internal/auth/cluster_rbac_apply.go` now use `.UTC()` so SQLite text-encoded `created_at` values round-trip byte-identically under the `WHERE created_at = ?` readback in `CreateRole` / `CreateMeasurementPermission`.

That fix was narrowly scoped to RBAC. The same shape exists elsewhere in the codebase and should be audited.

### Confirmed-needs-audit sites

The Phase A token applier path uses the same `time.Unix(0, ...)` shape without `.UTC()`:

- `internal/auth/cluster_apply.go:132` — `createdAt := time.Unix(0, entry.CreatedAtUnixNano)`
- `internal/auth/cluster_apply.go:135` — `expiresAt = time.Unix(0, entry.ExpiresAtUnixNano)`
- `internal/auth/cluster_apply.go:204` — `expiresAt = time.Unix(0, entry.ExpiresAtUnixNano)`

There is no current `WHERE created_at = ?` or `WHERE expires_at = ?` equality readback paired with these inserts, so it doesn't round-trip-break today — but it's the same latent bug shape that bit PR #459, and any future readback added against these columns would inherit the timezone-mismatch risk.

### Broader audit (lower priority)

Walk the codebase for proposer/applier-side timestamp pairs that pass through SQLite text serialisation:

1. Grep for `time.Unix(0,` and `time.Now()` in `internal/auth/`, `internal/cluster/raft/`, `internal/database/`, `internal/audit/`, and anywhere else timestamps land in SQLite.
2. For each site that inserts a `time.Time` into SQLite, confirm whether any read path uses `WHERE <col> = ?` equality. If yes, both sides must be UTC.
3. For sites that only insert and never equality-match, normalise to UTC anyway for log/metric/diagnostic consistency.

## Suggested approach

- Add `.UTC()` to all `time.Unix(0, ...)` sites in `internal/auth/cluster_apply.go` (3 sites confirmed above) as a no-behaviour-change defensive normalisation.
- Run a broader grep + reviewer-agent sweep to enumerate every timestamp insert/readback pair across the auth, audit, and cluster packages.
[...]

PR #712, opened 2026-09-10
fix(auth): enforce token expiry on cache hits

xe-nvdk:
Closes a token-expiry gap in `AuthManager.VerifyToken`, plus a timezone inconsistency found while verifying it on a live binary.

Detail accompanies the forthcoming security advisory. Reported by @rexpository.

## Summary

- **Expiry is now enforced on the cache-hit path.** `VerifyToken` memoizes a successful lookup for up to `auth.cache_ttl` (default 300s). The cache-hit branch checked only that entry deadline, never the token's own `expires_at`, so a token that expired *while cached* kept authorizing requests until the entry aged out. The same token was correctly rejected whenever the lookup reached SQLite, so authorization depended on cache state rather than on the credential. An entry whose token has expired is now evicted and re-validated against the database, which rejects it.
- **Cache entries are never held past the token's own expiry** (defense in depth), so `cleanupExpiredCache` and the size-eviction scan — both of which order by the cache deadline — cannot prefer entries that can no longer authorize anything.
- **`expires_at` now stored in UTC** on the standalone create and update paths. `created_at` comes from SQLite's `CURRENT_TIMESTAMP` (always UTC) while `expires_at` was text-encoded in the caller's own location, so a non-UTC deployment stored `2026-09-10 15:32:45.958903-06:00` beside `2026-09-10 21:32:41` in the same table. The clustered apply path already normalized ([#459](https://github.com/Basekick-Labs/arc/pull/459)/[#460](https://github.com/Basekick-Labs/arc/issues/460)); this matches it and the "Arc-stamped timestamps are UTC" rule from [#546](https://github.com/Basekick-Labs/arc/issues/546).

**Scope:** revoke, delete, update and rotate all call `InvalidateCache()` and were never affected — only passive expiry was. No token was ever accepted or rejected incorrectly by the timezone issue: expiry is compared as parsed instants in Go, never as strings, and there is no SQL-level comparison on the column. Existing rows are left as they are.

## Configuration matrix

|---|---|---|
| `auth.enabled=false` | No — `authManager == nil`, middleware calls `c.Next()` (`middleware.go:114`) | n/a |
| Auth on, `auth.cache_ttl=300` (default) | Yes — cache-hit branch `auth.go:855` | `entry.info` non-nil by construction (`auth.go:983`) |
| **Auth on, `auth.cache_ttl=600` (NON-DEFAULT)** | Yes — exercised in the live binary run below; startup log confirms `"cache_ttl":600` | 200 while valid, 401 after expiry |
| `auth.cache_ttl=0` / negative (non-default) | No hit ever (`now.Before(now)` is false) → DB lookup each time | No panic; `cleanupLoop` clamps its ticker to 10s (`auth.go:468`) |
| `auth.max_cache_size` small (non-default) | Yes — victim is the soonest `expiresAt`, now capped at token expiry | Expiring tokens evicted first; harmless |
| Non-expiring token (`ExpiresAt == nil`) | Nil-guarded before deref (`auth.go:856`) | Full cache TTL preserved (test) |
| Token already expired at first verify | No — DB path rejects before caching (`auth.go:915`) | Unchanged |
| Cluster mode, FSM-applied tokens | Same `VerifyToken`; those writes already UTC (`cluster_apply.go:159`) | `InvalidateCache` on apply unchanged |
| `GET /api/v1/auth/verify` | Yes — second caller of `VerifyToken` | Live-verified 401 after expiry |

## Test plan

- [x] **Pre-fix proof (revert-run-restore):** with `auth.go` reverted to `HEAD`, the four new regression tests FAIL for the right reasons — `200` instead of `401`, uncapped cache deadline, no eviction, offset-bearing `expires_at`. Two negative controls pass in both states.
- [x] `go test -race ./internal/auth/...` — **ok, 179s**, full package
- [x] `gofmt -l ./internal/auth` empty; `go vet ./internal/auth/...` clean
- [x] **Live binary** (`go build -tags=duckdb_arrow`), `auth.cache_ttl=600`, token created via `POST /api/v1/auth/tokens` with `expires_in`: `200/200/200` while valid → `401/401` after expiry on both `/api/v1/databases` and `/api/v1/auth/verify`. With the entry still well inside its 600s TTL, unfixed code would have returned 200 for ~10 more minutes.

New tests: `TestVerifyToken_CachedTokenRejectedAfterExpiry` (deterministic), `TestVerifyToken_ExpiresWhileCached_EndToEnd` (real timing), `TestVerifyToken_CacheDeadlineCappedAtTokenExpiry`, `TestCreateToken_ExpiresAtStoredInUTC`, `TestMiddleware_CachedExpiredToken`, plus negative controls for still-valid and non-expiring tokens.

### DIFF
--- a/RELEASE_NOTES_2026.09.2.md
+++ b/RELEASE_NOTES_2026.09.2.md
@@ -82,7 +82,45 @@ and errgroup primitives used by tiering and Iceberg. The auth, cluster-security,
 tiering, and Iceberg suites were verified against the new versions, the latter
 two under `-race`.
 
+### Expired API tokens are now rejected on cache hits
+
+Arc caches successful token verifications in memory for `auth.cache_ttl`
+seconds (default 300) so that ingestion does not pay a SQLite lookup per
+request. That cache checked only its own entry deadline, not the token's
+`expires_at`, so a token that expired *while cached* kept authorizing requests
+until the entry aged out — up to one cache TTL past the expiry an operator
+configured. The same token was correctly rejected whenever the lookup reached
+the database, so the behaviour depended on cache state rather than on the
+credential.
+
+Token expiry is now enforced on every request regardless of cache state: an
+entry whose token has expired is evicted and re-validated against the database,
+which rejects it. Cache entries are additionally never held past the token's own
+expiry. Revoking, deleting, updating, or rotating a token already invalidated
+the cache immediately and was never affected; only passive expiry was.
+
+Full technical detail will accompany the corresponding security advisory once it
+is published. Responsibly reported by **[@rexpository](https://github.com/rexpository)**.
+
 ## Bug fixes
+
+### Token `expires_at` is now stored in UTC
+
+`api_tokens.created_at` is filled by SQLite's `CURRENT_TIMESTAMP` and is always
+UTC, but `expires_at` was bound as a Go `time.Time` and text-encoded using the
+caller's own location — so an Arc running in a non-UTC zone stored
+`2026-09-10 15:32:45.958903-06:00` in one column and `2026-09-10 21:32:41` in
+the other. Both denote the same instant and expiry enforcement compares parsed
+values in Go, so no token was ever accepted or rejected incorrectly; the effect
+was two timezone domains in one table and the two columns rendering in
+different zones over the token API.
+
+Writes now normalize to UTC on both the create and update paths, matching the
+clustered apply path, which already did this ([#459](https://github.com/Basekick-Labs/arc/pull/459) /
+[#460](https://github.com/Basekick-Labs/arc/issues/460)), and the "Arc-stamped timestamps are
+UTC" rule from [#546](https://github.com/Basekick-Labs/arc/issues/546). Existing rows are left as
+they are — they parse back correctly and are compared as instants, never as
+strings.
 
 ### Arrow IPC streaming has direct disconnect regression coverage ([#425](https://github.com/Basekick-Labs/arc/issues/425))
 
--- a/RELEASE_NOTES_2026.09.2.md
+++ b/RELEASE_NOTES_2026.09.2.md
@@ -106,21 +106,24 @@ is published. Responsibly reported by **[@rexpository](https://github.com/rexpos
 
 ### Token `expires_at` is now stored in UTC
 
-`api_tokens.created_at` is filled by SQLite's `CURRENT_TIMESTAMP` and is always
-UTC, but `expires_at` was bound as a Go `time.Time` and text-encoded using the
-caller's own location — so an Arc running in a non-UTC zone stored
-`2026-09-10 15:32:45.958903-06:00` in one column and `2026-09-10 21:32:41` in
-the other. Both denote the same instant and expiry enforcement compares parsed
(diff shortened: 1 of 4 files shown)


xe-nvdk:
## Review pass complete

Two reviewers per CLAUDE.md: one deep adversarial, one Security Checklist. Both independently ran revert-run-restore proofs and probed go-sqlite3 directly rather than reasoning from assumption.

**Security Checklist reviewer — 0 Blockers / 0 High / 0 Medium.** Verdict: *"the expiry bypass is fully closed."* It verified the migration-safety claim empirically (a throwaway go-sqlite3 module): legacy offset-bearing rows parse to the identical instant post-change, so **no existing token's lifetime shifts**. It also confirmed fail-closed behaviour on a DB error, no metrics double-count, no DoS amplification (at most one extra query per expired-token request, and the deadline cap prevents an expired entry from ever being re-created), and no new logging of secrets.

**Deep reviewer — 0 Blockers / 2 High / 1 Medium**, all on the UTC half; the security fix itself was confirmed correct, with sound concurrency and genuine pre-fix-failing tests. All three findings are fixed in 0047a21:

- **H1 — the UTC test passed for the wrong reason.** It scanned `expires_at` into a Go `string`, which is the driver's RFC3339 re-rendering, not the stored bytes — so it was asserting on go-sqlite3's output rather than on the write. Now asserts the parsed value's zone offset is zero (encoding-independent) plus pins raw bytes via `CAST(... AS TEXT)`. Re-verified failing against unfixed code on the parsed offset (`-21600 seconds`) under both `TZ=UTC` and `TZ=Europe/Berlin`.
- **H2 — the storage-format claims were wrong, and I had overclaimed the fix.** Probed the actual bytes:

  ```
  created_at (CURRENT_TIMESTAMP)  "2026-09-10 22:09:40"
  expires_at (UTC time.Time)      "2026-09-10 21:32:45.958903+00:00"
  expires_at (UTC-6 time.Time)    "2026-09-10 15:32:45.958903-06:00"
  ```

  A correct UTC write carries an explicit `+00:00`, not a bare `Z`, and the column still differs textually from `created_at` after the fix. What normalizing removes is the offset *variance* — two rows holding the same instant no longer sort differently as text depending on which node wrote them. Comment and release note corrected; both now state that `expires_at` must still be compared as a parsed instant, never as text.
- **M1 — a test comment claimed coverage it does not provide.** `TestVerifyToken_ExpiresWhileCached_EndToEnd` takes a natural cache miss by the time it re-verifies (the capped deadline has passed), so it proves the deadline cap, not the cache-hit branch. Confirmed by instrumenting it. Comment corrected to point at the two tests that do cover that branch.

Worth recording from the security review: at the SQL string level the two formats genuinely diverge — its probe showed `WHERE expires_at > ?` wrongly excluding a same-instant offset-bearing row. Arc performs **no** SQL comparison on this column anywhere today (every read is `SELECT`-then-compare-in-Go), so it was unexploitable; the fix closes it before someone adds such a query.

**Re-verified after the fixes:** `go test -race ./internal/auth/...` ok (181s); all four security regression tests still fail against unfixed `auth.go`; `gofmt`/`go vet` clean.

PR #713, opened 2026-09-11
fix(cluster): authenticate every coordinator handshake field and response

xe-nvdk:
Detail accompanies the forthcoming security advisory. Reported by @rexpository.

> **BREAKING for clustered Enterprise deployments.** The handshake wire format changed: a node on this version cannot authenticate with an older one. Upgrade as a coordinated restart — **stop all nodes, upgrade all, restart all**. See the release notes for the full mixed-version behaviour. Single-node, non-clustered and OSS deployments are unaffected.

## Summary

The join, heartbeat and leave messages authenticated only `{message type, nonce, node ID, cluster name, timestamp}` while their handlers went on to consume other fields from the same message — a joining node's role and its advertised Raft/API/coordinator addresses, a heartbeat's self-reported state. With `cluster.tls_enabled` unset (the default), a peer able to modify inter-node traffic could rewrite any of those and the signature still verified. The three responses (join result, leader redirect, forward-apply ack) carried no authentication at all.

**Every non-auth field of every handshake message is now covered by its MAC, in both directions.** Responses are signed over the request's nonce, so a requester validates with the nonce it generated and needs no replay cache of its own. The rule is deliberately "the whole message" rather than "the fields a handler reads today" — the latter is what left the next added field unprotected.

Also included:

- **Replay protection** on join/heartbeat/leave via fail-closed `Validate*HMACWithReplay` wrappers (reusing the existing `edgesync_auth.go` pattern rather than hand-rolling a third copy). The nonce is consumed only after the MAC verifies, so a forged message cannot burn a legitimate nonce's cache slot.
- **Nonce cache lifetime corrected** to `2T+1s`, derived from the tolerance rather than equal to it. A message may be stamped up to one tolerance in the *future*, so a `T`-lived entry could be evicted while its MAC was still acceptable. Fixed in the constructor, so cache-invalidate and edge-sync are tightened too.
- **Uniform pre-auth join errors.** The cluster-name mismatch previously echoed the expected name back to an unauthenticated caller, and the three failure modes were individually distinguishable. Causes remain in the server log.
- **Length-prefixed canonical encoding** instead of NUL-delimited: this transport is raw TCP carrying JSON and passes NUL through, so a delimiter-joined payload could be re-partitioned by an attacker who controls a field's contents.
- **Per-destination heartbeat/leave signing** on a per-peer copy. Registry entries can share an address (node IDs are hostname+PID, so a bare-metal restart leaves the old entry), so one nonce per tick would be rejected as a replay every tick; the shared struct was also marshalled concurrently by N goroutines.

### A bug found by the new tests during implementation

`c.nonceCache` is a typed nil inside the `ReplayGuard` interface, so `guard == nil` was **false** and `Track` panicked on the nil receiver. A handshake arriving at a Coordinator whose cache had not been installed would have **crashed the node rather than rejecting the message**. Fixed at both layers (`Track` is nil-safe; the wrappers reject a typed-nil guard) with a regression test.

## Configuration matrix

|---|---|---|
| OSS standalone, `cluster.enabled=false` | No — coordinator never constructed (`main.go:1420` gate) | **Binary-run verified**: ready, no panics, `/health` 200 |
| Cluster + Raft, `raft_advertise_addr` **set** | Yes — all senders/validators | **Binary-run verified**: join OK, signed role/addr registered |
| Cluster + Raft, `raft_advertise_addr` **UNSET (default)** | Yes — fallback must precede signing (`coordinator.go`) | **Binary-run verified**: join succeeds. No unit test covers this; tests build `Coordinator` literals |
| Cluster, no Raft | Yes — registry path | `registry`/`localNode` set in `NewCoordinator`; `nonceCache` at `Start()` before `acceptLoop` |
| `nonceCache == nil` + secret set | Reject, fail-closed | Tests only in production (cache installed before listener accepts); typed-nil covered |
| `shared_secret` empty | Auth + replay + response validation skipped | Unchanged; unsupported since #505 |
| Two registry entries sharing an address | Per-destination nonces | Distinct nonces asserted; `-race` clean |
| `cluster.tls_enabled` / `server.tls_enabled` | MAC path identical | Independent flags |
| `licenseClient` nil vs set | `validateCoreLimitForJoin` unchanged | `CoreCount` now signed, closing a forged-count bypass |
| Mixed-version join/heartbeat/leave | Both directions reject | Documented in the upgrade note |
| Mixed-version forward-apply (new follower → old leader) | Old leader applies, returns unsigned ack, follower rejects | **Write applied but reported failed** — documented |
| Mixed-version graceful follower restart under old leader | Removed from Raft by its own leave, cannot rejoin | Documented; worst rolling-restart consequence |

## Test plan

- [x] `go test -race ./internal/cluster/... ./internal/api/...` — **all 7 packages ok** (cluster 22s, api 62s)
- [x] `gofmt -l` clean on every touched file; `go vet` clean
- [x] **Live licensed 2-node cluster**, both `raft_advertise_addr` set and unset:
  - Join succeeds; `role=reader`, `address=127.0.0.1:9102` registered — the **signed** values
  - 60s steady-state heartbeats: **0** auth failures, **0** replay warnings
  - Graceful leave: `peers_notified: 1`, accepted, removed from FSM + registry
  - **Replay**: 1st `ACCEPTED`, byte-identical 2nd `REJECTED: authentication failed`; server logged `replay: true` with `security: handshake nonce already used`
  - **Error oracle**: wrong cluster name / missing MAC / bad MAC / wrong secret all return the identical `"authentication failed"`; cluster name never leaves the server
- [x] Pre-fix evidence: the reporter's Appendix A reproducer was run on `main` earlier in this work and demonstrated the violation. The new tests use the new API and cannot compile against `main`, so no revert-run-restore is claimed for them.

New tests: per-field mutation tables for all six message types (including `Nodes` add/remove/reorder), canonical-encoding collision, nonce binding, replay + fail-closed + typed-nil guard, live-shaped join/leave/heartbeat handler tests, joiner-side response validation, `checkForwardAck` forged-success **and forged-error-code** cases, and per-destination heartbeat signing under `-race`.

### DIFF
--- a/RELEASE_NOTES_2026.09.2.md
+++ b/RELEASE_NOTES_2026.09.2.md
@@ -82,6 +82,72 @@ and errgroup primitives used by tiering and Iceberg. The auth, cluster-security,
 tiering, and Iceberg suites were verified against the new versions, the latter
 two under `-race`.
 
+### Cluster hardening: fully authenticated coordinator handshake (Enterprise)
+
+Every field of every coordinator-handshake message is now covered by its
+HMAC, in both directions. Previously the join, heartbeat and leave messages
+authenticated only `{message type, nonce, node ID, cluster name, timestamp}`
+while the handlers went on to consume other fields from the same message — a
+joining node's role and its advertised Raft, API and coordinator addresses; a
+heartbeat's self-reported state. On a cluster with `cluster.tls_enabled`
+unset (the default), an attacker positioned on the inter-node network could
+rewrite any of those and the signature still verified. The responses —
+join result, leader redirect, and the forward-apply acknowledgement — carried
+no authentication at all, so the same position allowed feeding a joining node
+a fabricated cluster membership list, redirecting it to an attacker-chosen
+coordinator, or telling a follower its replicated write had failed (or
+succeeded) when it had not.
+
+Three further changes close the same class of gap:
+
+- **Replay protection** on join, heartbeat and leave. These messages were
+  previously bounded only by the five-minute freshness window, inside which a
+  captured message could be resent verbatim. Nonces are now consumed on
+  receipt. The nonce cache's lifetime was also corrected — it must outlive the
+  freshness window by more than the window itself, because a message may be
+  stamped up to one tolerance in the future — which likewise tightens the
+  existing cache-invalidate and edge-sync replay guards.
+- **Uniform pre-authentication join errors.** A rejection that happens before
+  the peer proves it holds the shared secret now returns one opaque string;
+  previously the cluster-name mismatch echoed the expected cluster name back
+  to an unauthenticated caller and the three failure modes were individually
+  distinguishable. The specific cause is still logged server-side.
+- **Canonical encoding.** The signed payload is length-prefixed rather than
+  NUL-delimited. The coordinator protocol is raw TCP carrying JSON, which
+  passes NUL through, so a delimiter-joined encoding let an attacker who
+  controlled a field's contents re-partition the signed input.
+
+Clustering is an Enterprise feature and is off by default; single-node and
+OSS deployments are unaffected.
+
+> **Upgrade note (clustered Enterprise deployments):** the handshake wire
+> format has changed, so a node on this version cannot authenticate with a
+> node on an older one. Upgrade the cluster as a coordinated restart: **stop
+> all cluster nodes, upgrade the binary on every node, then restart all
+> nodes.**
+>
+> A rolling restart does not degrade gracefully. Across a mixed-version
+> window: cross-version nodes fail each other's heartbeats and are marked
+> unhealthy within roughly three health-check intervals (~15s by default),
+> withdrawing them from query routing; with `cluster.failover_enabled` set, an
+> automatic writer or compactor failover is committed through Raft about 30
+> seconds in, while the original primary is still alive, and that promotion
+> persists after the upgrade completes. A follower that is restarted
+> *gracefully* under a not-yet-upgraded leader is removed from the Raft
+> configuration by its own leave notification and cannot rejoin until the
+> leader is upgraded. Writes forwarded from an upgraded follower to an
+> older leader are applied by that leader but reported to the follower as
+> failed, because the acknowledgement is unsigned — file registrations are
+> retried by anti-entropy, compaction retries on its next tick, and an
+> upgraded node may log `Failed to create initial admin token` at startup
+> (non-fatal).
+>
+> The handshake authenticates inter-node messages; for confidentiality on the
+> interconnect, enable `cluster.tls_enabled`.
+
+Full technical detail will accompany the corresponding security advisory once
+it is published. Responsibly reported by **[@rexpository](https://github.com/rexpository)**.
+
 ### Expired API tokens are now rejected on cache hits
 
 Arc caches successful token verifications in memory for `auth.cache_ttl`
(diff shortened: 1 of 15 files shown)


xe-nvdk:
## Review pass complete

Two reviewers per CLAUDE.md — one deep adversarial, one Security Checklist. Both ran independent verification rather than reasoning from the diff.

**Security Checklist reviewer — 0 Blockers / 0 High / 3 Medium.** Verdict: *"The forgery class is closed. The replay class is closed for the handshake family specifically."* It did not take the canonicalization on faith:

- **Brute-forced collisions**: 1,884 tuples over an adversarial alphabet (`""`, `":"`, `"0:"`, `"12:abc"`, arity 1–3) → **zero**. Separately probed the variable-length `Nodes` slice (2,451 canonical strings) → **zero**.
- **Cross-family replay excluded structurally**, not assumed: handshake input always matches `^[0-9]+:`, while `cache-invalidate`/`replicate-sync`/`replicate-checkpoint`/`raft-auth` begin with a letter. For the two fetch/forward families that start with an attacker-chosen nonce, it confirmed NUL *does* reach handshake fields and then showed length-prefixing makes it inert.
- **The `2T+1s` bound is exactly tight** — it modelled `T` (unsafe), `2T` (sub-second window), `2T+1s` (safe, zero margin).
- **The pre-auth signed error is not an oracle**: every other field of an error response is responder-chosen, and the output is unusable outside a `join-resp` verifier expecting that exact nonce.

**Deep reviewer — 0 Blockers / 0 High / 3 Medium.** Verified field-source symmetry for all six message types (every struct field maps 1:1 to its canonical list), confirmed the `RaftAddr` fallback sits above the signing block with no post-signing mutation on any path, and checked all four mixed-version release-note claims against config defaults — `~15s` = `health_check_interval` 5 × `UnhealthyThreshold` 3, `~30s` = `failover_timeout` 30 — finding nothing overstated.

### Findings addressed in 26a2743

| Finding | Disposition |
|---|---|
| Sec M1 — two fail-open replay guards | **Fixed** — now fail closed + regression test |
| Deep M1 — §16.14.3 wrong at source | **Fixed** at the source, not annotated |
| Deep M3 / Sec M2 — nonce TTL comment | **Fixed** — drift on the changed line |
| Sec M2b — `nonceCache` field comment | **Fixed** — said "future work" for what this PR does |
| Sec M3 — `handleHeartbeat` comment | **Fixed** — claimed join/leave don't replay-check |
| Sec Style4 — freshness guard parity | **Fixed** |
| Deep Style4 — overstated typed-nil claim | **Fixed** — narrowed to what the code guarantees |
| Deep M2 — `SupportsBinaryEntries` unsigned | **Deferred → #714** |

**On the one deferral:** the reviewer is right that its stated mixed-version rationale is obsolete now that the handshake is a hard cutover. It stays unsigned here because it is a different message on a different validator, and widening a wire-break inside a security fix is its own hazard. The obsolete reasoning is corrected in the comment and #714 carries the impact analysis, with a note to bundle it with the next release that already requires a coordinated restart.

**Sec M1 was the only behavioral change**, so it got treated as a new matrix cell rather than a comment fix: a fail-closed regression test **plus a re-run of the live two-node cluster**, since it touches the real forward-apply and replicate-sync paths — join succeeds, zero auth failures, zero replay warnings, zero panics.

**Re-verified after the fixes:** `go test -race ./internal/cluster/... ./internal/api/...` — all 7 packages ok; `gofmt`/`vet` clean on every touched file.

### Matrix cell to inherit

The deep reviewer named one the current matrix does not: **forward-apply ack on a pooled/reused connection**. `handleForwardApplyLoop` serves multiple requests per connection while the client holds `forwardMu` across send+receive, so each ack is bound to its own request's nonce. Correct as built, but it is exactly the multi-request-per-connection interaction the matrix exists to surface — worth carrying into the next cluster PR.

Issue #698, opened 2026-09-03
Replication cannot send WAL entries between ~75MB and 100MB: JSON+base64 envelope exceeds the message cap


xe-nvdk:
## Problem

A WAL entry can be valid for the WAL but unsendable by replication. The WAL accepts payloads up to `MaxWALPayloadSize` (100MB, `internal/wal/wal.go`), and the replication protocol also caps a message at 100MB (`MaxMessageSize`, `internal/cluster/replication/protocol.go:172`). But `WriteMessage` JSON-marshals the frame (`protocol.go:180`), and `ReplicationEntry.Payload` is a `[]byte`, so encoding/json base64-encodes it with a 4/3 expansion. Any entry whose payload exceeds roughly 75MB therefore produces a frame over 100MB, and `WriteMessage` rejects it with `message too large` (`protocol.go:187-188`) even though the WAL stored it happily.

## Consequence

The send error propagates out of the entry write path in `internal/cluster/replication/sender.go`, tearing down that reader's stream. On reconnect the stream resumes from the last acknowledged sequence and hits the same entry again, so an entry in the 75 to 100MB band is a deterministic poison pill: replication to that follower stalls behind it and never makes progress, while the writer side looks like a reconnect loop.

## Scope

Pre-existing, not introduced by #696. The #677 chunking work targets a 32MB chunk size precisely so normal chunks clear this envelope comfortably, but two paths can still legally produce a 75 to 100MB entry:

- a single ingest payload in that band was never split (it is under the WAL cap), exactly as before #696
- a chunk can overshoot the 32MB target when one large element lands on a chunk boundary (the hard check only enforces the 100MB WAL cap)

Surfaced while reviewing #696; see the review thread there for the measurements.

## Options

1. Bound the effective single-entry limit by what replication can actually carry: reject or split at roughly `MaxMessageSize * 3/4` minus JSON overhead when a replication hook is registered.
2. Make the replication framing binary for entry payloads (length-prefixed raw bytes instead of JSON plus base64), which removes the 4/3 expansion and makes the two 100MB caps actually align.
3. Split oversized entries at the replication layer before framing.

Option 2 fixes the mismatch at its root and shrinks frames for every entry, but it is a protocol change and needs versioning against older peers. Option 1 is a small, safe stopgap.

PR #715, opened 2026-09-11
security(cluster): bind SupportsBinaryEntries into the replicate-sync HMAC

### LINKED ISSUE #714: security(cluster): ReplicateSync.SupportsBinaryEntries is outside the replicate-sync HMAC
Follow-up from the 26.09.2 coordinator-handshake work (#713).

`protocol.ReplicateSync.SupportsBinaryEntries` (`internal/cluster/protocol/messages.go`) is consumed by the writer — `coordinator.go` reads it to decide whether to send `MsgReplicateEntryBin` frames — but is not covered by the replicate-sync HMAC.

**The original justification is now obsolete.** The comment said folding it in "would break every mixed-version handshake". 26.09.2 made the coordinator handshake a hard cutover regardless, so that argument no longer holds as written. It was left unsigned in #713 because replicate-sync is a separate message and validator from the handshake family, and widening a wire-break during a security fix is the kind of scope creep that has bitten us before (#534). The comment has been corrected to say so.

**Impact is low.** An on-path attacker who flips the flag gets a framing downgrade (binary → JSON+base64) or a dropped connection — both already available to anyone who can modify the stream. There is no data-integrity or authorization consequence. It is a completeness gap, not an exploitable one.

**Suggested fix:** bind it into `ComputeReplicateSyncHMAC` / `ValidateReplicateSyncHMAC`, adopting the same "the MAC covers every non-auth field of the message" rule the handshake family now follows. That is another wire-break for the replication channel, so it should ride along with the next release that already requires a coordinated restart rather than forcing one on its own.

While in there, check `FetchFileRequest.ByteOffset` (`messages.go`) — also unsigned, but its rationale (a resumed fetch would otherwise invalidate the MAC) is still valid and should be left alone with the reasoning documented.

Found by internal review on #713.

pujitha24:
Motivation:
protocol.ReplicateSync.SupportsBinaryEntries negotiates binary-framed
(MsgReplicateEntryBin) vs JSON+base64 WAL entry replication, but rode
outside the replicate-sync HMAC while every other field of the message
was covered. It was left unsigned on the reasoning that folding it in
would break mixed-version handshakes; that reasoning no longer holds
now that the coordinator handshake is a hard cutover regardless, so
this rides along with the same already-mandated coordinated-restart
release rather than forcing a second cutover.

Impact is low, not a new exploitable gap: an on-path tamperer who
flips the unsigned flag can only force a framing downgrade (binary to
JSON) or a dropped connection, both already available to anyone who
can modify the stream. This closes a completeness gap, matching the
"the MAC covers every non-auth field" rule the handshake family
already follows.

FetchFileRequest.ByteOffset, flagged in the same report as similarly
unsigned, is intentionally left alone: a resumed fetch would otherwise
invalidate the MAC, and that rationale still holds.

Approach:
- ComputeReplicateSyncHMAC / ValidateReplicateSyncHMAC (security/auth.go)
  take a new supportsBinaryEntries bool, folded into the signed message
  right before the timestamp.
- Both call sites updated: the writer (coordinator.go, validates) now
  passes syncReq.SupportsBinaryEntries; the reader (replication/receiver.go,
  computes) uses one local constant for both the struct field and the MAC
  input, so the two can't drift apart.
- Doc comments on ReplicateSync / SupportsBinaryEntries (protocol/messages.go)
  updated to say the field is now covered.

Validation:
- go build ./... and go vet ./internal/cluster/... — clean.
- gofmt -l on every changed file — clean.
- go test -tags=duckdb_arrow -race ./internal/cluster/... — all packages
  pass, including internal/cluster/security.
- Updated TestComputeReplicateSyncHMAC_Determinism's pinned hash for the
  new wire format (independently recomputed, matches).
- Added TestValidateReplicateSyncHMAC_RejectsFlippedSupportsBinaryEntries:
  computes a MAC with the field true, confirms it validates, then confirms
  validation now fails once the field is flipped to false (and the
  symmetric case). It does not compile against the pre-fix code, since the
  old function signature ignored the field entirely — a weaker guarantee
  than "fails on the pre-fix code," but it still pins the signed-tuple
  property directly and would catch a future change that dropped the field.

Report: https://github.com/Basekick-Labs/arc/issues/714

Fixes #714

### DIFF
--- a/RELEASE_NOTES_2026.09.2.md
+++ b/RELEASE_NOTES_2026.09.2.md
@@ -191,12 +191,16 @@ is published. Responsibly reported by **[@rexpository](https://github.com/rexpos
 ## Upgrade notes
 
 1. **Clustered Enterprise deployments require a coordinated restart.** The
-   coordinator handshake wire format changed (see *Cluster hardening* above):
-   stop all cluster nodes, upgrade the binary on every node, then restart all
-   nodes. A rolling restart causes cross-version nodes to mark each other
-   unhealthy, can trigger an automatic writer failover, and can remove a
-   gracefully-restarted follower from the Raft configuration until its leader
-   is upgraded. Single-node, non-clustered and OSS deployments need no action.
+   coordinator handshake **and the replicate-sync handshake** wire formats
+   changed (see *Cluster hardening* and *Replicate-sync now authenticates
+   `SupportsBinaryEntries`* above): stop all cluster nodes, upgrade the binary
+   on every node, then restart all nodes. A rolling restart causes
+   cross-version nodes to mark each other unhealthy, can trigger an automatic
+   writer failover, and can remove a gracefully-restarted follower from the
+   Raft configuration until its leader is upgraded. A cross-version reader
+   additionally cannot establish WAL replication with a writer and will fall
+   behind until both ends are upgraded. Single-node, non-clustered and OSS
+   deployments need no action.
 2. **No configuration change is required.** Existing `arc.toml` files and
    license keys work as-is; no new keys were added.
 
--- a/RELEASE_NOTES_2026.09.2.md
+++ b/RELEASE_NOTES_2026.09.2.md
@@ -148,6 +148,26 @@ OSS deployments are unaffected.
 Full technical detail will accompany the corresponding security advisory once
 it is published. Responsibly reported by **[@rexpository](https://github.com/rexpository)**.
 
+### Replicate-sync now authenticates `SupportsBinaryEntries` ([#714](https://github.com/Basekick-Labs/arc/issues/714))
+
+The replicate-sync handshake (a reader requesting WAL replication from a
+writer) authenticates `{reader_id, last_known_seq, nonce, cluster_name,
+timestamp}`, but the `SupportsBinaryEntries` flag that negotiates binary-framed
+WAL entries (see *Replication can now carry WAL entries up to the full
+payload cap* below) rode outside the MAC. It was left unsigned on the
+reasoning that folding it in would break mixed-version handshakes; that
+reasoning no longer holds once the coordinator handshake above became a hard
+cutover regardless, so the flag is now bound into `ComputeReplicateSyncHMAC`
+and `ValidateReplicateSyncHMAC` alongside the rest of the message. An
+on-path tamperer who flipped the previously-unsigned flag could only force a
+framing downgrade or a dropped connection — both already available to anyone
+who can modify the stream — so this closes a completeness gap rather than a
+previously exploitable one. It carries the same replicate-sync wire format
+change and coordinated-restart requirement as the handshake hardening above,
+so it ships in the same release rather than forcing a second cutover.
+
+Contributed by [@pujitha24](https://github.com/pujitha24).
+
 ### Expired API tokens are now rejected on cache hits
 
 Arc caches successful token verifications in memory for `auth.cache_ttl`
--- a/RELEASE_NOTES_2026.09.2.md
+++ b/RELEASE_NOTES_2026.09.2.md
@@ -166,7 +166,7 @@ previously exploitable one. It carries the same replicate-sync wire format
 change and coordinated-restart requirement as the handshake hardening above,
 so it ships in the same release rather than forcing a second cutover.
 
-Contributed by [@pujitha24](https://github.com/pujitha24).
+Contributed by [@pujitha24](https://github.com/pujitha24) in [#715](https://github.com/Basekick-Labs/arc/pull/715).
 
 ### Expired API tokens are now rejected on cache hits
 
--- a/internal/cluster/coordinator.go
+++ b/internal/cluster/coordinator.go
@@ -1582,7 +1582,7 @@ func (c *Coordinator) handleReplicateSync(conn net.Conn, syncReq *protocol.Repli
 	}
 	if err := security.ValidateReplicateSyncHMAC(
 		c.cfg.SharedSecret, syncReq.Nonce, syncReq.ReaderID, syncReq.ClusterName,
-		syncReq.LastKnownSequence, syncReq.Timestamp, syncReq.HMAC, security.HMACTimestampTolerance,
+		syncReq.LastKnownSequence, syncReq.SupportsBinaryEntries, syncReq.Timestamp, syncReq.HMAC, security.HMACTimestampTolerance,
 	); err != nil {
 		c.logger.Warn().
 			Err(err).
--- a/internal/cluster/protocol/messages.go
+++ b/internal/cluster/protocol/messages.go
@@ -94,12 +94,12 @@ func (m MessageType) String() string {
 
 // ReplicateSync is sent by a reader node to request WAL replication.
 //
-// The five auth fields (Nonce / Timestamp / HMAC / ClusterName / and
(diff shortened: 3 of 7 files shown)


xe-nvdk (commented):
Thanks for picking this up, and for reading the issue carefully — the scoping decisions here are right, including the ones that required judgment.

I reviewed this in a worktree rather than from the diff alone. **The code is correct and I would merge it.** One blocker that is not about the code, one substantive gap, and a nit.

## Verified

- **The pinned vector is genuinely recomputed, not copy-pasted from a failing run.** I rebuilt it from first principles — raw `hmac.New(sha256.New, []byte("secret"))` over the literal `"replicate-sync\x00nonce-abc\x00reader-1\x00cluster-A\x0042\x00true\x001700000000"` — and got `6da169ff…41c5b`, matching both your constant and the package helper. This is the single easiest thing to get subtly wrong in a change like this, so it is worth saying explicitly that it holds.
- **Signer and validator agree end-to-end.** The receiver signs with its local `supportsBinaryEntries` const while the coordinator validates with the wire value `syncReq.SupportsBinaryEntries`. I probed the real pairing: untampered validates, flipped wire flag rejects. Using one `const` for both the struct field and the MAC input is exactly the right move — it makes the drift you would otherwise have to test for structurally impossible.
- **All call sites updated.** Exactly one signer (`replication/receiver.go:288`), one validator (`coordinator.go:1583`), one `protocol.ReplicateSync{}` constructor. No stragglers.
- `go build ./...`, `go vet ./internal/cluster/...`, `gofmt -l` — clean. `go test -race ./internal/cluster/...` — all 6 packages pass.
- **`%t` is unambiguous here.** `true`/`false` cannot collide with the adjacent `%d` fields, and I tried to forge the bool position via a NUL-bearing `readerID` — no collision. The NUL-delimited format stays sound for this fixed-arity message (which is why the handshake family needed length-prefixing and this one does not).

## Blocker (process, not code)

**The CLA check is failing** — `Committers of Pull Request number 715 have to sign the CLA`. Nothing to change in the branch; comment on this PR with:

```
```

Everything else is green once that clears.

## One substantive gap: the upgrade note

This is a **second** wire-format break, on a different channel than the one already documented. `## Upgrade notes` item 1 names only the coordinator handshake and describes only its failure modes (nodes marking each other unhealthy, writer failover, a gracefully-restarted follower dropping out of Raft). Replicate-sync failing looks different — a reader cannot establish WAL replication with a writer of the other version, so replication stalls and the reader falls behind.

The same stop-all/upgrade-all/restart-all covers both, so no new procedure. But an operator reading item 1 would not learn that replication is also affected. Suggest extending item 1, roughly:

> The coordinator handshake **and the replicate-sync handshake** wire formats changed (see *Cluster hardening* and *Replicate-sync now authenticates `SupportsBinaryEntries`* above) … A cross-version reader additionally cannot establish WAL replication and will fall behind until both ends are upgraded.

## Nit

The PR body says the new test "fails on the pre-fix code". It does not compile against pre-fix code, because the function signature changed — which is a weaker claim. I hit the same thing on #713 and called it out there for the same reason: worth stating precisely, since "fails pre-fix" is the property we normally rely on to prove a regression test is real. The test is still a good one; it pins the property directly and would catch a future change that dropped the field from the signed tuple.

## On the scoping calls

Both were right, and I want to say so since they are the parts a reviewer would otherwise second-guess:

- **Leaving `FetchFileRequest.ByteOffset` alone** is correct. Its rationale (a resumed fetch at a different offset would invalidate the MAC) is still live, unlike the one this PR retires.
- **Riding along with the release that already mandates a coordinated restart** is the right trade. Two cutovers for one release would be worse than one, and the "widening a wire-break inside a security fix" concern from #714 does not apply here — that fix has already shipped, so this is no longer expanding anything in flight.

Sign the CLA and add the upgrade-note sentence, and this is good to go.

pujitha24:
- - -
- - -

Extended Upgrade notes item 1 to name the replicate-sync handshake alongside the coordinator handshake, with the note that a cross-version reader can't establish WAL replication and falls behind until both ends are upgraded. Also fixed the PR body wording — you're right that the test doesn't compile against pre-fix code rather than compiling and failing, which is a weaker but accurate claim; updated it in the description.

xe-nvdk (approved):
Reviewed this one in depth, since it revisits a decision I made deliberately in #698 and touches the replication auth path. It is correct, and the reasoning that reverses my original call is sound: #713 turned the coordinator handshake into a hard cutover, which is exactly the condition under which leaving this flag unsigned stopped paying for itself. Approving.

What I verified rather than took on faith:

**The MAC construction.** `%t` renders inside the existing NUL-delimited format, and since neither `true` nor `false` can contain `\x00` the encoding stays unambiguous with a fixed field count. Compute and validate take the flag in the same position, and the only two production call sites (the receiver's compute, the coordinator's validate) both moved together.

**The golden value is real.** I recomputed the determinism constant independently from the format documented in the doc comment: HMAC-SHA256 over `"replicate-sync\x00nonce-abc\x00reader-1\x00cluster-A\x0042\x00true\x001700000000"` keyed with `"secret"` gives `6da169ff...b741c5b`, matching the new constant exactly. As a control, the pre-change format reproduces the old constant. So the comment, the implementation, and the test agree, which is the property that actually matters for a pinned MAC.

**Signed value cannot drift from sent value.** The receiver's `const supportsBinaryEntries = true` feeding both the struct field and the HMAC call is the right shape; it removes the classic failure where the two are edited apart later.

**The binding is not bypassable by ordering.** Validation at coordinator.go:1583 precedes every consumption of the flag, both the `replSyncReq` hand-off and `EnableBinaryEntries`. And the `omitempty` tag is safe here: stripping the field in transit makes the writer validate against `false` while the reader signed `true`, so tampering fails closed rather than silently downgrading.

**The documented wire break is real.** The upgrade note is the operationally load-bearing claim, and nothing tested it, so I wrote a throwaway probe: an old-format MAC is refused by the new validator whether it validates with the flag true or false, a new-format MAC differs from what an old writer would expect, and a new pair still handshakes. Both directions break cleanly as an auth rejection, so the coordinated-restart wording is accurate.

`ByteOffset` correctly left alone with its existing rationale intact. Build, vet, gofmt clean; all `internal/cluster/...` suites green under `-race`.

Two trivia I pushed to your branch rather than bounce back (1f2932f): the credit line was missing its PR link, unlike every other entry in the file, and the `MsgReplicateEntryBin` comment in `replication/protocol.go` still explained the per-reader gate as letting "mixed-version pairs keep speaking JSON", which this change makes obsolete since cross-version readers can no longer complete the handshake at all.

The one thing still blocking the merge is the CLA, which I have written up separately in a comment.

xe-nvdk:
@pujitha24 one snag on the CLA, and it affects #707 too.

Your signing comment did not register with the bot. It arrived wrapped in `- - -` separators with an extra paragraph after it, and the action needs a comment whose body is the signing sentence on its own. The signature registry currently has no entry for you, which is why both this PR and #707 still show the CLA check red.

Could you post a new comment containing exactly this and nothing else, no separators, no additional text:

The signature is recorded once and applies across the repo, so that single comment turns the check green on both PRs. Anything you want to add about the change itself is welcome as a separate comment.

Both PRs are reviewed and approved on the technical side, so this is the only thing left before they merge.

pujitha24:

PR #766, opened 2026-09-13
fix(mqtt): prevent TOCTOU race in RestartSubscription with nil-placeholder (#301)

### LINKED ISSUE #301: high(mqtt): TOCTOU race in RestartSubscription
manager.go:406-435 — no slot reservation. Fix: use nil-placeholder pattern.

Thundercloud12:
## Summary

* Fixed a TOCTOU race in `RestartSubscription` where deleting `m.subscribers[id]` before restarting allowed concurrent `StartSubscription` or `RestartSubscription` calls to create duplicate subscribers.
* Added the existing nil-placeholder reservation pattern used by `StartSubscription`.
* Concurrent start/restart attempts now return `ErrSubscriptionAlreadyRunning` while the slot is reserved.
* Moved `subscriber.Stop()` outside `m.mu` to avoid holding the global lock during MQTT disconnect timeouts.

Fixes #301

### `internal/mqtt/manager.go`

* Reject restart when `m.subscribers[id]` is already reserved (`subscriber == nil`).
* Reserve the slot with `m.subscribers[id] = nil` under the mutex.
* Stop the existing subscriber and load the subscription config outside the lock.
* Remove the placeholder if fetching or starting the subscriber fails.

### `internal/mqtt/manager_test.go`

* Added coverage for concurrent restart rejection.
* Added coverage for placeholder cleanup when the subscription is not found.

### `RELEASE_NOTES_2026.09.2.md`

* Documented the fix in the `2026.09.2` release notes.

## Test Plan

* [x] `go test ./internal/mqtt/... -v -run "TestManager_RestartSubscription"`
* [x] `go test ./internal/mqtt/...`


Thundercloud12:

xe-nvdk (changes requested):
Thanks — the fix is correct and I verified it live. I built this branch and a pre-fix build, ran each against a Mosquitto broker, and fired 20 concurrent `POST /restart` at one subscription:

| | PR #766 | pre-fix `main` |
|---|---|---|
| HTTP responses | 3× 200, **17× 500** | 20× 200 |
| subscribers started (manager log) | 4 | 21 |
| broker CONNECTs from that `client_id` in 4s | 4 | 76 — orphans kicking each other off the shared client_id, and it never settles |

So #301 is fixed: no duplicate subscribers, and moving `Stop()` out from under `m.mu` is a real improvement (a restart used to hold the manager write lock for the unsubscribe waits + 1s disconnect, blocking stats/list). Three things before merge, one suggestion:

**1. `handleRestart` doesn't map the new error (required).** `RestartSubscription` can now return `ErrSubscriptionAlreadyRunning`, but `internal/api/mqtt_subscriptions.go#handleRestart` only has the generic branch, so the 17 rejected restarts above came back as `500 "Failed to restart subscription"` plus an error-level log line each. A double-click on "restart" in the UI now looks like an outage. Mirror `handleStart`:

```go
if errors.Is(err, mqtt.ErrSubscriptionAlreadyRunning) {
    return c.Status(fiber.StatusConflict).JSON(fiber.Map{
        "success": false,
        "error":   err.Error(),
    })
}
```

and add a handler test for it (the start-side tests in `mqtt_subscriptions_test.go` are the pattern).

**2. gofmt (required).** `gofmt -l ./internal/mqtt` flags `manager_test.go` (trailing blank line at EOF). CI doesn't gate gofmt yet, so please run it locally before pushing.

**3. Release notes (required).** Add the `Contributed by [@Thundercloud12](https://github.com/Thundercloud12) in [#766](https://github.com/Basekick-Labs/arc/pull/766).` line like your #670 entry, and say what an operator actually sees: a start/restart that lands during an in-flight restart now gets `409 Conflict` (once item 1 is in), and the old subscriber's disconnect no longer runs under the manager lock.

**4. Test the lock-out path (suggested).** `PlaceholderCleanedOnNotFound` is fine, but note it also passes on `main` (the old code never inserted a placeholder). The path that matters most is a failed `startSubscriber` — broker unreachable, bad TLS config — because a leaked placeholder there means the subscription can never be started or restarted again. This version is deterministic (no network: the bogus CA path fails in `buildClientOptions`) and passes on your branch:

```go
// TestManager_RestartSubscription_PlaceholderCleanedOnStartFailure verifies the
// reserved slot is released when the new subscriber fails to start, so the
// subscription is not locked out (#301).
func TestManager_RestartSubscription_PlaceholderCleanedOnStartFailure(t *testing.T) {
	mgr := newTestManager(t)
	ctx := context.Background()

	sub := &Subscription{
		Name:       "test-sub",
		Broker:     "ssl://localhost:8883",
		ClientID:   "test-client",
		Topics:     []string{"sensors/#"},
		QoS:        1,
		Database:   "iot",
		TLSEnabled: true,
		TLSCAPath:  filepath.Join(t.TempDir(), "missing-ca.pem"),
	}
	sub.SetDefaults()
	if err := mgr.repo.Create(ctx, sub); err != nil {
		t.Fatalf("repo.Create: %v", err)
	}

	if err := mgr.RestartSubscription(ctx, sub.ID); err == nil {
		t.Fatal("expected start failure, got nil")
	}

	mgr.mu.RLock()
	_, exists := mgr.subscribers[sub.ID]
	mgr.mu.RUnlock()
	if exists {
		t.Fatal("placeholder left in subscribers map after failed start; subscription would be locked out")
	}

	got, err := mgr.repo.Get(ctx, sub.ID)
	if err != nil || got == nil {
		t.Fatalf("repo.Get: %v", err)
	}
	if got.Status != StatusError {
		t.Errorf("status = %q, want %q", got.Status, StatusError)
	}

	if err := mgr.RestartSubscription(ctx, sub.ID); errors.Is(err, ErrSubscriptionAlreadyRunning) {
		t.Fatalf("second restart rejected as already running: %v", err)
	}
}
```

Pre-existing and not for this PR: `startSubscriber` installs into the map unconditionally, so a `Delete` (or `Shutdown`) racing an in-flight start/restart can still leave a live subscriber for a row that no longer exists. `StartSubscription` has the same shape; filed as #770.

Thundercloud12:
@xe-nvdk resolved the conflicts and updated the pr, please let me know if any more changes are required!

xe-nvdk:
> @xe-nvdk resolved the conflicts and updated the pr, please let me know if any more changes are required!

Awesome job. Thank you. I'm reviewing this now.

xe-nvdk (approved):
Second pass on 09bc671, verified the same way as the first round. Approving.

- **Handler mapping:** the new `TestMQTTSubscriptionHandler_RestartAlreadyRunningConflict` fails on the old handler (`status: got 500, want 409`) and passes here. Live, against a Mosquitto broker with 20 concurrent `POST /restart` on one subscription: **2× 200, 18× 409, zero 500s**, one broker client, three connects total (the initial start plus the two restarts that won). Same run on round one gave 17 × 500.
- **Lock-out test:** in, and it passes on this branch for the right reason (the bogus CA path fails in `buildClientOptions` before any network I/O).
- gofmt and vet are clean on both packages, `internal/mqtt` passes with `-race`, and the branch is rebased on current `main`.
- Release note has the credit line and says what an operator sees. One tidy-up I'll do at merge rather than send you another round: entries under `## Bug fixes` are ordered newest-first, so yours belongs at the top of that section, not the end. No action needed from you.

Optional, not for this PR: `handleRestart` and `handleStart` both log the benign conflict at Error level before returning the 409. Worth a Warn in a follow-up, since 18 error lines for a double-click is noise.

Thank you for a clean second round. Once CI finishes this merges, and #770 (the install race I filed during review) is yours if you want it: same file, same pattern.

xe-nvdk:
Merging — thank you @Thundercloud12, two clean rounds and a fix I could verify live. README credit follows in a moment. #770 is yours if you want it.

PR #767, opened 2026-09-13
security(query): add PREPARE and EXECUTE to dangerousSQLPattern (#739)

### LINKED ISSUE #739: Add PREPARE/EXECUTE to dangerousSQLPattern as defense-in-depth
Found while fixing GHSA-w6w2-x8xv-q8x2.

`PREPARE` and `EXECUTE` are DuckDB's other way to run SQL indirectly. They are **not exploitable today** because they need two statements (PREPARE then EXECUTE) and `ValidateSQLRequest`'s single-statement rule rejects the chain, and a bare `EXECUTE foo` errors with "Prepared statement does not exist".

But they are not in `dangerousSQLPattern` (they live only in the separate `dangerousQueryPatterns` WHERE-clause blocklist). If the statement-splitter is ever relaxed, or a future DuckDB accepts a one-shot form, the dynamic-SQL family reopens with no guard.

Cheap defense-in-depth: add `PREPARE` and `EXECUTE` to `dangerousSQLPattern` (`internal/api/query.go`, ~line 737) so they are rejected up front regardless of statement count, and extend the validator test family.

Deliberately kept out of the GHSA-w6w2 fix, which is a targeted patch for a confirmed-exploitable path; this is hardening for a path that is not currently reachable.

Thundercloud12:
## Summary

* Blocked `PREPARE` and `EXECUTE` in `dangerousSQLPattern` to prevent dynamic SQL execution through all user-SQL endpoints:

  * `/api/v1/query`
  * `/api/v1/query/arrow`
  * `/api/v1/query/msgpack`
* Added defense-in-depth protection against dynamic SQL execution attempts, following up on #738 / GHSA-w6w2-x8xv-q8x2.

Fixes #739

## Motivation & Background

DuckDB supports indirect SQL execution through `PREPARE` and `EXECUTE`. While Arc's current single-statement restriction prevents the known attack path, these keywords were not previously blocked by `dangerousSQLPattern`.

Blocking both keywords ensures dynamic SQL execution is rejected at the read-API boundary, even if SQL parsing or statement restrictions change in the future.

### `internal/api/query.go`

* Added `\bPREPARE\b` and `\bEXECUTE\b` to `dangerousSQLPattern`.

### `internal/api/query_test.go`

* Added tests for:

  * Direct `PREPARE`/`EXECUTE` statements.
  * Arguments and comment-based bypass attempts.
  * Double-quoted identifiers such as `"PREPARE"`.
  * String literals containing `PREPARE`/`EXECUTE`.
* Added `TestValidateSQLRequest_BlocksPrepareAndExecute`.

### `RELEASE_NOTES_2026.09.2.md`

* Added a release note entry referencing #739.

## Test Plan

* [x] `go test -tags=duckdb_arrow ./internal/api/ -run "TestValidateSQLRequest" -v`
* [x] `go test ./internal/api/ -run "TestValidateSQLRequest"`

### DIFF
--- a/RELEASE_NOTES_2026.09.2.md
+++ b/RELEASE_NOTES_2026.09.2.md
@@ -1391,3 +1391,8 @@ warehouse metadata directory, after the files are removed, so a racing
 reconcile pass can only empty tables rather than recreate them. Cleanup
 failures are logged and re-running the DELETE retries them, including when the
 files are already gone.
+
+### PREPARE and EXECUTE statements blocked in query validator ([#739](https://github.com/Basekick-Labs/arc/issues/739))
+
+`ValidateSQLRequest` now rejects `PREPARE` and `EXECUTE` commands up front in `dangerousSQLPattern`, hardening against dynamic SQL execution attempts across all user query endpoints.
+
--- a/internal/api/query.go
+++ b/internal/api/query.go
@@ -764,6 +764,10 @@ var dangerousSQLPattern = regexp.MustCompile(`(?i)(?:` +
 	// bypassable via `CALL(proc)` (no whitespace before paren) — gemini
 	// round 1.
 	`|\bCALL\b` +
+	// PREPARE / EXECUTE match bare keywords as defense-in-depth against
+	// dynamic SQL evaluation (#739).
+	`|\bPREPARE\b` +
+	`|\bEXECUTE\b` +
 	// Secrets manager: CREATE/DROP SECRET lets any authenticated user replace
 	// or delete Arc's S3 credentials (redirecting reads to attacker-controlled
 	// creds, or knocking out S3 access entirely) and probe the secret manager.
--- a/internal/api/query_test.go
+++ b/internal/api/query_test.go
@@ -1006,18 +1006,27 @@ func TestValidateSQLRequest_BypassesAndFalsePositives(t *testing.T) {
 		{name: "DROP dash-comment then newline", sql: "DROP --evil\nTABLE users", shouldFail: true},
 		{name: "ATTACH block-comment", sql: "ATTACH /* sneaky */ 'evil' AS x", shouldFail: true},
 		{name: "LOAD block-comment", sql: "LOAD /**/'http://x'", shouldFail: true},
+		{name: "PREPARE statement blocked", sql: "PREPARE p AS SELECT * FROM cpu", shouldFail: true},
+		{name: "EXECUTE statement blocked", sql: "EXECUTE p", shouldFail: true},
+		{name: "EXECUTE with arguments blocked", sql: "EXECUTE p(1, 2)", shouldFail: true},
+		{name: "PREPARE block-comment evasion", sql: "PREPARE/**/p AS SELECT 1", shouldFail: true},
+		{name: "EXECUTE block-comment evasion", sql: "EXECUTE/**/p", shouldFail: true},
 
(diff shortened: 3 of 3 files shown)


xe-nvdk (commented):
Thanks — this does exactly what #739 asked, and I verified it end to end:

- All 13 new cases fail on `main` and pass on this branch (I ran them against the pre-fix `query.go`).
- The pattern runs after literal masking and comment stripping, so the `"PREPARE"` identifier and `'EXECUTE query'` literal false-positive cases are genuinely exercised, and `EXECUTE(p)` / `EXECUTE/**/p` are caught.
- Every user-SQL entry point goes through `ValidateSQLRequest`: `/query`, `/query/arrow`, `/query/msgpack` (thin wrapper over `executeQuery`), `/query/estimate`, the measurement endpoint, and the CQ probe — so "all user query endpoints" in the note is accurate.
- Cost: I benchmarked `ValidateSQLRequest` on three typical queries, ABAB. `main` ≈36µs/query, this branch ≈42µs/query. Real but immaterial next to DuckDB execution; noting it for the record, not asking for a change.

Two small things before merge:

**1. Release note placement and wording.** This is hardening, not a fix for a reachable hole (#739 says so explicitly: the single-statement rule already rejects the PREPARE→EXECUTE chain). Please move the entry from the end of `## Bug fixes` to `## Security fixes`, directly under the `Read-SQL validator hardening (GHSA-w6w2-x8xv-q8x2)` entry it follows up, and say that. Suggested text:

> ### PREPARE and EXECUTE are rejected by the read-SQL validator ([#739](https://github.com/Basekick-Labs/arc/issues/739))
>
> `PREPARE` and `EXECUTE`, DuckDB's indirect-execution statements, are now blocked up front by the read-SQL validator on every user query endpoint. They were not exploitable before this change — the single-statement rule already rejects the two-statement chain — so this is defense-in-depth in case the statement splitter is ever relaxed. As with the other blocked keywords, a column literally named `prepare` or `execute` must be double-quoted.
>
> Contributed by [@Thundercloud12](https://github.com/Thundercloud12) in [#767](https://github.com/Basekick-Labs/arc/pull/767).

**2. gofmt.** `gofmt -l ./internal/api` flags `query_test.go` (trailing blank line at EOF). It was already dirty on `main`, but since you're touching the file, please run gofmt on it so CI can gate it later.

Heads-up: #766 appends to the same spot in the release notes, so whichever of the two merges second will need a quick rebase.

Thundercloud12:
@xe-nvdk requested changes are done!

xe-nvdk (approved):
Both points addressed on 1eeb37c. The note now sits under `## Security fixes` directly after the GHSA-w6w2 entry with the credit line, `gofmt -l ./internal/api` is empty (the file is cleaner than `main` was), vet is clean, the validator family passes with `-race`, and the code is byte-identical to the round I verified end to end. Approving.

Two whitespace nits you can fold into #766's rebase rather than push a commit for: there are three blank lines after the credit line before `### Dependency bump`, and the last hunk leaves a stray blank line at the end of the file. That trailing blank line is also the one spot where this PR still touches the same region as #766, so whichever of the two merges second gets a one-line rebase. Dropping it here would make #766 merge clean.

xe-nvdk:
Merging — thank you @Thundercloud12. Clean, well-tested, and the notes are exactly where they belong. You'll be credited in the README shortly. Looking forward to the #766 update.

PR #769, opened 2026-09-13
fix(ingest): log numeric host coercion at debug level

### LINKED ISSUE #768: fix: nil check on dispatch handler before processing request
Hi team,

I was reading through the dispatch handler code and noticed there's no explicit nil check before processing the request. If the handler ever ends up nil due to a race condition or misconfiguration, this could panic in production.

I saw your blog post about writing good first issues with clear file/line guidance and the intended shape of the fix, so I wanted to propose this as a candidate:

- Add a guard clause at the start of dispatch to check if the handler is nil
- Return a controlled error instead of panicking

Is this something that would fit as a good first issue? If so, I'd be happy to give it a shot myself.

Thanks!

lecodev-26:
## What this PR does

Fixes #768.

The `extractHost` fallback that coerces numeric values into `host_<num>` strings is **intentional** (covered by `TestMessagePackDecoder_ExtractHost`), so this PR keeps the existing behaviour unchanged.

However, that fallback silently swallows a likely sign of misconfigured clients. This PR adds a **debug-level log** when the numeric fallback fires, so operators can spot it without flooding logs on the hot ingest path.

It also adds two test cases (`float host`, `boolean host`) which fall through to the `default` branch and return `"unknown"` — covering the non-numeric, non-string path.

## Why debug level

The hot ingest path handles tens of millions of records per second. A warn/info log here would be noise. Debug is opt-in and appropriate for a signal that is useful but not urgent.

## Files changed

- `internal/ingest/msgpack.go` — added debug log in `extractHost`
- `internal/ingest/msgpack_test.go` — added 2 test cases


lecodev-26:

lecodev-26:
recheck

lecodev-26:
Hi @xe-nvdk,

I've signed the CLA (comment above) and also tried `recheck` and an empty commit to retrigger the workflow, but the `CLA/CLA signature` check is still failing.

This looks like the known issue with CLA Assistant Lite doing an exact-match comparison — a trailing newline or invisible character in the comment can make it miss a valid signature.

Let me know if there's anything else I need to do on my end, or if the workflow needs a manual re-run on your side. Happy to adjust whatever's needed.

Thanks!

xe-nvdk (changes requested):
Thanks for the write-up and for correcting course on #768 — the numeric-host coercion is a real signal worth surfacing. I ran the ingest suite with `-race` on this branch (green) and benchmarked the change. A few things to sort out before it can go in:

**1. CLA — the signature doesn't cover the committer (blocking).** You signed as `@lecodev-26`, but both commits are authored with an email GitHub attributes to a different account (`axiom-systemstech`), and the CLA bot checks committers, not the PR opener. Two ways to fix it:
- re-author the two commits with an email linked to `lecodev-26` (your GitHub noreply address works: `git rebase -r --exec 'git commit --amend --no-edit --reset-author'` after setting `user.email`), force-push, then comment `recheck`; or
- post the signature sentence from the `axiom-systemstech` account as well.

**2. The log line doubles the cost of this path even when debug is off (required).** `fmt.Sprintf("%v", v)` runs before zerolog gets a chance to drop the event, so every numeric-host record pays a second format + allocation regardless of level. Measured on this branch, debug disabled:

| | ns/op | allocs/op |
|---|---|---|
| `main` | 40 | 1 |
| this PR | 80 | 2 |
| format once, log the same string | 44 | 1 |

Format the host once and log that:

```go
case int, int8, int16, int32, int64, uint, uint8, uint16, uint32, uint64:
	host := fmt.Sprintf("host_%v", v)
	d.logger.Debug().Str("host", host).Msg("msgpack: numeric host coerced to host_<num>")
	return host
```

**3. Consider a counter instead of (or alongside) the log (suggested).** Operators run at info level, so they'll only see this line after switching to debug, and at that point a misconfigured client emits one line per record. The decoder already keeps `totalDecoded` / `totalErrors` atomics and exposes them through `GetStats()` on `GET /api/v1/write/msgpack/stats`. A `numericHostCoercions atomic.Uint64` bumped on this path and reported as `numeric_host_coercions` is always-on, costs one atomic add, and is what an operator would actually notice.

**4. Tests should exercise the change.** The `float host` / `boolean host` cases are fine to have, but they pass on `main` unchanged — nothing in the PR verifies that the log (or a counter) fires for a numeric host and stays quiet for a string one. A decoder built on `zerolog.New(&buf).Level(zerolog.DebugLevel)` lets you assert on `buf` directly.

**5. Release notes.** Every code change in this repo carries a `RELEASE_NOTES_2026.09.2.md` entry; please add a short one under `## Bug fixes` (or `## Changed`) with a `Contributed by [@lecodev-26](https://github.com/lecodev-26) in [#769](https://github.com/Basekick-Labs/arc/pull/769).` line.

**6. Small things.** Commit messages here follow `fix(scope): …` (`fix(ingest): log numeric host coercion at debug level`). And the new doc comment says the coercion exists "for backwards compatibility" — the history only shows it arriving with the Go port, so I'd drop the rationale and just describe what the code does.

axiom-systemstech:

lecodev-26:
Hi @xe-nvdk, thanks for the detailed review — really appreciated.

I've applied all six points:

1. **CLA**: signed from `axiom-systemstech` (the commit author). CLA check is now green.
2. **Sprintf double call**: replaced with `host := fmt.Sprintf("host_%v", v)` and reused for both the log and the return value. Benchmark impact should now match the "format once" line you measured.
3. **Counter suggestion**: I kept the debug log for now (the atomic counter is a bigger change I'd rather propose in a separate PR, unless you'd prefer it here).
4. **Test coverage**: added `TestMessagePackDecoder_ExtractHost_LogsNumericCoercion`, which uses `zerolog.New(&buf).Level(zerolog.DebugLevel)` and asserts the log fires for numeric hosts and stays silent for string and nil hosts.
5. **Release notes**: added a `## Fixed` entry under `RELEASE_NOTES_2026.09.2.md`.
6. **Commit message + comment**: commits now use `fix(ingest): ...` / `test(ingest): ...`, and the doc comment no longer mentions "backwards compatibility" — it just describes what the code does.

Happy to adjust anything else. Let me know if you'd prefer the counter approach instead of (or alongside) the log.

xe-nvdk (changes requested):
Second pass, on 3c2e2a8. CLA is green, gofmt and vet are clean, the ingest suite passes with `-race`, and the new log test fails on `main` and passes here — exactly what a regression test should do. I also approved the CI run; first-time contributors need a maintainer for that.

**Performance: the shape you chose is the right one, and you can drop the counter idea entirely.** I re-benchmarked the numeric-host path with debug disabled, serial and under `RunParallel` (which is what ingest looks like):

| | serial ns/op | parallel ns/op | allocs |
|---|---|---|---|
| `main` | 40 | 49 | 1 |
| this PR (format once, debug off) | 44 | 49.5 | 1 |
| per-record atomic counter instead | 42 | **75** | 1 |

Under parallel load your version is indistinguishable from `main`; a shared per-record atomic is 50% slower from cache-line contention. So please don't open the counter follow-up as a per-record atomic. If you ever want an always-on signal, it has to be a local count per decode batch folded into the existing `totalDecoded`-style atomics once per batch, but that is not needed for this PR.

Two small things and this merges:

**1. Release-notes entry format.** The file already has a `## Bug fixes` section (line 314); the new `## Fixed` header at the end of the file starts a second top-level section. Per CONTRIBUTING, make it a `###` entry inside `## Bug fixes`, in the same shape as its neighbours. Suggested text, placed after the last existing `###` entry in that section:

```markdown
### Numeric MessagePack host values are logged when coerced

The MessagePack decoder accepts a numeric `host` and coerces it to `host_<num>`.
That behaviour is unchanged, but the coercion now emits a debug-level log line
naming the resulting host, so a misconfigured client can be spotted without
changing ingest behaviour. The log reuses the string already built for the
return value, so the cost with debug disabled is not measurable under parallel
ingest.

Contributed by [@lecodev-26](https://github.com/lecodev-26) in [#769](https://github.com/Basekick-Labs/arc/pull/769).
```

Note that #766 and #767 touch the same end-of-file region, so expect a trivial rebase if either merges first.

**2. Doc comment reflow.** Removing "For backwards compatibility" left `A` dangling at the end of a line in the `extractHost` comment. Reflow the paragraph. (A blank line before `TestMessagePackDecoder_NormalizeTimestamps` would also be nice.)

I've set the PR title to the conventional-commit form myself, since it becomes the squash commit subject.

lecodev-26:
Hi @xe-nvdk, thanks for the second pass — really appreciate the benchmark data, and good to know the per-record atomic was the wrong shape. Dropping that idea.

Both points addressed in `ae6da7a`:

1. **Release notes**: moved the entry into the existing `## Bug fixes` section (after the last `###` entry), using your suggested wording verbatim. Also verified no second top-level `## Fixed` remains.
2. **Doc comment**: reflowed the `extractHost` paragraph (removed the dangling `A`), and added the blank line before `TestMessagePackDecoder_NormalizeTimestamps`.

`gofmt -l` is clean on both files.

Ready for another look when you have a moment.

xe-nvdk (approved):
Third pass on ae6da7a. Both points addressed: the note is now a `###` entry inside `## Bug fixes` with the credit line, no stray top-level section remains, the `extractHost` comment reads cleanly, and the blank line before the next test is in. The only code change since the round I benchmarked is that comment reflow, so the numbers stand. `gofmt -l` and `go vet` are clean and the ingest suite passes with `-race` on this head. I approved the CI run for the new commit. Approving.

One thing I'll tidy at merge rather than send you another round: entries under `## Bug fixes` are ordered newest-first, so yours belongs at the top of that section, and there's a double blank line before the heading. My earlier note said "after the last existing entry", which was my mistake, not yours. No action needed.

Thank you for three careful rounds. Once CI is green this merges, and you'll be credited in the README.

xe-nvdk:
Merging — thank you @lecodev-26 for three careful rounds. README credit follows in a moment.

lecodev-26:
Hi @xe-nvdk, thanks again for the three rounds — the review process was genuinely valuable.

I wrote a short write-up about the experience, including the benchmark data you shared (which I found illuminating):

https://lnkd.in/p/eTeNDTmT

Figured it might be useful signal for the project, but no action needed on your side.

Thanks again!

Issue #747, opened 2026-09-12
Cleanup and replication loops retry storage.ErrInvalidPath forever instead of quarantining


xe-nvdk:
Follow-up from #743, which made `storage.ErrInvalidPath` a permanent, identifiable error and documented that callers driving cleanup or reconciliation loops should quarantine rather than retry. Nothing branches on it yet, and these are the loops the doc comment is about.

## Why it cannot be fixed by tightening the validator

The obvious fix, making `raft.ValidateManifestPath` (`internal/cluster/raft/path_validation.go:66`) delegate to `storage.ValidateKey`, is **not safe**. It runs inside FSM `Apply`, which includes log replay, so tightening it would make a node reject an entry an older binary accepted. Two nodes on different versions would then build different state from the same log. That is worse than the looseness it would fix, and #743 records the constraint at the function.

So the gap closes at the consumer instead.

## The loops

**Compaction manifest recovery.** `internal/compaction/manifest.go:309` calls `Delete`, and on failure `:311` calls `Exists` to decide whether the file was already gone. Both now return the same permanent error for a bad key, so the "already deleted, continue" branch cannot be taken, `recoverManifest` errors, the manifest is never deleted, and `GetFilesInManifests` keeps excluding those partitions from compaction indefinitely.

**Reconciliation sweep.** `internal/reconciliation/sweep_manifest.go:235` and `:263`. `:113-121` explicitly treats every `Exists` failure as transient, so an entry with an unusable path becomes permanently unsweepable.

**Cluster file replication.** `internal/cluster/filereplication/puller.go:890` (`StatFile`, the first call every entry hits), `:1083` (`WriteReader`, after the peer body is already streaming), `:1125` (`Delete` of a corrupt partial, so the partial survives and the next attempt resumes from it), and `:1096`/`:1104` where a `statErr` returns `(0, nil)`, making an invalid key indistinguishable from "no partial". `internal/cluster/coordinator.go:741` and `:1757`/`:1824` have the same shape.

**Compaction job.** `internal/compaction/job.go:605-618`: `ReadTo` fails, the recovery calls `Exists` and is gated on `checkErr == nil`, so the "already compacted, skipping" escape hatch is unreachable and a skippable input becomes a hard job failure.

## Shape of the fix

`errors.Is(err, storage.ErrInvalidPath)` at each site, and on a match record the entry as permanently unprocessable (log once, count a metric, drop it from the work set) rather than returning it as a retryable failure.

[...]

PR #773, opened 2026-09-13
fix(tiering): quarantine a permanently unusable storage key instead of retrying it every cycle (#758)

### LINKED ISSUE #758: Tiering migration retries an unusable storage key every cycle with no backoff or attempt cap
Follow-up from #747, which closed this shape in compaction, reconciliation, replication and edge-sync export but left tiering.

`internal/tiering/scheduler.go:129` `runMigration` is cron-driven. Candidates come from `metadata.GetFilesOlderThan`, recomputed every cycle, with **no failure backoff, no skip list and no attempt cap**.

`MigrateFile` to `copyFileStreaming` calls `src.ReadTo` (`migrator.go:369`) and `dst.WriteReader` (`:376`). A key `storage.ValidateKey` refuses returns `ErrInvalidPath` there, wrapped with `%w` so `errors.Is` survives. The migration counts an error, the candidate stays at its old tier, and it is re-selected next cycle, forever. Each cycle also writes a failed-migration row via `CompleteMigration`.

The provenance chain is closed and does not need an attacker: `manager.go:470` `ScanAndRegisterFiles` populates the SQLite file index straight from `objectLister.ListObjects(ctx, "")`, and local listings do not filter unusable keys (#756), so a file a pre-#741 Arc wrote under a folded name is indexed and becomes a migration candidate.

Lower severity in the same file: `migrator.go:425` `hotBackend.Exists` in `reconcileOrphanedFiles` counts an error and continues every cycle, bounded only by the 48-hour reconcile window.

## The fix

The shape #747 settled on: `errors.Is(err, storage.ErrInvalidPath)` at the site that first observes it, then log once at Error, count `metrics.Get().IncStorageInvalidPathQuarantined()`, and drop the candidate from the work set rather than returning it as a retryable failure. Dropping needs a little care here because the work set is recomputed from SQLite each cycle, so "dropped" has to be persisted, most likely as a terminal state on the file index row rather than an in-memory skip list.

Worth pairing with a test that a **transient** migration failure still retries, which is the over-correction #747 guarded against at every site it touched.

xe-nvdk:
Closes #758. Follow-up to #747, which closed this shape in compaction, reconciliation, replication and edge-sync export and left tiering.

## Problem

The tiering migration is cron-driven and rebuilds its candidate set from the SQLite file index every cycle, with no backoff, skip list or attempt cap. A hot file whose key the storage contract refuses (indexed by the pre-migration scan, since local listings do not filter such keys) failed `copyFileStreaming` with the same permanent `storage.ErrInvalidPath` every night, stayed in the hot tier, was re-selected the next night, and wrote a fresh failed-migration row each time. `ReconcileOrphanedFiles` had the smaller loop: a cold row with such a key failed `hotBackend.Exists` on every pass for the 48-hour window.

## Fix

- Both sites branch on `errors.Is(err, storage.ErrInvalidPath)`, log once at Error, count `arc_storage_invalid_path_quarantined_total`, and **quarantine the file index row**.
- The quarantine is persisted (`tier_files.quarantined_at`, `quarantine_reason`; added via tolerant `ALTER TABLE` on an existing database) because the work set is recomputed from the table every cycle. `GetFilesOlderThan` and `GetRecentlyMigratedFiles` exclude quarantined rows; every other read keeps returning them, since on local storage the file is a real data file in the partition glob and the query path still serves it. The row keeps its tier and the file is not deleted.
- `RecordFile`'s upsert never touches the new columns, so the nightly rescan cannot clear the mark. Nothing clears it on purpose: the remedy is renaming the object, which registers as a fresh row.
- `copyFileStreaming` reports the permanent error whichever side of the pipe fails first; the goroutine ordering is a race and classification must not depend on it.
- `MigrateFile` returns `ErrCandidateQuarantined` (wrapping the cause) so `MigrateBatch` can skip its duplicate Error line while still counting the cycle's one failure. Locals named `errors` in the migrator are renamed so the package can be imported.
- Tiering status reports `quarantined_files`; the files endpoint shows the two new fields.

## Tests

- Migration: unusable key is a candidate, first cycle reports one failure and quarantines, second cycle selects nothing, exactly one history row, rescan keeps the mark, metric increments.
- Reconciliation: same shape on a cold row.
- Over-correction guards at both sites: a transient `ReadTo` / `Exists` failure still retries, does not quarantine, does not move the counter, and migrates once the blip clears.
- `copyFileStreaming` classification in all four side/order combinations plus transient-both.
- Schema upgrade from a pre-#758 `tier_files` table, opened twice.
- `QuarantineFile` idempotence, unknown-path error, and which reads do and do not filter.
- Each quarantine test was verified to fail with its branch neutralised. The tiering mock backend now enforces `storage.ValidateKey` like the production backends so these branches are not dead code under test.

Release notes for 26.09.2 updated.

https://claude.ai/code/session_01LsoYkWgjiDoZbWtPsKixWe

---
_Generated by [Claude Code](https://claude.ai/code/session_01LsoYkWgjiDoZbWtPsKixWe)_


xe-nvdk:

Issue #744, opened 2026-09-12
Remaining many-to-one path mappings outside LocalBackend.validatePath


xe-nvdk:
Follow-up from #741. That change removed the `..` fold from `LocalBackend`, which was the root of #574 and #737. Three mappings of the same shape remain elsewhere, found while auditing it.

**1. `.part` suffix collision.** `internal/storage/local.go:155-157` builds the staging name as `fullPath + ".part"`. A caller writing the key `x.parquet.part` and a caller writing `x.parquet` use the same temporary location, so two concurrent writes can corrupt each other. `internal/edgesync/receive.go:502` adds a second `.part` layer on top. Nothing reserves the suffix.

**2. Compaction partition fold.** `internal/compaction/manifest.go:81`:

```go
sanitizedPartition := strings.ReplaceAll(partitionPath, "/", "_")
```

Exactly the fold class #741 removed: `a/b` and `a_b` produce one manifest name. A `jobID` suffix makes a collision unlikely rather than impossible.

**3. `SanitizeS3Prefix`.** `internal/storage/s3.go:579-604` rewrites rather than rejects, and returns `""` when it cannot, so a misconfigured prefix silently becomes the bucket root. Tracked with the backend-contract issue as well since they touch the same file.

None is reachable from validated input today, which is why they are a follow-up rather than part of #741. They are recorded because the pattern is what produced two security reports, and the next one will come from whichever of these a caller reaches first.

PR #775, opened 2026-09-13
fix(retention): report unusable storage files

### LINKED ISSUE #771: Retention's skipped-file counter has been dead since #744 — feed it from ListUnusable
Split out of #762 (item 2), sized for a first contribution.

## Problem

`internal/api/retention.go:91-94` documents `SkippedFiles` / `SkippedReason` on `ExecuteRetentionResponse` as existing "so a policy that can never fully apply does not keep recording clean runs" (added in #746). They are fed from `readParquetPath` failing on a key returned by `h.storage.List` (lines 865, 904, 926).

Since #744, `List` never returns a key that `readParquetPath` would refuse. The counter cannot fire, and retention records clean runs over measurements it cannot fully process, which is the exact state the field was added to prevent.

## What to change

Give the counter a live source. #765 added `storage.UnusableLister` with `ListUnusable(ctx, prefix)`, which returns exactly what `List` drops. At the two retention listing sites, when the backend implements `UnusableLister`, call `ListUnusable` with the same prefix and count the result into `SkippedFiles`, with a `SkippedReason` naming the class, the way `internal/backup/backup.go` `findUnaddressable` (lines 362-397) does for backups. Keep the existing `readParquetPath` feed; it becomes the second line rather than the first.

## Worth knowing

- Dry-run must report the same skip count as a real run.
- Do not delete unaddressable files and do not try to repair their keys. Reporting is the whole job.
- Backends that do not implement `UnusableLister` keep today's behaviour.

## Scope

One handler, one package. Regression test: on a `LocalBackend` in a temp dir, plant a file under the policy's prefix with a key `storage.ValidateKey` refuses (write it with `os.WriteFile` to bypass the contract), execute the policy, and assert `skipped_files == 1` with a reason. `go test -tags=duckdb_arrow ./internal/api/ -run Retention`. Release-notes entry per CONTRIBUTING.

efegokdemir:
## Summary

- feed retention's skipped-file reporting from `storage.UnusableLister`
- count files hidden by normal storage listings without deleting or repairing them
- preserve the existing `readParquetPath` skip handling as a fallback
- report the same skipped-file state for dry-run and real execution
- add regression coverage using a `LocalBackend` with an invalid storage key

## Validation

- `go test -tags=duckdb_arrow ./internal/api/ -run Retention -v`
  - passed
- `go test ./...`
  - passed
- `go vet ./internal/api/`
  - passed
- `go build ./...`
  - passed
- `gofmt -w internal/api/retention.go internal/api/retention_test.go`
  - clean
- `git diff --check`
  - clean

## Scope

This only addresses the retention skipped-file reporting described in #771. Unusable files are not deleted, repaired, or rewritten, and backends without `UnusableLister` retain their previous behaviour.

Closes #771


efegokdemir:

xe-nvdk (changes requested):
Thanks — this is the right idea, the regression test is real (it fails on `main` with `skipped_files = 0, want 1` and passes here), dry-run parity holds, and gofmt, vet and the retention tests with `-race` are all clean on your head. I approved the CI run. One thing has to change before merge, and it drags three smaller asks along with it, because one restructuring fixes all four.

**1. Measurement discovery from unusable keys can abort the whole policy (blocking, regression).** `getMeasurementsToProcess` now adds `parts[0]` of every unusable key as a measurement. If that segment is itself unusable, `deleteOldFiles` lists `database/<segment>/`, the backend refuses the prefix, and `ExecutePolicy` aborts on the error. Reproduced on your branch with one good measurement plus `testdb/bad\meas/2020/01/01/00/x.parquet`:

```
retention aborted for policy 1: failed to list files: invalid prefix: storage: invalid path: "testdb/bad\\meas" contains a backslash, which Azure treats as a separator
```

Before this PR that file was invisible and retention ran. After it, retention never runs again for that database, and the error doesn't say which file caused it. A measurement name derived from an unusable key must pass `storage.ValidateKeySegment` (and not be dot-prefixed) before it becomes a listing prefix; files under a name that fails are counted as skipped at the database level and logged with their key, since they can never be listed.

**2. Count data files only.** `ListUnusable` returns everything a listing hides, including OS debris like `.DS_Store`. Backup filters what it counts (`findUnaddressable` keeps `isBackupPayload` paths); retention should count `.parquet` entries only. Add a case with a `.DS_Store` planted under the measurement that asserts it is not counted.

**3. "see the log for the keys" is no longer true.** The keys from `ListUnusable` are never logged; only the old `readParquetPath` fallback logs one. Log a bounded sample with each object's `Err` (the reason the listing hid it), cap 32 like backup's `sampleUnaddressable`, at the same Error line that reports the count.

**4. List once per policy, not once per database plus once per measurement.** The PR calls `ListUnusable` over `database/` in discovery and again over `database/measurement/` for every measurement. On S3 and Azure that is an extra full listing pass plus a per-measurement pass on every run (3am daily by default); on local it is an extra directory walk per measurement. Do it once: in `getMeasurementsToProcess`, bucket the unusable entries by measurement segment and hand the per-measurement count and sample to the loop, so `deleteOldFiles` never lists unusable objects itself. A policy pinned to one measurement lists just that prefix. This is also where item 1 falls out naturally: bucketing is where you decide which names are processable.

Sketch of the shape (names are yours to choose):

```go
type unusableByMeasurement struct {
	counts  map[string]int      // measurement → data files the listing hides
	samples map[string][]string // measurement → up to 32 keys
	orphans int                 // files under a measurement name that cannot be listed at all
}
```

`getMeasurementsToProcess` returns the measurements plus this; the loop adds `counts[m]` to the measurement's skipped total (and `orphans` once), and `deleteOldFiles` keeps only its existing `readParquetPath` fallback.

**5. Rebase and tidy.** The release-notes conflict is positional: two entries landed at the top of `## Bug fixes` after your base. Rebase, keep yours at the top (that section is newest-first, so your placement was right), and drop the blank lines you have above the heading and after the credit line. Worth a sentence saying why the counter existed and why it was dead (since #744 the listing never returns a key `readParquetPath` refuses), so a reader knows what was fixed.

Nits: the test fixture uses `0755`/`0644`; the house style is `0o700`/`0o600`. And the `retentionSkippedReason` constant reads fine, keep it.

Looking forward to the next round — you clearly read the issue and the surrounding code, which is most of the work.

xe-nvdk (approved):
Second pass on 6b14cf7. Everything from round one is in, and done the way I'd have done it. Approving.

- **The abort is gone.** `TestRetentionSkipsInvalidMeasurementSegmentWithoutAborting` is the regression for exactly the failure I reproduced: one file under `bad\meas` next to a healthy measurement, and the policy completes, deletes the good file, leaves the bad one, and reports `skipped_files = 1`. On round-one code that fixture aborted the whole policy.
- **One listing per policy.** Discovery does the single `ListUnusable` pass and buckets by measurement; `deleteOldFiles` keeps only its `readParquetPath` fallback. Pinned policies list just their prefix.
- **Only data files count.** The `.DS_Store` planted in the first test proves it, and the case-insensitive `.parquet` match is fine.
- **The log now backs the message.** A capped sample with each object's reason goes on the same Error line that reports the count, so "see the log for the keys" is true again.
- Release note says why the counter was dead, house-style perms, and gofmt, vet, and the retention family with `-race` are clean on your head.

One thing I did on your branch rather than ask for: #780 merged after your rebase and landed at the same spot in the notes, so the PR showed a conflict. I merged `main` into `fix/retention-unusable-count` (83372c6, a merge commit, no rewrite) keeping your entry on top with #780's below it, and verified the merged tree builds and passes. Pull before you push again.

Nit for a follow-up, not this PR: for a pinned policy `getMeasurementsToProcess` still calls `h.storage.List(ctx, prefix)` and never uses the result, and `deleteOldFiles` lists the same prefix right after. Skipping the List when `pinned` saves a listing per pinned policy per run.

Merges when CI on the merge commit is green. Thank you for a thorough second round.

xe-nvdk:
Merging — thank you @efegokdemir. Two of four in. #776 and #777 are waiting on you whenever you get to them.

efegokdemir:
Thanks for the detailed review and for helping resolve the release-notes conflict. I really appreciate the second-pass feedback — especially the points around bucketing unusable entries and keeping the listing to one pass. Glad we got this one merged. I’ll pick up #776 and #777 next.

Issue #737, opened 2026-09-11
Edge-sync spoke IDs containing an embedded `..` collide with another spoke's local namespace


xe-nvdk:
Reported privately by **[@rexpository](https://github.com/rexpository)** in GHSA-jxr7-m365-vvg7. Moving it here as an ordinary hardening bug; reasoning on that below. The technical finding is accurate and I reproduced both halves of it.

## The bug

`validateSpokeID` (`internal/edgesync/receive.go:617-631`) rejects empty IDs, path separators, NUL, the exact values `.` and `..`, and any leading dot. It does not reject an embedded `..`:

```go
if spokeID == "." || spokeID == ".." || strings.HasPrefix(spokeID, ".") {
```

The local storage backend then applies `sanitizePath` (`internal/storage/local.go:698-709`), which collapses every `..`:

```go
// Replace .. with _ to prevent directory traversal
path = strings.ReplaceAll(path, "..", "_")
```

So two accepted spoke IDs map onto one directory:

| accepted spoke ID | resolved namespace |
|---|---|
| `rocket_01` | `rocket_01/...` |
| `rocket..01` | `rocket_01/...` |

A spoke authenticated as `rocket..01` writes into `rocket_01`'s namespace. The HMAC proves who is asking, and the namespace mapping is what decides where they may write, so the identity check passing is not the issue. The existing comment in `TestReceiver_RejectsMaliciousSpokeIDs` says exactly this, which is why the ID has to be injective as a path component.

## Scope

- **Local filesystem backend only.** S3 and Azure preserve key spelling, so no collision. Worth a separate look to confirm.
- **Requires an admin to create the colliding ID first.** Registration is behind `auth.RequireAdmin` (`internal/api/edgesync_admin.go:64-68`), secrets are generated server-side and shown once, and there is no self-registration path. An attacker cannot mint `rocket..01` on their own.
- **Edge-sync is opt-in.** `edge_sync.enabled` defaults to false and is not in the shipped `arc.toml`.
- The reported proof creates a file at an absent path. It does not demonstrate overwriting an existing one.

## Why this is not being published as a security advisory

[...]

PR #776, opened 2026-09-13
fix(edgesync): reject colon in spoke IDs

### LINKED ISSUE #751: Edge-sync spoke IDs may contain a colon, which the Raft manifest validator refuses
Found while reconciling the path validators for #746.

`edgesync.validateSpokeID` (`internal/edgesync/receive.go:621`) has no character allowlist. It rejects separators, leading dots, `..`, NUL, control characters, whitespace and anything over 128 bytes, but a colon passes:

```
validateSpokeID("rocket:01") = <nil>
validateSpokeID("a:b")       = <nil>
```

The spoke ID is the first path segment of everything that spoke writes into the hub's storage root, so those files get keys like `rocket:01/db/cpu/2026/09/12/13/f.parquet`.

`storage.ValidateKey` accepts that key, and every backend stores it happily. `raft.ValidateManifestPath` (`internal/cluster/raft/path_validation.go:99-103`) **rejects** it with `ErrPathScheme`: it refuses any `:` outside a Windows drive prefix, as an anti-worm-primitive guard against `s3://attacker-bucket/...` shaped manifest paths.

So on a clustered deployment, a spoke whose ID contains a colon writes files that are stored successfully and can never be registered in the cluster manifest.

This is the one axis on which `ValidateManifestPath` is **stricter** than the key contract rather than looser. Everywhere else it is a superset, which is deliberate and documented (it runs inside FSM Apply including log replay, so it cannot be tightened). `TestManifestValidatorAcceptsSupersetOfContract` in `internal/storage/validator_reconciliation_test.go` pins this divergence so it cannot widen silently, and points here.

The fix is at the spoke ID, not in the FSM: `validateSpokeID` should reject a colon for the same reason it already rejects a separator. That is a startup-time rejection for an operator-typed value, which is where #741 and #737 both concluded such rules belong.

efegokdemir:
## Summary

- reject `:` in edge-sync spoke IDs during registration/input validation
- prevent spoke IDs from producing storage keys that the cluster manifest validator cannot register
- preserve all existing spoke ID validation behaviour
- add regression coverage for `rocket:01` and `a:b`
- verify representative currently valid spoke IDs remain accepted

## Validation

- `go test ./internal/edgesync/... ./internal/storage/...`
  - passed
- `go test ./internal/edgesync -run 'TestReceiver_RejectsMaliciousSpokeIDs|TestValidateSpokeIDAcceptsExistingIDs'`
  - passed
- `go test ./...`
  - passed
- `go vet ./internal/edgesync/... ./internal/storage/...`
  - passed
- `go build ./...`
  - passed
- `gofmt -w internal/edgesync/receive.go internal/edgesync/receive_test.go`
  - clean
- `git diff --check`
  - clean

## Scope

This keeps the fix at spoke ID validation time. `raft.ValidateManifestPath` and the storage key contract are unchanged.

Closes #751


xe-nvdk (changes requested):
Thanks — the code change is exactly right and the smallest it could be. Verified on your head: both colon cases fail on `main` (`spoke ID "rocket:01" was accepted`) and pass here, the accepted-IDs guard passes, `ValidateManifestPath` and the key contract are untouched, gofmt and vet are clean, and the edgesync and storage suites pass with `-race`. I approved the CI run. Two things before merge, both in prose rather than code:

**1. The release note has to say what happens to an existing colon ID after upgrade.** `validateSpokeID` runs at eight sites, and three of them matter to an operator who already has such a spoke: the hub refuses every sync from it (`Receiver.Receive` → 400), the spoke process refuses to start (`agent.go` at construction), and the hub's registry flags it on load (`InvalidStoredIDs`). That is the intended "startup-time rejection of an operator-typed value" from #751, but nobody should discover it from a 400. Add to your entry something like:

> A spoke registered with a colon before this version is refused after upgrade: the hub rejects its syncs and the spoke will not start. Re-register it under a new ID; files already written under the old namespace stay where they are.

and a one-line item under `## Upgrade notes` pointing at the entry (that section is a numbered list; see the existing items for the shape).

**2. The comment you updated now contradicts the code.** `validator_reconciliation_test.go` still says "It is reachable: `edgesync.validateSpokeID` accepts `rocket:01`", and this PR makes that false. The divergence between the key contract and the manifest validator is still real and still pinned by that test; what changed is that the spoke-ID path to it is closed. Reword along the lines of: "It was reachable through `edgesync.validateSpokeID` until #776; the key contract itself still accepts a colon, so the divergence is pinned here."

Also move the entry to the top of `## Bug fixes` (that section is newest-first; you put it next to the related #737 entry, which is a reasonable instinct, but consistency wins), and rebase when you do, since two entries landed at the top today.

Nit: the error text reads better as `contains a colon, which the cluster manifest path validator refuses`.

efegokdemir:
Thanks for the detailed review — addressed all requested changes.

documented the upgrade impact for existing colon-containing spoke IDs and added an Upgrade notes entry
moved the bug-fix entry to the top of the section
updated the validator reconciliation comment so it no longer says edgesync.validateSpokeID accepts colon IDs
adjusted the validation error text to match the requested wording
rebased onto current main

Validation:

go test ./...
go vet ./internal/edgesync/... ./internal/storage/...
go build ./...
git diff --check

All pass locally. Thanks again for the review.

xe-nvdk (approved):
Second pass on 0e884f8. All four asks are in: the entry now says what happens to an existing colon ID after upgrade and there is a matching Upgrade-notes item, the `validator_reconciliation_test.go` comment reads in the past tense and stays accurate about what the test still pins, the entry sits at the top of Bug fixes, the error text reads plainly, and the branch is on current `main`. The only code change since round one is that error string. gofmt and vet are clean and the edgesync and storage suites pass with `-race` on this head; CI is green. Approving.

Merges on the maintainer's word. Thank you.

xe-nvdk:
Merging — thank you @efegokdemir. Three of four in; #777 is the last.

PR #777, opened 2026-09-13
fix(tiering): remove redundant metadata mutex

### LINKED ISSUE #344: medium(tiering): MetadataStore global RWMutex serializes reads
metadata.go:23 — Fix: rely on SQLite WAL.

efegokdemir:
## Summary

- remove the redundant global `MetadataStore.mu`
- allow metadata reads and writes to use the thread-safe `*sql.DB` directly
- preserve all existing `tierCacheMu` protection
- keep SQL, schemas, APIs, configuration, and database behaviour unchanged
- retain existing tier-cache invalidation behaviour

## Validation

- `go test -race ./internal/tiering/...`
  - passed
- `go test ./internal/tiering/...`
  - passed
- `go vet ./internal/tiering/...`
  - passed
- `go test ./...`
  - passed
- `go build ./...`
  - passed
- `gofmt -w internal/tiering/metadata.go`
  - clean
- `git diff --check`
  - clean

## Scope

This removes only the redundant `MetadataStore.mu` serialization. The dedicated `tierCacheMu` locking remains unchanged.

Closes #344


xe-nvdk (changes requested):
Thanks — the diff is exactly the shape the issue asked for: `mu` and every lock pair gone, `tierCacheMu` untouched, nothing else changed, gofmt and vet clean, and the suite passes with `-race` on your head. I approved the CI run. Reviewing it properly turned up two things, one of which is mine to own: the issue's premise was wrong, and following it faithfully opened a real race. Both need to land before this merges.

**1. The mutex was not fully redundant: it ordered the tier-cache fill against invalidation (blocking).** Trace `GetTiersForMeasurement` on a cache miss: it runs the `SELECT DISTINCT tier` query, iterates `rows`, and only then takes `tierCacheMu` to store the result. `database/sql` closes the `Rows` and releases the connection the moment `rows.Next()` returns false, which is before that store. So this interleaving is now possible:

1. reader: query returns `{hot}` and the row loop finishes (connection released)
2. writer: `UpdateTier` moves the last hot file to cold and calls `invalidateTierCache` (deletes the entry)
3. reader: stores `{hot}` into the cache with a fresh 30-second TTL

Before this PR the reader held `mu.RLock` across the whole function and the writer held `mu.Lock` across "update + invalidate", so step 2 could not run between 1 and 3. After it, query routing can read a stale tier set for up to `tierCacheTTL` and skip the cold tier for a measurement whose data just moved there. The window is microseconds, but tiering runs on a schedule against live queries, so over an uptime it will happen.

The fix is small and keeps the mutex gone: a generation counter under `tierCacheMu`. `invalidateTierCache` increments it; the miss path reads it (you already hold `tierCacheMu.RLock` at the miss check) and, when storing, skips the store if the generation moved:

```go
// under tierCacheMu
tierCacheGen uint64

func (s *MetadataStore) invalidateTierCache(database, measurement string) {
	s.tierCacheMu.Lock()
	delete(s.tierCache, database+"/"+measurement)
	s.tierCacheGen++
	s.tierCacheMu.Unlock()
}

// miss path: gen captured under the RLock before the query ...
s.tierCacheMu.Lock()
if s.tierCacheGen == gen {
	s.pruneExpiredTierCache(time.Now())
	s.tierCache[cacheKey] = &tierCacheEntry{...}
}
s.tierCacheMu.Unlock()
```

A global generation is fine: an unrelated invalidation during the query just costs one cache fill. Pull the guarded store into a method (`storeTierCacheIfUnchanged(key, tiers, gen)`) and unit-test it directly: a stale `gen` must not store, the current one must. Then add one concurrent test that hammers `GetTiersForMeasurement` while other goroutines `RecordFile` / `UpdateTier` / `DeleteFile` the same key, so `-race` is actually exercising the store: today `internal/tiering` has no concurrent test, which is why the race detector said nothing.

**2. The release note claims a concurrency gain that cannot exist.** The store's `*sql.DB` is the shared handle from `sharedSQLiteHandle(authManager, cfg.Auth.DBPath)` (`cmd/arc/main.go:3424-3425`), and both openers set `SetMaxOpenConns(1)` (`internal/auth/auth.go:206`, `cmd/arc/main.go:3804`). Every metadata query already queues on that one connection; the mutex was duplicating the pool's serialization, not adding to it. Reword the entry to what actually changed: a lock that duplicated the connection pool's own serialization is gone, and the tier-cache fill is now ordered against invalidation by a generation counter. Please don't say reads no longer serialize; #344's "rely on SQLite WAL" premise assumed a pool larger than one, and that is on me, not you.

Also move the entry to the top of `## Bug fixes` (newest-first) and rebase; several entries landed above it today.

For the record, what was verified: `mu` has no other role (every removed critical section touched only `s.db`; `tierCache` was already under `tierCacheMu`), the `sync` import is still needed, and the `CleanupOldMigrations` comment update is accurate.

efegokdemir:
Thanks for the detailed review — fixed the cache fill/invalidation race with the generation guard, added direct stale/current generation coverage plus a concurrent regression test, and updated the release note to reflect the actual SQLite connection-pool serialization. Also rebased onto current main.

xe-nvdk (approved):
Second pass on 0f18bd5. The race is closed the way it should be, and the note now says what actually changed. Approving.

- **Generation guard:** captured under the same `RLock` as the miss check, bumped in `invalidateTierCache`, and `storeTierCacheIfUnchanged` drops the fill when it moved. `TestMetadataStore_StoreTierCacheIfUnchanged` pins both directions.
- **Concurrent test:** four goroutines per round across forty rounds, with `SetMaxOpenConns(1)` to mirror the shared handle in production, so `-race` is now exercising the store for real. I ran the tiering suite three times under `-race`: clean. gofmt and vet clean.
- **Release note** describes a lock that duplicated the pool's serialization and the new ordering guarantee, with no throughput claim. Good.

One thing I did on your branch: `main` moved again after your rebase (#775 and #776 landed at the top of Bug fixes), so the PR showed a conflict. I merged `main` into `fix/tiering-remove-global-mutex` (7c45f37, a merge commit, no rewrite), kept your entry on top, and verified the tiering suite on the merged tree. Pull before pushing again.

Nit, not blocking: a one-line comment on the `tierCacheGen` field (guarded by `tierCacheMu`; bumped on invalidation so a fill that raced one is dropped) would save the next reader the trace.

Merges when CI on the merge commit is green. Thank you for taking the correction on the issue in stride and turning it into the right fix.

PR #780, opened 2026-09-13
refactor(sql): centralize DuckDB path quoting

### LINKED ISSUE #752: Four spellings of the DuckDB path escaper, while quotePath claims to be the only one
Found while auditing the read path for #746.

`internal/api/query.go:2946` `quotePath` says of itself:

> This is the single source of truth for path interpolation in the query layer — every `read_parquet('` site goes through this helper to close the SQL-injection vector via the x-arc-database header / measurement names / manifest entries (review/query-path-criticals C2).

It is not. There are four spellings of the same escape:

1. `quotePath` (`internal/api/query.go:2946`), `"'" + sqlutil.EscapeStringLiteral(path) + "'"`
2. `escapeDuckDBPath` (`internal/api/delete.go:49`), `strings.ReplaceAll(path, "'", "''")`, used at `delete.go:668`, `:702`, `:785` and `:861` (and inline in the path list at `:596`)
3. inline `strings.ReplaceAll(filePath, "'", "''")` at `internal/api/retention.go:989`
4. inline `strings.ReplaceAll(path, "'", "''")` at `internal/query/parallel_executor.go:199`

They agree today, so this is not a live injection bug. It is the same defect #746 was filed for, one layer over: a claim that one function owns a property, with copies of that function elsewhere that nothing keeps in step. The copies are how #258 introduced the retention prefix bug (it updated one of two copies of a URI builder), and the lesson from #743 was that "a property of the type" is not achieved by adding another name for it.

Worth doing: route 2-4 through `quotePath`/`EscapeStringLiteral`, or move the helper into `sqlutil` and delete the rest, so the doc comment becomes true.

Related follow-ups from the same audit:
- `internal/compaction/job.go:379` builds a storage **key** with `filepath.Join`, which cleans `..` and emits backslashes on Windows. That is the rewrite-instead-of-reject shape #741 removed from `LocalBackend`, reintroduced on the key-construction side.
- `internal/arcxrouter/router.go:255-276` and `internal/api/delete.go:963` do longhand local path joins with a hand-rolled containment check, where `LocalBackend` now guarantees containment by construction.

efegokdemir:
## Summary

- add a shared `sqlutil.QuoteStringLiteral` helper for DuckDB string-literal quoting
- route query, delete, retention, and parallel executor path interpolation through the shared helper
- remove the redundant `escapeDuckDBPath` helper and inline escaping copies
- preserve the existing escaping and SQL behavior
- add regression coverage for single quotes, double quotes, backslashes, `%`, and non-ASCII input

## Validation

- `go test -tags=duckdb_arrow ./internal/api/ ./internal/query/ ./internal/sql/`
  - passed
- `go test ./...`
  - passed
- `go vet ./internal/api/ ./internal/query/ ./internal/sql/`
  - passed
- `go build ./...`
  - passed
- `gofmt` on changed Go files
  - clean
- `git diff --check`
  - clean

## Scope

This is a behavior-preserving consolidation only.

The related `filepath.Join` and local path-joining follow-ups mentioned in #752 are intentionally left out of scope.

Closes #752


xe-nvdk (approved):
Approving. This is the consolidation #752 asked for, done cleanly:

- All four spellings now go through `sqlutil.QuoteStringLiteral`; `escapeDuckDBPath` and both inline copies are gone, and `quotePath` is a thin alias so `query.go`'s call sites stay put. I grepped for any remaining hand-built `read_parquet('` in non-test code: none, only comments.
- Output is byte-identical at every site (`'%s'` → `%s` with the quoted literal), so this is behaviour-preserving, and the `TO '%s'` targets in the two COPY statements are covered too.
- The SECURITY rationale moved from `quotePath` to the helper's comment in `mask.go`, which is the right home now that the helper is the thing every site calls.
- Verified on your head: `go build ./cmd/... ./internal/...`, gofmt and vet clean, and `internal/sql`, `internal/query`, and the delete/retention/query-path tests in `internal/api` pass with `-race`. Release note is at the top of Bug fixes with the credit line, and the branch is on current `main`. I approved the CI run; this merges when it is green.

Out of scope here, filed separately: two more copies of the same escape exist outside the query layer, `internal/compaction/job.go` (`escapeSQLString`, `escapeSQLPath`) and `internal/database/duckdb.go` (`escapeSQLString`). The compaction one also doubles backslashes, which DuckDB's standard string literals do not un-escape, so it has already drifted from the rule everything else uses. Same shape as this PR, if you want a second one.

Nit, not blocking: an empty-string case in `TestQuoteStringLiteral` (`""` → `''`) would pin the degenerate input.

Thank you — four PRs from the label queue in one afternoon, each scoped exactly to its issue, is the way to do it.

xe-nvdk:
Merging — thank you @efegokdemir. First of your four in; README credit follows in a moment. #781 is the natural next one if you want it.

PR #782, opened 2026-09-13
fix(sql): centralize remaining DuckDB string escaping

### LINKED ISSUE #781: Two more copies of the DuckDB string-literal escape outside the query layer; the compaction one has already drifted
Follow-up to #752 / #780, which consolidated the four query-layer spellings into `sqlutil.QuoteStringLiteral` (`internal/sql/mask.go`). Two copies remain outside that layer:

- `internal/compaction/job.go` — `escapeSQLString` (single-quote doubling, identical to the helper) and `escapeSQLPath`, which **also doubles backslashes**:

  ```go
  path = strings.ReplaceAll(path, "\\", "\\\\")
  path = strings.ReplaceAll(path, "'", "''")
  ```

  DuckDB's standard string literals do not un-escape backslashes (only `E'...'` strings do), so `'C:\\data'` is the seven-character string `C:\\data`, not `C:\data`. This copy has already drifted from the rule every other site uses, which is the failure mode #752 was filed for. It is mostly latent because the key contract now refuses backslashes in keys (#741/#743), but a local path with one would be mis-quoted by compaction and quoted correctly by the query layer.
- `internal/database/duckdb.go` — `escapeSQLString`, identical to the helper, used for configuration values.

**What to change:** route both files through `sqlutil.QuoteStringLiteral` / `sqlutil.EscapeStringLiteral` and delete the local copies. Decide explicitly what to do about the backslash doubling in `escapeSQLPath`: the query layer does not do it and is correct, so dropping it aligns compaction with everything else. If any input reaching `escapeSQLPath` can legitimately contain a backslash on a supported platform, say so in the PR and add a test for it; otherwise pin the new behaviour with a case like `` `back\slash` `` → `` `'back\slash'` `` (the same case #780's test uses).

**Worth knowing:** `internal/sql` must stay import-cycle-free from `internal/compaction` and `internal/database`; check `go build ./...` after wiring. Behaviour must not change for any path that has no backslash. Same shape as #780, which is the model to copy.

**Scope:** two files plus tests. `go test -race ./internal/compaction/ ./internal/database/ ./internal/sql/`. Release-notes entry per CONTRIBUTING.

efegokdemir:
## Summary

- route the remaining compaction and database DuckDB string escaping through the shared `internal/sql` helpers
- stop compaction from incorrectly doubling ordinary backslashes in standard DuckDB string literals
- use `sqlutil.QuoteStringLiteral` for compaction path literals
- keep the existing package-level helper names as thin delegating wrappers where other same-package callers still depend on them, avoiding unrelated call-site churn
- add regression coverage for single quotes and `back\slash`

A local path reaching compaction can legitimately contain a backslash on supported Unix-like systems, where `\` is a valid filename character. DuckDB standard string literals do not unescape ordinary backslashes, so doubling them changes the path rather than escaping it.

## Validation

- `go test -race ./internal/compaction/ ./internal/database/ ./internal/sql/`
- `go test ./internal/compaction/ ./internal/database/ ./internal/sql/`
- `go vet ./internal/compaction/... ./internal/database/... ./internal/sql/...`
- `go test ./...`
- `go build ./...`
- `git diff --check`

All pass locally.

No import cycle was introduced.

Closes #781


xe-nvdk (approved):
Approving. Same shape as #780, and the one behaviour change it makes is the right one.

- Both packages now delegate to `sqlutil.EscapeStringLiteral`, and no hand-rolled `ReplaceAll(…, "'", "''")` remains anywhere in `internal/` outside the helper. Keeping `escapeSQLPath` / `escapeSQLString` as thin wrappers is the right call: they still have dozens of callers across `dedup.go`, `subprocess.go`, `duckdb.go`, `sandbox.go` and the tests, and rewriting those would be churn without benefit.
- The backslash decision is correct, and I checked it against DuckDB rather than taking it on faith: `SELECT 'a\b', length('a\b'), 'a\\b', length('a\\b')` gives `a\b`, 3, `a\\b`, 4. The old doubling changed the path; the query layer never did it. The new `back\slash` case fails on `main` (`got "back\\\\slash"`) and passes here.
- `go build ./cmd/... ./internal/...`, gofmt and vet on the touched files, and the sql, database and compaction suites under `-race` are all clean. CI green.

One thing I did on your branch: `main` moved after your base, so the note conflicted positionally. I merged `main` into `fix/duckdb-remaining-string-escaping` (b6920b9, a merge commit, no rewrite), kept your entry on top, verified the merged tree, and pushed. Pull before pushing again.

Nit, not blocking: the two wrappers lost their doc comments. A one-liner each ("delegates to sqlutil.EscapeStringLiteral; kept for the package's existing call sites") would save the next reader a grep.

Thanks — that closes #752 and #781 together, and the escaping rule now has exactly one home.

xe-nvdk:
Merging — thank you @efegokdemir. The escaping rule now has exactly one home.

PR #783, opened 2026-09-13
fix(edgesync): bound source paths to storage budget

### LINKED ISSUE #757: Edge-sync hub accepts a source path it can never store, then answers 503 forever
Found while working #747. This one is **live, network-reachable, and needs no upgrade**.

`internal/edgesync/receive.go:716` bounds `sourcePath` at `MaxKeyLen - len(PartSuffix)`, with a comment saying the headroom is for the staging suffix. But the key the hub actually stores is namespaced first: `NamespacedPath` (`:602`) prepends `spokeID + "/"` (up to 129 bytes) and `stagingPathFor` (`:606`) prepends `".sync-staging/" + spokeID + "/"` (up to 143).

So the bound leaves headroom for `.part` and none for the prefixes the key actually gets. A path the validator accepts can be up to 143 bytes too long by the time it reaches storage.

Reproduced end to end against a real `LocalBackend` and the real `Receiver`:

```
sourcePath len=1014, validateSyncPath=<nil>          // accepted
Receive err = edgesync: receive failed for a hub-side reason:
              stat "...": invalid path: key is 1143 bytes, over the limit
errors.Is(err, ErrReceiveInternal)     = true
errors.Is(err, storage.ErrInvalidPath) = true
```

`ErrReceiveInternal` maps to a **503** at `internal/api/edgesync.go:311`, which is the one answer that explicitly tells the spoke to retry (`receive.go:30-33` documents that intent). With a 128-byte spoke ID the window is every path over roughly 881 bytes.

The spoke side is attempt-capped, so it ends in `failed` rather than looping literally forever, but each attempt logs the full oversized path at Error on the hub, and `RequeueFailed` restarts the same doomed loop.

## Why this is filed separately from #747

#747 is about consumers that meet a permanent error and should stop retrying. This is the opposite direction: the hub should **not accept** a path it cannot store. The fix is to bound `sourcePath` by the budget the namespaced and staging keys actually consume, and to answer a 4xx rather than a 503, so the spoke stops rather than retries. Quarantining at the storage call would paper over an input validator that is simply measuring the wrong string.

Related: `storage.MaxUsableKeyLen` and `MaxUsableKeySegmentLen` were exported in #747 for exactly this class of mistake, and are the right numbers to bound against here.

efegokdemir:
## Summary

- calculate the edge-sync source-path limit from the actual receiving spoke ID
- account for both namespaced and staging prefixes before accepting a path
- use `storage.MaxUsableKeyLen` and `storage.MaxUsableKeySegmentLen`
- reject paths that cannot be stored before reaching the backend
- keep those validation failures as client errors instead of retryable 503 responses
- apply the same spoke-aware validation during reconcile

## Regression coverage

- exact maximum accepted source path is validated against both final stored key forms
- Linux exercises the exact boundary through a real `LocalBackend` and verifies the file is committed
- macOS keeps the exact validator boundary assertions but skips only the physical max-length write because the temporary root plus key can exceed Darwin's absolute filesystem path limit
- one byte over the calculated budget is rejected before storage
- the oversized-path API case returns 400 Bad Request, not 503
- existing valid edge-sync upload behaviour remains covered

## Validation

- `go test ./internal/edgesync/... ./internal/api/...`
- `go test -race ./internal/edgesync/... ./internal/api/...`
- `go vet ./internal/edgesync/... ./internal/api/...`
- `go test ./...`
- `go build ./...`
- `git diff --check`

All pass locally.

Closes #757


xe-nvdk (approved):
Approving. The budget is computed from the right strings and the fix lands where #757 said it should, in the validator rather than at the storage call.

- `sourcePathMaxLen` subtracts the larger of the two prefixes the key actually gets (`.sync-staging/<spoke>/` always wins) from `MaxUsableKeyLen`, which already reserves the `.part` headroom, and segments are bounded by `MaxUsableKeySegmentLen`. `Receive` and `Reconcile` both use the spoke-aware budget; the generic `validateSyncPath` keeps the storage budget for the spoke-side callers, which is right since a spoke does not know the hub's staging layout.
- The rejection stays a plain validation error, so the handler's existing mapping answers 400. I ran the new API test against pre-fix `edgesync`: `status = 503, want 400; body=map[error:hub temporarily unable to accept this file reason:hub_unavailable]`, which is the exact loop the issue described, and it passes here.
- The receiver test pins the boundary from both sides (both stored key forms pass `ValidateKey` at the maximum, one byte over is refused and is not `ErrReceiveInternal`), and on Linux it commits the maximum-length path through a real `LocalBackend`. The darwin skip is only the physical write, for the absolute-path limit, and CI is Linux, so the boundary is exercised where it matters.
- Verified on your head: `go build ./cmd/... ./internal/...`, gofmt and vet clean, the edgesync suite and the API edge-sync tests under `-race`. CI green.

One thing I did on your branch: `main` moved after your base, so the note conflicted positionally. I merged `main` into `fix/edgesync-source-path-budget` (9f498ce, a merge commit, no rewrite), kept your entry on top, verified the merged tree, and pushed. Pull before pushing again.

A thought, not an ask: the spoke's discovery filter (`agent.go`, `validateSyncPath(p) == nil`) still uses the storage budget, so a spoke will still send a path the hub is now going to refuse. That is the better outcome of the two: the spoke gets a 400 with the reason and marks the file failed, which is visible, whereas filtering it out on the spoke would make it silently never sync. Worth a comment there someday saying so.

Thank you — the boundary test with both key forms is exactly what this needed.

xe-nvdk:
Merging — thank you @efegokdemir. The boundary test with both key forms was exactly what this needed.

PR #785, opened 2026-09-13
fix(iceberg): refuse incomplete exports

### LINKED ISSUE #760: Iceberg export silently omits files the listing hides, so the exported table answers queries wrong
Found while working #756.

`internal/iceberg/source.go:138` reads:

```go
uri, err := s.resolver.Resolve(p)
if err != nil {
    // The listing produced a key this backend would itself refuse.
    // Fail rather than skip: a skipped file is missing from the
    // exported table with no other signal.
    return nil, nil, fmt.Errorf("listed file %q has no usable storage path: %w", p, err)
}
```

I wrote that in #746. It is now unreachable. `FilesAndLocal` gets its input from `s.backend.List(ctx, prefix)` (`source.go:126`), and since #744 (`dca623f`) `List` silently drops every key that fails the contract. So a key that would have triggered that error never arrives, and the branch it guards cannot fire.

The consequence is the one the comment names, now with nothing to catch it: the file is **silently absent from the exported Iceberg table**, and the export reports success.

This is worse in kind than the backup case #756 is about. A backup that is missing a file is discovered at restore. An Iceberg table that is missing a file **answers queries wrong, now**, to every engine that reads it (Spark, Trino, DuckDB), with no indication that anything is missing. Arc's own query path still returns those rows because `read_parquet` globs the filesystem, so the same measurement gives different answers through Arc and through the Iceberg copy.

Reproducing it needs only what #756 documents: a file whose key fails the contract but which holds real rows, for example a backslash in the filename (a legal POSIX filename) or a whole key over the 1019-byte bound built by nesting. Both are what a pre-#741 Arc wrote, so this is upgrade-supplied rather than hypothetical.

The fix wants the enumeration #756 adds (`storage.UnusableLister.ListUnusable`): the exporter should ask for the hidden set under the prefix it is exporting and refuse to publish a table it knows is incomplete, rather than publishing one that is silently wrong. Failing the export for that measurement is the behaviour the existing comment already argues for.

Related: #756 (the same silence in backup, where it is fixed), #757, #758, #759.

efegokdemir:
## Summary

- make Iceberg export consult `storage.UnusableLister` before normal listing
- refuse to publish a measurement when a hidden/unusable Parquet data file would make the exported table incomplete
- preserve existing behavior for backends that do not implement `UnusableLister`
- keep irrelevant hidden files excluded by the existing Iceberg data-file predicate
- preserve the hidden-first enumeration ordering established by the storage listing fixes

## Regression coverage

- uses the real `LocalBackend`
- proves a backslash-containing Parquet file is hidden from normal `List`
- proves the same file is reported by `ListUnusable`
- verifies Iceberg source enumeration fails instead of returning a partial table
- separately verifies irrelevant hidden files such as `.DS_Store` do not block export

## Validation

- `go test ./internal/iceberg/...`
- `go test -race ./internal/iceberg/...`
- `go vet ./internal/iceberg/...`
- `go test ./...`
- `go build ./...`
- `git diff --check`

All pass locally.

Closes #760


xe-nvdk (changes requested):
Thanks — this is the behaviour #760 asked for, in the right place. I traced the consequences before deciding, since Iceberg is the path where a wrong answer reaches other engines:

- `reconcileOne` returns the refusal, the pass logs `Iceberg reconcile: measurement failed` at Error with the measurement, counts it as failed, and continues with every other measurement. No new snapshot is committed, so the table stays at its last published state instead of being republished without the row set. That is the correct outcome, and it repeats every reconcile pass (5 minutes by default) until the file is renamed, which is what you want for a table that is wrong.
- `TestFiles_RefusesHiddenDataFile` fails on `main` (`Files returned 1 files, want hidden data-file error`) and passes here, and the `.DS_Store` case proves the existing data-file predicate keeps debris from blocking an export. Build, gofmt, vet and the Iceberg suite under `-race` are clean; CI green.

One thing has to change before merge, and one is prose:

**1. The error must name the file (required).** The message says "a data file is hidden from the normal storage listing" and stops. The operator's only job here is to find and rename that file; an Error line every five minutes with no path leaves them grepping the filesystem. `ListUnusable` hands you the path and the reason (`UnusableObject.Err`, which wraps `ErrInvalidPath` and says exactly which rule the key broke). Collect the data-file hits, cap the list (backup uses 32, `unaddressableSampleCap`), and put them in the error:

```
refusing Iceberg export for mydb/cpu: 2 data files are hidden from the normal storage listing and the published table would omit their rows; rename them and the next pass will publish: mydb/cpu/2026/07/14/15/ba\d.parquet (storage: invalid path: contains a backslash …), …
```

Add the path assertion to the test.

**2. Release note.** Say what the operator sees and does: the measurement's export fails and logs at Error on every reconcile pass, its table stays at the last published snapshot, and renaming the named files unblocks it. Two more things worth a clause: Iceberg export is local-only, so this is a second directory walk per measurement per pass, accepted for the same reason backup and retention accepted it; and if that ever shows up in reconcile time, a single walk that returns both listable and hidden objects is the follow-up, not dropping the check.

Also rebase: `main` moved after your base, so the entry conflicts positionally; keep it at the top of `## Bug fixes`.

For the record: hidden-first ordering is right (a rename between the two passes can only cause a spurious refusal, never a silent omission, and the next pass clears it), and backends without `UnusableLister` keep today's behaviour, which today means all three real backends implement it.

xe-nvdk (approved):
Both asks are addressed, and I re-verified the rebased head rather than the diff alone:

- The refusal now carries the count, up to 32 paths, and the storage reason for each (`ba\d.parquet: … invalid path …`), so the Error line the reconciler logs every pass is enough to go and rename the file. `TestFiles_RefusesHiddenDataFile` asserts the path, the reason and the wording, and still fails against `main`'s `source.go` (`Files returned 1 files, want hidden data-file error`); the `.DS_Store` case keeps debris from blocking an export.
- Release note is accurate on all three points (Error on every pass, table stays at the last published snapshot, rename unblocks) and sits at the top of `## Bug fixes` on current `main` with the Contributed-by line.
- Build, vet, `go test -race` for `internal/iceberg` and `internal/ingest` clean; CI green; the three commits touch exactly `source.go`, `source_test.go` and the notes.

Thanks for the careful follow-through on the error text — that is the line an operator will actually read.

PR #786, opened 2026-09-13
fix(compaction): clean exact job temp directory

### LINKED ISSUE #749: Compaction temp-dir cleanup can os.RemoveAll a concurrently running job's working directory
Found while working #744. This outranks everything listed there.

`internal/compaction/manager.go:557`:

```go
partitionPrefix := sanitizeDBForName(candidate.Database) + "_" +
    strings.ReplaceAll(candidate.PartitionPath, "/", "_") + "_"
...
if entry.IsDir() && strings.HasPrefix(entry.Name(), partitionPrefix) {
    os.RemoveAll(dirPath)
}
```

`_` is both the field separator and the `/` replacement, and it is legal in database and measurement names. `PartitionPath` already begins with the database, so the database is folded in twice. Two ordinary names therefore produce one prefix:

| database | measurement | partition path | prefix |
|---|---|---|---|
| `a` | `a_a_cpu` | `a/a_a_cpu/2026/09/12/14` | `a_a_a_a_cpu_2026_09_12_14_` |
| `a_a` | `cpu` | `a_a/cpu/2026/09/12/14` | `a_a_a_a_cpu_2026_09_12_14_` |

No edge-sync pseudo-database is needed; both names pass `isValidDatabaseName` and `isValidMeasurementName`.

## Why it matters

Partitions run `MaxConcurrent` at a time within a tier (`manager.go:929`, `manager.go:1008`). So when one job finishes, this sweep deletes the **other job's live temp directory**, which holds its downloaded source files and its in-progress compacted output. The victim job then fails or produces a truncated result, and the sweep is not guarded by anything: `HasPrefix` on a folded name, then `os.RemoveAll`.

The `_b{batch}` and nanosecond components that make `JobID` unique do not help, because the prefix truncates before them.

## Related, currently defused

A daily prefix such as `prod_prod_cpu_2026_09_12_` is a prefix of every hourly temp directory for that day, so a daily job would sweep all 24 hourly working directories. That one is only defused because tiers run sequentially under `m.cycleRunning` (`manager.go:825`, `manager.go:901`). It is one scheduling change away from live.

## Suggested fix

Stop deriving the sweep key from a folded string. The job already has an unambiguous identity, `JobID`, and the temp directory is created from it. Sweep by exact directory name, or record the created path and remove that, rather than prefix-matching a lossy encoding.

If prefix matching has to stay, the separator must be a byte that cannot appear in the encoded components, and the database must not be folded in twice.

efegokdemir:
## Summary

- replace lossy partition-prefix temp cleanup with exact JobID-owned directory cleanup
- prevent one completed compaction job from deleting another concurrent job's working directory
- reuse the existing JobID validation before removing the temp path
- preserve parent-side cleanup for subprocess crash/OOM cases
- leave startup orphan cleanup behavior unchanged

## Regression coverage

- reproduces the colliding-name shape from #749
- proves two distinct jobs would have shared the old underscore-based cleanup prefix
- verifies cleanup for job A removes only job A's directory
- verifies job B's directory and sentinel remain intact
- covers unsafe/invalid JobID protection

## Validation

- `go test ./internal/compaction/...`
- `go test -race ./internal/compaction/...`
- `go vet ./internal/compaction/...`
- `go test ./...`
- `go build ./...`
- `git diff --check`
- `go build -tags=duckdb_arrow ./...`
- `go vet -tags=duckdb_arrow ./...`

All pass locally.

Closes #749


xe-nvdk (changes requested):
Thanks — this is the right fix for #749, and the test reproduces the issue's exact collision (`a` + `a/a_a_cpu/…` and `a_a` + `a_a/cpu/…` both folding to `a_a_a_a_cpu_2026_09_12_14_`) before proving job A's cleanup leaves job B's directory and sentinel alone. I traced the part the tests cannot: the subprocess really does create its working directory at `filepath.Join(TempDirectory, JobID)` (`job.go:309`), with the DuckDB spill directory beneath it (`subprocess.go:175`), so removing that one path cleans everything the job made, on success and on crash alike. `validateJobID` already refuses separators, `..`, dot-prefixed and empty IDs, and the parent-generated ID (`{db}_{partition}_{unixnano}_b{batch}` with `/` in a pseudo-database folded to `.`) always passes it. Build, gofmt, vet and the compaction suite under `-race` are clean, and CI is green.

One small change before merge, because this is a deletion path:

**The path is now computed in three places that have to agree.** `job.go:309` builds it to create the directory, `subprocess.go:175` builds it for the spill dir, and `cleanupSubprocessTempDir` builds it to delete. Today they are the same expression; the whole fix depends on that staying true, and nothing pins it. Add one helper and use it at all three sites:

```go
// jobTempDir is the working directory a job owns under the compaction temp
// root. Creation, the spill directory and parent-side cleanup all go through
// it so cleanup can never drift from what the subprocess actually made.
func jobTempDir(tempDirectory, jobID string) string {
	return filepath.Join(tempDirectory, jobID)
}
```

and a test asserting that the directory `Job.Run` creates for a given `TempDirectory`/`JobID` is the one `cleanupSubprocessTempDir` removes (create via the helper, clean via the function, assert gone). That is the same lesson as #752: a property that three copies happen to share is not a property until one function owns it.

Two notes, not blocking:
- The old sweep also happened to remove leftover directories of the same partition from earlier runs whose parent-side cleanup had failed (`RemoveAll` returning an error, say `EBUSY`). Those now wait for `CleanupOrphanedTempDirs` at the next boot. That is acceptable, and the issue's other hazard (a daily prefix matching every hourly directory of that day) is gone with the prefix itself; both are worth one sentence in the release note.
- `main` moved after your base, so the note conflicts positionally; rebase when you push the change above and keep your entry at the top of `## Bug fixes`.

xe-nvdk (approved):
The change I asked for is in and it is the right shape: `jobTempDir` is the single owner of the path, and all three sites go through it — creation (`job.go`), the DuckDB spill directory (`subprocess.go`) and parent-side cleanup (`manager.go`), plus the two log lines. `TestCleanupSubprocessTempDirRemovesJobTempDir` pins the coupling (create through the helper, remove through the function), the collision test still proves job A's cleanup leaves job B's directory and sentinel alone, and `TestCleanupSubprocessTempDirRejectsUnsafeJobID` keeps a bad ID from ever resolving to the temp root.

Verified on the head: build, vet and the compaction suite under `-race` are clean (the two gofmt hits in this package are pre-existing files you did not touch), CI green. The release note says what an operator will notice, including that a failed parent cleanup now waits for the startup sweep. I pushed a merge of `main` onto the branch to resolve the positional release-notes conflict (#785 landed first); your entry stays at the top of `## Bug fixes`, nothing else changed.

Nice fix, and thanks for reproducing the exact prefix collision from the issue in the test.

PR #787, opened 2026-09-13
fix(cluster): propagate manifest apply context

### LINKED ISSUE #394: Context plumbing through RegisterFileInManifest / DeleteFileFromManifest
## Problem
`RegisterFileInManifest` and `DeleteFileFromManifest` on the Coordinator don't accept a context parameter. The caller's timeout (e.g. CompletionWatcher's ApplyTimeout) is not respected — the coordinator uses a hardcoded 5s `forwardApplyTimeout` internally.

## Proposed Solution
Update the `bridgeCoordinator` interface to accept `context.Context` on both methods. Propagate through Coordinator, CompactionBridge, file_registrar, and all test stubs.

## Context
Flagged by Gemini on PR #389. Deferred due to broad interface change ripple.

## Effort
Medium — interface change affects coordinator, bridge, file_registrar, and ~10 test stubs.

efegokdemir:
## Summary

- propagate caller `context.Context` through manifest register/delete operations
- add a context-aware batch apply path for the compaction CompletionWatcher flow
- preserve the existing five-second fallback budget for callers without a shorter deadline
- propagate caller deadlines through follower dial, send, and receive
- preserve existing Raft commit-wait semantics
- propagate worker/request contexts from file registration and edge-sync callbacks

## Context

Issue #394 predates the current batched compaction manifest path.

CompletionWatcher now applies each completion manifest through `BatchFileOps`, so fixing only `RegisterFileInManifest` / `DeleteFileFromManifest` would leave the active compaction path without caller-context propagation.

This change therefore adds `BatchFileOpsInManifestContext(ctx, ops)` for that path while leaving unrelated callers on the existing bounded API.

## Behavior

- caller cancellation is checked before manifest mutation
- caller deadlines bound the pre-apply check and leader-side enqueue timeout
- follower dial, send, and receive use the caller-derived deadline
- the Raft `future.Error()` commit wait retains its existing Raft semantics
- coordinator lifecycle cancellation is respected
- standalone behavior and existing Raft error wrapping remain unchanged

## Tests

- register, delete, and batch bridge paths propagate caller context
- a silent follower leader respects a 100ms caller deadline and returns well under one second
- cancelled coordinator lifecycle context returns `context.Canceled` wrapped with `raft.ErrManifestApply`
- updated integration/unit stubs cover the new signatures

## Validation

- `go test ./...`
- `go test -tags=duckdb_arrow ./...`
- `go test -race ./internal/cluster/... ./internal/compaction/...`
- `go vet ./internal/cluster/... ./internal/compaction/...`
- `go vet -tags=duckdb_arrow ./internal/cluster/... ./internal/compaction/...`
- `go build ./...`
- `go build -tags=duckdb_arrow ./...`
- `git diff --check`

All pass locally.

Closes #394


xe-nvdk (changes requested):
Thanks — the plumbing is careful (the lifecycle merge, the cancel discipline, keeping the old `BatchFileOpsInManifest` as a bounded wrapper, and the standalone early-returns are all right), and I put real weight behind checking it: builds under both tag sets, gofmt and vet clean, the cluster and compaction suites under `-race`, a second reviewer working the cluster-operations checklist, and a live three-writer cluster built from this branch, run twice, next to the same run on `main`. One thing has to change before this can close #394, and two are prose.

**What the live cluster showed (enterprise-local compose, this branch vs `main`, debug logging):**
- healthy: a write on a follower is in the leader's manifest within 1s, both builds;
- leader paused (`docker pause`): the follower's write returns in 0.4 ms and its forwarded apply fails at ~5.4 s with `raft: manifest apply failed`, both builds — the old promise that a stuck leader cannot block the writer flush path still holds;
- recovery after unpause: a new leader within 1 s and the next write registered on the new quorum within 1 s, both builds (one PR run took ~30 s to elect; it did not recur, and `main` has the same election shape, so that was Raft, not you).

So there is no regression. There is also less new behaviour than the description says:

**1. The caller's deadline never reaches the forwarded round trip (required).** `forwardApplyToLeader` computes the ctx-derived deadline and sets it on the connection (`forward_apply.go:146-150`), but the next two lines pass the constant `forwardApplyTimeout` to `protocol.SendMessage` (`:152`) and `protocol.ReceiveMessage` (`:160`), and those helpers set their own fresh 5 s write/read deadlines and clear them on return (`protocol/codec.go:214-237`), overwriting yours. Only the dial honours the context (`:261-267`). Since a compactor is typically not the Raft leader (`forward_apply.go:20-22`), this is the exact path #394 is about, and on it the caller deadline currently changes nothing. The fix is two arguments: pass `manifestApplyTimeout(ctx)` to both calls (it is already in scope), and either drop the now-dead `SetDeadline` block or keep it as belt-and-braces. Add a test that proves it: a fake leader that accepts the connection and never acks, a 100 ms caller deadline, and an assertion that the call returns in well under 5 s. There is no test of `forwardApplyToLeader` in the tree today, so this would be the first.

**2. What the Raft timeout bounds (prose, required).** This repo's `Node.Apply` hands the timeout to hashicorp's `Apply`, which bounds only enqueueing onto the apply channel, and then waits on `future.Error()` with no bound (`raft/node.go:380-395`). That wait ends on commit or on leadership loss, not on a clock. That was true before your change, so it is not a regression, but the doc comment on `BatchFileOpsInManifestContext` ("must bound the actual Raft apply") and the release note ("preserve caller deadline budgets", "the five-second cap remains the maximum apply duration") say more than the code does. Please state it precisely: the deadline bounds the pre-apply check, the leader-side enqueue, and the follower's dial and (after item 1) send/receive; the commit wait keeps its Raft semantics. Note also that the CompletionWatcher's `ApplyTimeout` is a hardcoded 5 s (`main.go:1661`), equal to the cap, so today the observable effect is shutdown cancellation and any caller that arrives with a shorter budget.

**3. Small things, please fold in:**
- The early `ctx.Err()` returns in the three methods are not wrapped in `raft.ErrManifestApply`, unlike every other manifest failure; `reconciliation`'s `isRaftError` would misclassify a shutdown-time `Canceled`. `errors.Join(raft.ErrManifestApply, …)` for consistency.
- `getOrDialLeader` (`:263-264`, pre-existing) can hand `net.DialTimeout` a zero or negative remaining, which means no timeout at all; you added the 1 ns floor in `manifestApplyTimeout`, reuse it there.
- One coordinator-level test: a stopped coordinator (`c.ctx` cancelled) → `RegisterFileInManifest` returns `context.Canceled` without touching Raft. The three bridge tests only prove the bridge forwards the context it was given.
- Rebase: `main` moved after your base, so the note conflicts positionally; keep it at the top of `## Bug fixes`.

Verified for the record: `manifestContext` is safe in every mode (`c.ctx` is set in `Start` before any caller is wired; `raftNode == nil` returns before it is touched; every path calls the returned cancel), the FSM register/delete/batch are idempotent so a timed-out-then-committed apply followed by a retry is harmless, and the lifecycle ordering at shutdown stops the registrar and the watcher before the coordinator cancels its context.

xe-nvdk (approved):
All three asks are in, and I checked each the same way as before: builds under both tag sets, gofmt and vet clean, cluster/raft/compaction suites under `-race`, and the live three-writer cluster.

- **Caller deadline now reaches the round trip.** `manifestApplyTimeout(ctx)` goes to `SendMessage` and `ReceiveMessage` as well as the dial, with the 1 ns floor. `TestForwardApplyToLeader_UsesCallerDeadline` (fake leader that never acks, 100 ms budget) is the proof: against `main`'s `forward_apply.go` it fails with `took 5.0022465s; caller's 100ms deadline was not enforced promptly`, on this branch it passes. The `ctx.Err()` early returns are wrapped in `raft.ErrManifestApply`, and `TestRegisterFileInManifest_CancelledCoordinatorContext` covers the stopped coordinator.
- **Prose.** The doc comments and the release note now say exactly what the deadline bounds (pre-apply check, leader-side enqueue, follower dial and send/receive) and that the commit wait keeps its Raft semantics.
- **Live, this branch merged onto current `main`:** healthy write on a follower is in the leader's manifest within 1 s; with the leader paused the follower's write returns in 0.4 ms and the forwarded apply fails at ~5.4 s (the registrar passes its lifecycle context, so the 5 s cap is the operative budget there, as documented); after unpause a fresh write registers on the new quorum within 2 s. Same shape as `main`, no regression. (The ~30 s election after unpause showed up again; it is Raft's, not this change's, and the new leader this time was the reader, which the harness did not expect but the cluster handled.)
- CI's `Build & Test` failure on your head was `TestBufferMetrics_FlushCountersMove`, the order-dependent ingest metrics test that `main` fixed yesterday; your branch predates the fix. I pushed two merges of `main` onto the branch (the release-notes entry conflicted positionally twice as other fixes landed); your entry stays at the top of `## Bug fixes`, no code changed, and CI is re-running on the merge commit.

Thanks for the follow-through; this is now the fix #394 asked for.

PR #788, opened 2026-09-14
fix(mqtt): validate placeholder reservation in startSubscriber before…

### LINKED ISSUE #770: mqtt: startSubscriber installs into the map unconditionally — Delete/Shutdown racing an in-flight start leaks a live subscriber
Found while reviewing #766 (the RestartSubscription TOCTOU fix for #301). Pre-existing; it affects `StartSubscription` on `main` today and `RestartSubscription` after #766 in the same way.

## Problem

`startSubscriber` (`internal/mqtt/manager.go:481`) connects to the broker and then installs the live subscriber into the map unconditionally (`manager.go:495`):

```go
if err := subscriber.Start(); err != nil { return err }
m.mu.Lock()
m.subscribers[sub.ID] = subscriber
m.mu.Unlock()
```

The nil-placeholder protocol reserves the slot before this, but nothing checks the reservation is still there when the install happens. Two operations can remove it in between:

- `Delete` (`manager.go:309`) removes the placeholder and deletes the DB row while `Start()` is still connecting. The install then lands a running subscriber for a subscription that no longer exists. It keeps consuming a broker connection and writing into the arrow buffer until process shutdown; `GetAllStats` never lists it because it iterates DB rows.
- `Shutdown` skips nil placeholders and replaces the map with a fresh one. An in-flight start then installs a live subscriber into the new map after shutdown has "completed", and nothing stops it.

## Suggested fix

Make the install a compare-and-swap on the placeholder:

```go
m.mu.Lock()
cur, reserved := m.subscribers[sub.ID]
if !reserved || cur != nil {
    m.mu.Unlock()
    _ = subscriber.Stop()
    return ErrSubscriptionNotRunning // slot was released (deleted/shut down) during start
}
m.subscribers[sub.ID] = subscriber
m.mu.Unlock()
```

`Start()` on the auto-start path calls `startSubscriber` without reserving a slot first, so it would need to reserve (or the check would need to allow "absent" only for that caller). Regression tests: inject a placeholder, call `startSubscriber` after removing it, assert the new subscriber is stopped and not installed.

Thundercloud12:
Fixes #770

## Problem

`startSubscriber` connects to the MQTT broker outside the manager lock and then unconditionally installs the running subscriber into `m.subscribers`:

```go
if err := subscriber.Start(); err != nil {
    return err
}

m.mu.Lock()
m.subscribers[sub.ID] = subscriber
m.mu.Unlock()
```

While `subscriber.Start()` performs network I/O, other operations can remove or invalidate the slot reservation.

### 1. `Delete(ctx, id)`

`Delete` can remove the `nil` placeholder and the corresponding SQLite record. When `startSubscriber` completes, it would still install the running subscriber into the map.

This creates an **orphaned subscriber** with no corresponding database record. It continues running indefinitely and cannot be stopped through the API.

### 2. `Shutdown(ctx)`

`Shutdown` skips `nil` placeholders and replaces `m.subscribers` with a fresh empty map.

If `startSubscriber` completes after this point, it installs the running subscriber into the newly initialized map of an already shut-down manager.

### 3. Boot-time auto-start

`Start(ctx)` was calling `startSubscriber` without first reserving the subscriber's slot with a `nil` placeholder, making the startup path inconsistent with the normal subscription flow.

### 1. Compare-and-Swap in `startSubscriber`

Before installing the running subscriber, `startSubscriber` now verifies that its reservation is still valid while holding the manager lock:

```go
m.mu.Lock()

cur, reserved := m.subscribers[sub.ID]
if !reserved || cur != nil {
    m.mu.Unlock()
    _ = subscriber.Stop()
    return ErrSubscriptionNotRunning
}

m.subscribers[sub.ID] = subscriber
m.mu.Unlock()
```

This handles the possible race conditions:

* `reserved == false` — the reservation was removed by `Delete` or `Shutdown`.
* `cur != nil` — another subscriber already occupies the slot.
* In either case, the newly started subscriber is immediately stopped to close the broker connection and prevent resource leaks.

### 2. Uniform Reservation in `Start(ctx)`

Boot-time auto-start now reserves the subscriber slot before calling `startSubscriber`:

```go
m.subscribers[sub.ID] = nil
```

If startup fails, the placeholder is cleaned up:

```go
delete(m.subscribers, sub.ID)
```

This keeps the startup path consistent with normal subscriber creation and ensures failed auto-starts do not leave stale reservations.

### 3. Release Notes

Added an entry to `RELEASE_NOTES_2026.09.2.md` referencing #770.

## Tests

Added `newMockMQTTBroker` to `internal/mqtt/manager_test.go`.

The mock broker uses a sub-millisecond in-process TCP listener to handle the required MQTT control packets.

### Added tests

* `TestManager_StartSubscriber_AbortsIfReservationDeleted`

  * Verifies that `startSubscriber` aborts when its reservation has been removed.
  * Verifies that the subscriber is stopped.
  * Verifies that no orphaned subscriber is installed into the map.

* `TestManager_StartSubscriber_AbortsIfReservationReplaced`

  * Verifies that `startSubscriber` aborts when the reserved slot is already occupied.
  * Verifies that the existing subscriber remains unchanged.

* `TestManager_Start_PlaceholderCleanedOnFailure`

  * Verifies that failed boot-time auto-starts release their reserved slot.
  * Verifies that the subscription is recorded with an error status.

## Verification

The following test suites pass successfully:

```bash
go test -race ./internal/mqtt/...
go test ./internal/api/...
```


Thundercloud12:
@xe-nvdk as you told applied the suggested fix also ensured whoever calls start subscriber now follows this updated thing!

xe-nvdk (changes requested):
Thanks — this is the fix #770 asked for, and the in-process broker is a genuinely useful addition: it lets the tests drive a real `Subscriber.Start()` instead of injecting state, and it held up across five `-race` runs of the whole package. Verified on your head: both compare-and-swap tests fail on `main` (`expected ErrSubscriptionNotRunning, got <nil>`) and pass here, gofmt and vet are clean, and CI is green. Two small things before merge, one of them prose:

**1. `Start` now overwrites whatever is in the slot.** The new reservation in the auto-start loop is `m.subscribers[sub.ID] = nil` unconditionally. Today nothing can be there, because the HTTP listener starts long after `mqttManager.Start` returns (`cmd/arc/main.go:1137` vs `:3694`), so this is not reachable. But this PR exists to make "never install over a slot you don't own" a local invariant, and the boot path should honour it too. Reserve only when the key is absent, and skip with a log line otherwise:

```go
m.mu.Lock()
if _, exists := m.subscribers[sub.ID]; exists {
	m.mu.Unlock()
	m.logger.Warn().Str("id", sub.ID).Msg("Auto-start skipped: subscription already starting or running")
	continue
}
m.subscribers[sub.ID] = nil
m.mu.Unlock()
```

A one-line test (pre-install a subscriber, call `Start`, assert it is untouched) pins it. Note `TestManager_Start_PlaceholderCleanedOnFailure` passes on `main` too, since the old `Start` never reserved anything; it is a guard for the cleanup, not a regression test, which is fine, just worth knowing.

**2. Release note.** Two fixes: move the entry to the top of `## Bug fixes` (the section is newest-first; you put it next to #766, which is a reasonable instinct, but consistency wins), and the credit line needs the PR link: `Contributed by [@Thundercloud12](https://github.com/Thundercloud12) in [#788](https://github.com/Basekick-Labs/arc/pull/788).`

For the record, what I traced and am fine with: the CAS abort returning `ErrSubscriptionNotRunning` flows into `StartSubscription`'s existing failure path (placeholder delete is a no-op after `Delete` or `Shutdown` removed it; the status write is harmless), `RestartSubscription`'s deferred cleanup only deletes a nil, and stopping the just-connected subscriber on abort is the right call even though it costs the one-second disconnect quiesce you see in the "replaced" test's timing. The ordering `Delete` → row gone → any later `StartSubscription` fails at `repo.Get` means the "slot re-reserved by someone else while I was connecting" case cannot install two subscribers.

Thanks for picking #770 up so quickly.

xe-nvdk (approved):
Second pass on 733aca4. Both points are in. Approving.

- **Auto-start no longer overwrites a held slot.** It reserves only when the key is absent and logs a Warn otherwise, and `TestManager_Start_SkipsExistingSubscriber` pins it by pre-installing a subscriber and asserting the same pointer survives `Start`. The compare-and-swap in `startSubscriber` is unchanged from the round I traced.
- **Release note** is at the top of Bug fixes with the PR link.
- Verified on your head: gofmt and vet clean, and the mqtt suite three times under `-race` with your in-process broker, all green.

One thing I did on your branch: `main` moved again after your rebase, so the note conflicted positionally. I merged `main` into `fix/mqtt-start-subscriber-cas` (4da8361, a merge commit, no rewrite), kept your entry on top, verified the merged tree, and pushed. Pull before pushing again.

Nit for another day: the new test writes `mgr.subscribers[sub.ID] = existingSub` without taking `mgr.mu`, unlike its neighbours. Harmless in a single-goroutine test, but matching the pattern keeps `-race` meaningful if someone later adds concurrency around it.

Merges when CI on the merge commit is green. Thank you — #766, #767 and now #770 closed in two days, and the mock broker will keep paying for itself.

xe-nvdk:
Merging — thank you @Thundercloud12. Three MQTT fixes in two days, and the in-process broker will keep paying for itself. README credit follows.

Issue #619, opened 2026-08-19
Edge sync: hub-side compaction of spoke namespaces


xe-nvdk:
Follow-up to #610/#618. With defer-until-synced shipped, the hub receives every spoke row exactly once — as raw files, forever. Hub compaction never touches them: spoke data lives at `{spoke_id}/{db}/{meas}/{y}/{m}/{d}/{h}/…`, one level deeper than the `{db}/{meas}/…` layout the tiers' partition scanners expect, so candidate discovery finds zero partitions under a spoke namespace (observed live: `partition_count: 0` for `telemetry/engine_temp` under `rocket-01`). Hub queries over spoke data pay a growing small-files cost.

## Sketch

1. **Namespace-aware candidate discovery**: compaction learns the registered spoke IDs (from the spoke registry) and scans each `{spoke}/{db}` as a pseudo-database, so everything downstream (measurement scan, partitions, jobs, manifests) operates on the standard two-level model with a prefixed database name.
2. **The receipt-index interplay — this is the #611 gate firing**: hub compaction would DELETE received raws under spoke namespaces. The `sync_received` index must not "forget" them (`confirmPresent` checks backend existence and forgets missing files — a spoke whose pruned-then-rediscovered raw is re-offered would then re-upload it, reintroducing the duplicate next to the compacted file). Receipts for compacted-away inputs must be marked (content still delivered, file merged) and treated as `present` without an existence check; re-delivery of the same path+sha stays a no-op. Requires the compacted-INPUT list to reach the parent process (currently only the output key crosses the subprocess IPC).
3. Cluster mode: compacted outputs/source deletes flow through the existing Phase-4 completion-manifest machinery (path-based) — verify with the namespace prefix.

Treat #611 (`HubIndex.Forget` gate) as a blocker on this work per that issue.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #789, opened 2026-09-14
fix(compaction): list each measurement once in the candidates preview

### LINKED ISSUE #316: medium(compaction): Redundant file listing when both tiers enabled
hourly.go:157, daily.go:161 — same listing twice. Fix: pre-fetch once.

alexeymoskalev-devops:
Closes #316

## Summary

- `HourlyTier` and `DailyTier` each listed `database/measurement/` inside `FindCandidates`, so `Manager.FindCandidates` — the candidates preview behind `GET /api/v1/compaction/candidates` — listed every measurement once per enabled tier on every call.
- `Tier` gains `FindCandidatesFromListing(database, measurement, objects) []Candidate`: the same scan over a listing the caller already holds, no I/O. `FindCandidates` is now `listObjects` + `FindCandidatesFromListing`. The grouping loops (`groupHourPartitions`, `groupDayPartitions`) are unchanged apart from no longer listing; parsing stays relative to the prefix (#619). `measurementPrefix` replaces the hand-built `database + "/" + measurement + "/"`.
- The preview lists once per measurement and hands the listing to every enabled tier.
- The compaction cycle keeps listing per tier, as discussed in the issue: hourly deletes its inputs and writes its outputs before daily runs, so a shared listing would be stale there. A comment at the tier loop in `runCycleInternal` records why, so nobody "finishes the optimization" later. Since hourly and daily run on separate schedulers with a single tier per cycle, the preview was the path that listed twice on every call.

## Test plan

- [x] `internal/compaction/listing_reuse_test.go`, written before the change. `countingBackend` wraps the local backend and counts `List` per prefix (it keeps `ListDirectories`, so database/measurement discovery is not counted).
- [x] `TestManager_FindCandidates_ListsMeasurementOnceAcrossTiers`: 6 old files in `db1/m1`, both tiers enabled → one hourly and one daily candidate, `db1/m1/` listed once. On `main`: `db1/m1/ listed 2 times across two tiers, want 1`.
- [x] `TestManager_FindCandidates_ListsPseudoDatabaseMeasurementOnce`: spoke-namespace pseudo-database `rocket-01/telemetry` via `SetNamespaceExpander`; both candidates carry `Database: rocket-01/telemetry`, `Measurement: m1`, all 6 files (prefix-relative parsing intact), `rocket-01/telemetry/m1/` listed once. On `main`: `rocket-01/telemetry/m1/ listed 2 times across two tiers, want 1`.
- [x] `FindCandidatesFromListing` ignores objects outside the prefix and returns nothing for a disabled tier (hourly and daily).
- [x] `go test -race ./internal/compaction/` — pass. `go build ./...`, `go vet ./internal/compaction/`, `gofmt` on touched files — clean (`scheduler.go` and `watcher.go` are unformatted on `main` and untouched here).
- [x] `go test -tags=duckdb_arrow -race ./...` (the CI command): 37 packages `ok`, exit 0.
- Environment: macOS 15 / arm64, Go 1.27.1, cgo + DuckDB available.


xe-nvdk (approved):
Approving. This is the (1)+(2) shape from the issue thread, done exactly as discussed, and the PR description is the kind reviewers hope for: it says what was run and what was observed, and it was right.

- `FindCandidatesFromListing` is the no-I/O scan, `FindCandidates` is list + scan, the grouping loops are untouched and still parse relative to the prefix, and `measurementPrefix` replaces the hand-built string in every place it was built.
- The preview lists once per measurement and hands the listing to every enabled tier. The cycle keeps one listing per tier, and the comment at the tier loop says why, so nobody "finishes the optimization" later.
- The counting fake stays on the `DirectoryLister` path so discovery is not counted, and the spoke-namespace case proves the pseudo-database prefix survives. I verified both counting tests fail on `main` with exactly the messages in your description (`listed 2 times across two tiers, want 1`) and pass here, `go build ./cmd/... ./internal/...` is clean, gofmt and vet are clean on the touched files, and the compaction suite passes with `-race`. `HourlyTier` and `DailyTier` are the only `Tier` implementations in the tree, so the interface change has no stragglers.
- Release note is accurate, at the top of Bug fixes, and explains why the cycle is unchanged.

One thing I did on your branch: `main` moved after your merge (three entries landed at the top of Bug fixes today), so the PR showed a positional conflict. I merged `main` into `fix/316-compaction-single-listing` (3c4b0b8, a merge commit, no rewrite), kept your entry on top, and verified the merged tree. Pull before you push again. First-time contributor runs need a maintainer to approve CI; I've done that, and this merges when it is green.

Thank you for finding the staleness hazard before writing a line, and for the write-up on the issue. That is the model for how these should go.

xe-nvdk:
Merging — thank you @alexeymoskalev-devops. Finding the staleness hazard before writing a line, and the write-up on the issue, is the model for how these should go. README credit follows in a moment.

Issue #759, opened 2026-09-12
A permanently unpullable manifest entry keeps the query gate red until the process restarts


xe-nvdk:
Follow-up from #747. Pre-existing and broader than that issue: it affects the whole permanently-unpullable class, not only invalid paths.

`catchupFailedPaths` is cleared only by `clearCatchUpFailureLocked` (`internal/cluster/filereplication/puller.go`), called only from `finishEntry` on a **successful pull of the same path**. Nothing else clears it. So when `query.gate_on_catchup` is enabled and a catch-up-tagged entry can never be pulled, `FullyCaughtUp()` stays false for the life of the process, and the node answers 503 to every query.

Two ways to get there:

1. A file no peer holds. `job.go` already says so in as many words: those entries stay in the manifest and catch-up pulls "will return ErrFileNotOnPeer forever until operator intervention or the periodic Phase 5 reconciler".
2. A manifest entry whose key no backend can address (#747). Quarantined now, and deliberately still counted against the gate, because the file really is absent and the read path is a glob that would silently return fewer rows.

Removing the offending entry from the cluster manifest is the operator's remedy in both cases, and today it does not help: the gate only reopens on restart. #747's log line says this explicitly so the advice is at least accurate, but the behaviour is still wrong.

## The fix

Wire the FSM delete callback (`coordinator.go` `onDelete`, fired from `fsm.go`) to `clearCatchUpFailure(path)`. Removing the manifest entry then reopens the gate without a restart, which is what makes a strict gate affordable, and it is the missing link between reconciliation reporting an unfixable entry and an operator being able to act on it.

Worth checking at the same time whether `catchupDroppedPaths` needs the same treatment.

Note for whoever picks this up: there is currently **no way to delete a manifest entry by path**. `/api/v1/cluster/files` is GET-only, the row-delete path reaches the Raft delete only after `read_parquet` opens the file, and retention builds eligibility from file metadata it also cannot read. So for the #747 class specifically, the escape hatch this issue describes needs `DELETE /api/v1/cluster/files?path=` to exist first, or the self-heal has nothing to trigger it.

Thundercloud12:
working on this!

xe-nvdk:
> working on this!

You are unstoppable! Thank you!

xe-nvdk:
[...]

PR #790, opened 2026-09-14
feat(cluster): add DELETE /api/v1/cluster/files and reopen query gate…

Thundercloud12:
Follow-up to #747 / Addresses #759

## Summary

When a file in the cluster manifest cannot be pulled during catch-up—for example, because no peer holds the file or the storage key is invalid under #747—`puller.catchupFailedPaths` permanently records the failure.

With `query.gate_on_catchup=true`, this causes `FullyCaughtUp()` to remain `false`, resulting in HTTP 503 responses for the remaining lifetime of the process.

Previously, removing the offending entry from the cluster manifest did not notify the puller that the failed path had been removed. As a result, operators had to restart the process to reopen the query gate.

Additionally, there was no HTTP endpoint for removing a manifest entry by path.

### 1. `DELETE /api/v1/cluster/files?path=` API

* Added `handleDeleteFile` to `internal/api/cluster.go`.
* Added admin authentication for the endpoint.
* Validates the `path` query parameter:

  * Returns `400 Bad Request` if `path` is empty.
  * Returns `404 Not Found` if the path does not exist in the manifest.
* Invokes `coordinator.DeleteFileFromManifest(path, reason)` through Raft consensus.
* Added unit test coverage in `internal/api/cluster_test.go`.

### 2. Self-Healing Query Gate

* Exported `ClearCatchUpFailure(path)` and `ClearCatchUpDrop(path)` on `filereplication.Puller`.
* Wired the FSM `onDelete` callback in `Coordinator.startFilePullerLocked` to invoke both methods when a manifest entry is deleted.
* Added integration test `TestPhase3CatchUpManifestDeleteHealsGate`.
* The integration test verifies that deleting a failed catch-up entry:

  1. Clears the recorded catch-up failure.
  2. Causes `FullyCaughtUp()` to transition from `false` to `true`.
  3. Reopens the query gate without requiring a process restart.

### 3. Release Notes

Updated `RELEASE_NOTES_2026.09.2.md` to document:

* The new `DELETE /api/v1/cluster/files?path=` endpoint.
* The self-healing behavior of the query gate after removing a failed manifest entry.

## Verification

The following checks pass successfully:

```bash
go test -race -v -run TestClusterHandler_DeleteFile ./internal/api
go test -race -v -run TestPhase3CatchUpManifestDeleteHealsGate ./internal/cluster
go test -race ./internal/cluster/filereplication/...
gofmt -l
```

All modified files are `gofmt` clean.


Thundercloud12:
@xe-nvdk please havbe a look at this one too, a big one as compared to the other's i think i shopuld have broken into two pr one for api and one for the implementation, please let me know if any changes!!

xe-nvdk (changes requested):
Thanks for taking this one, and you were right on both counts: it is bigger than the others, and it should be two PRs. Below is what I found, how I found it, and the split I'd suggest. Verified first: builds under both tag sets, gofmt and vet clean on the touched files, the cluster, file-replication and API suites under `-race`, a second reviewer working the cluster-operations and security checklists, and a live three-writer-plus-reader cluster built from this branch.

**1. The endpoint deletes files, and nothing about it says so (blocking).** On `main`, the FSM delete callback does not stop at the manifest: on every node with local storage it enqueues a real `backend.Delete` for the path (`coordinator.go` `onDelete` → delete-worker pool, 500 ms grace, then unlink), and Raft applies the entry on every node. I ran it live: a file present on all three writers, `DELETE /api/v1/cluster/files?path=…` with no other parameter, 200, and within six seconds the file was gone from every writer. That is the right mechanism for the case #759 describes, where no node holds bytes, but on a healthy path it is a cluster-wide, unrecoverable delete in one admin request, and the PR body, the handler comment and the release note all describe it as removing a manifest entry. On S3/Azure the effect flips: nothing is unlinked and the object becomes an orphan for the Phase 5 reconciler. So:
- require `?confirm=true` and refuse with 400 otherwise, the way `DELETE /api/v1/databases/:name` already does for a destructive admin call;
- say what it does, in the handler comment, the note and the docs: removes the manifest entry; on local-storage nodes, deletes the file on every node; on shared backends, leaves an orphan object;
- log at Warn with node, path and reason, and get the target into the audit trail: the global audit middleware records `api.delete` with actor and status but never captures the query string, so today the audit row does not say which file was deleted. Either move the path into the URL (`/api/v1/cluster/files/*`) like `DELETE /cluster/nodes/:id`, or add it to the audit event.

**2. The self-heal can make the permanent-red state it exists to cure (blocking).** `FullyCaughtUp` needs `catchup_failed == 0`, and a catch-up pull records its failure in the deferred `finishEntry` only after the retries run out: up to three attempts per peer with a 60 s fetch timeout each, so minutes, during which the operator sees the entry failing and deletes it. In that ordering the callback's clear is a no-op (nothing recorded yet), the worker then gives up, `isCatchUpPath` is still true, and `catchup_failed` becomes 1 with no way to clear it: a second API call gets 404 because the entry is gone, and `applyDeleteFile` returns before invoking the callback when the path is not in the manifest. Only a restart heals, which is #759's symptom. The second reviewer reproduced it in a throwaway in-package test (`markCatchUp+inflightAdd → ClearCatchUpFailure/Drop → finishEntry(failed)` ⇒ `FullyCaughtUp=false`, `catchup_failed=1`). Fix: replace the two exported clears with one `puller.OnManifestDelete(path)` that, under `inflightMu` in a single critical section, clears the failure, clears the drop, and **removes the catch-up tag**, so a later `finishEntry` neither counts the failure nor double-decrements (both clears are already membership-guarded). Add that ordering as a test.

**3. Test scaffolding in production code, including a non-consensus FSM write (required).** `NewTestCoordinator`, `SetRaftFSM`, the three `*ForTest` hooks on the puller (unused by any test), and the `raftNode == nil && raftFSM != nil` branch in `DeleteFileFromManifest` that applies a command straight to the FSM with a fabricated `Index: 1`. I traced the constructor: the FSM and the Raft node are created in the same block and a node failure aborts construction, so that branch is unreachable in every supported mode, which means it exists only so the API unit test can run without Raft. It is a one-node, no-quorum mutation sitting in the method every retention, compaction and delete caller uses; a refactor that ever separates the two constructions turns it live. Remove all of it. The API handler needs two methods: use the pattern `delete.go` already has (`DeleteCoordinator` interface with `GetFileEntry`) — an unexported interface field set to the coordinator in the constructor and stubbed in `cluster_test.go`. For the real path, `package cluster` tests already build a coordinator around a bootstrapped Raft node (`filereplication_integration_test.go`, `multi_writer_integration_test.go`); one test driving `DeleteFileFromManifest → raftNode.DeleteFile → FSM → onDelete → puller` through production wiring is what proves the feature, since today all five handler tests exercise only the scaffolding branch and the integration test hand-wires the callback rather than going through `startFilePullerLocked`.

**4. Smaller things to fold in:**
- Map `errors.Is(err, raft.ErrManifestApply)` (no leader, leader unreachable, apply timeout) to 503 with `Retry-After`, as `delete.go` and `compaction.go` do for retry-later, instead of 500 with the raw error string.
- Run `clusterraft.ValidateManifestPath` on `path` before the manifest lookup, and cap `reason` (it goes into the Raft log and every node's logs unbounded).
- The dead `if puller != nil` (the closure captures a non-nil puller by construction) and the two lowercase shims kept only for `puller_test.go`.
- Release notes: the self-heal is a bug fix; the endpoint is a new feature and belongs in a `## New` entry, with the effect from item 1 stated plainly and "immediately clears" dropped (item 2). Docs: the enterprise clustering page (`configuration/clustering.md`, per-endpoint curl section) and the API reference overview both need the endpoint; separate docs PR.

**The split I'd suggest, and the order:** PR 1 = `OnManifestDelete` on the puller, the coordinator wiring, and an in-package test through a real bootstrapped Raft node. Small, a pure bug fix, closes the mechanism #759 describes, needs no scaffolding. PR 2 = the endpoint, with confirm, validation, the 503 mapping, the audit target, interface-based handler tests, the `## New` note and the docs pages. They don't depend on each other at the code level, and keeping them apart means the destructive endpoint gets reviewed on its own terms.

For the record: auth is present and wired (`filesGroup.Use(auth.RequireAdmin)`, constructor gets the auth manager), manifest-before-storage ordering is correct (Raft delete first, local unlink from the callback), and the live run's healthy-path checks were clean. I still owe you the live #759 scenario end to end (reader started against an unpullable entry, 503, DELETE, 200); it is blocked on license activation slots on my side, not on your code, and I'll add it as a comment when it runs.

Thundercloud12:
@xe-nvdk got it, for now I will be closing this pr to move it out of your backlog and in the coming days open two seperate pr's as suggested, thanks!

xe-nvdk:
Live follow-up on item 2, both orderings, on a three-writer + gated-reader cluster built from 7374b2c (`ARC_CLUSTER_QUERY_GATE_ON_CATCHUP=true` on the reader).

**Ordering 1 — the pull fails first, then the DELETE: the PR works.** Reader stopped, a file written and flushed on the leader and replicated to all writers, the file removed from every writer, reader started. Its startup catch-up found the entry, every peer answered not-found, and the pull gave up after about 2 s (`catchup_failed=1`, query → 503). `DELETE /api/v1/cluster/files?path=…` on the leader → 200, and 1.2 s later the reader answered 200 with no restart. That is #759 healed, and it is the part of this PR that should ship.

**Ordering 2 — the DELETE lands while the pull is still in flight: the gate stays red until a restart.** Same setup, with one follower paused so the reader's fetch to it hangs. (Correction to my review: the handshake deadline is 5 s, so three attempts is about 15 s in flight, not minutes; the 60 s fetch timeout only applies to a peer that accepts the handshake and then stalls. Fifteen seconds is still a wide-open window for an operator watching a failing entry.) Timeline from the reader's logs and the 503 body:

```
17:52:22  reader up, query → 503        catchup_inflight=1 catchup_failed=0
17:52:22  DELETE …/cluster/files?path=…  → 200
17:52:22  reader FSM: "File removed from cluster manifest"   ← the clear runs here; nothing recorded yet
17:52:37  reader puller: "File pull giving up after max attempts"
17:52:39  query → 503                    catchup_inflight=0 catchup_failed=1
17:53:04  follower unpaused, cluster fully healthy → still 503
17:53:04  second DELETE of the same path → 404   (entry gone; the FSM returns before the callback)
17:53:20  docker restart reader → 200
```

So on this branch an operator who runs the DELETE while the entry still shows as in-flight lands in exactly the state #759 describes, and the API can no longer get them out of it. The fix is the one in the review: a single `puller.OnManifestDelete(path)` that, under `inflightMu`, clears the failure, clears the drop, and removes the catch-up tag, so the later `finishEntry` has nothing to count. The repro is cheap enough to go in the test plan: `deploy/docker-compose/enterprise-local`, reader with the gate env set, `docker stop` the reader, write + flush on the leader, `rm` the file on all three writers, `docker pause` a non-leader writer, `docker start` the reader, DELETE within ~10 s of the first 503, then watch `catchup_failed` in the 503 body.

Item 1 also reproduced on this build: a file present on every node, DELETE with no confirm parameter → 200, and the file was gone from all four nodes within 6 s.

xe-nvdk:
Filed the two findings as tracking issues for the split: #794 is the spec for the endpoint PR (confirm, wording, audit target, validation, 503 mapping, no scaffolding), #795 is the in-flight ordering the self-heal PR has to handle on top of #759. Point PR 1 at #759 + #795 and PR 2 at #794 and you will have everything from the review in one place each.

xe-nvdk:
@Thundercloud12 to take one of the two off your plate: we will do the self-heal half (#759 + #795) on our side today, since it is small and it blocks the endpoint's integration test. The endpoint half (#794) is yours whenever you are ready, and once #795 is in, your test can assert the gate reopens. Thanks again for the digging on #790, the live runs confirmed the approach.

xe-nvdk:
@Thundercloud12 the self-heal half is merged: #800 closes #759 and #795 (release note credits your #790 groundwork). Two things worth knowing for the endpoint PR (#794): with #800 in, the reader forgets the entry on the FSM delete callback in every ordering, so your endpoint only has to issue the Raft delete (`DeleteFileFromManifest`) and the gate reopens by itself; and while running it live I hit #799 (a reader's startup walk can run before its log replay), which is unrelated to the endpoint but will show up as a 200 instead of a 503 if you test with a freshly restarted reader, so seed the entry while the reader is up, as the harness in #795's comment does.

Issue #638, opened 2026-08-20
medium(iceberg): compaction's upload-then-delete window commits snapshots with duplicate data (~2x rows transiently)


xe-nvdk:
Found during the 2026-08-20 adversarial audit of the Iceberg export subsystem.

## Problem

Compaction uploads the compacted file **before** deleting its sources (`internal/compaction/job.go:409` upload → `:435-445` delete) — correct crash-safety ordering for Arc itself. But a reconcile pass that lists storage inside that window registers the compacted file **and** all N raw files into one committed snapshot: external readers see ~2× rows for that hour/day until the next pass removes the raws (up to ~2 reconcile intervals at the default 300s).

The window per partition is small, but compaction runs continuously across many measurements, so it *will* be hit in steady state. It self-heals, and it also lengthens the #632 chain (the raws' removal snapshot later drives the ENOENT expiry noise).

## Fix shape

At minimum, document the transient-duplication window in the Iceberg export docs (external `COUNT(*)`/aggregations can transiently over-count during compaction). A real fix needs either compaction-aware quiescence (skip measurements with an in-flight compaction job) or excluding raws when a sibling `_compacted` file covering the same partition exists in the listing.

Issue #637, opened 2026-08-20
medium(backup): iceberg.warehouse outside the storage root is silently omitted from backups — DR restore wedges every measurement


xe-nvdk:
Found during the 2026-08-20 adversarial audit of the Iceberg export subsystem. Same non-default-config blind-spot class as #534.

## Problem

Backup discovers Iceberg warehouse metadata only by walking **data storage** (`internal/backup/backup.go:101,120` — `isIcebergMetadata` over the dataStorage listing). But a warehouse **outside** the storage root is a supported configuration: config validation checks only backend type, tiering, and retain_snapshots (`config.go:1220-1233`), and the exporter explicitly handles the outside-root case as "hint unaddressable" (`exporter.go:594-604, 638-658`).

With such a config, the backup contains catalog rows (in the shared SQLite DB) whose `metadata_location`s point at files that were **never backed up** — and reports success.

## Failure scenario (DR restore permanently wedged)

Restore to a fresh host: `LoadTable` fails reading the missing metadata file → `EnsureTable` falls through to `CreateTable` → the catalog `INSERT` hits the primary-key conflict, which is **not** iceberg-go's `ErrTableAlreadyExists` sentinel (`catalog/sql/sql.go:307-323` vs `exporter.go:746-749`) → every measurement fails every reconcile pass, forever, until manual catalog surgery.

## Fix shape

At minimum: backup logs a prominent warning (or fails, per operator flag) when `iceberg.warehouse` is set and not under the data storage root. Better: back up the warehouse tree wherever it lives. Additionally, `EnsureTable` should treat the PK-conflict shape as "table exists but metadata missing" and surface a clear, actionable error instead of a permanent conflict loop.

Issue #634, opened 2026-08-20
high(iceberg): edge-sync hub spoke namespaces are exported as garbage tables — source discovery predates #619


xe-nvdk:
Found during the 2026-08-20 adversarial audit of the Iceberg export subsystem. Same class as the #619 compaction fix — the Iceberg source never got the spoke-namespace treatment.

## Problem

An edge-sync hub stores received data at `{spoke_id}/{db}/{meas}/{y}/{m}/{d}/{h}/*.parquet` under the storage root (`internal/edgesync/receive.go` `NamespacedPath`). Compaction learned in #619 to expand `{spoke}/{db}` into pseudo-databases; the Iceberg walk did not: `StorageWalkSource.Measurements()` takes top-level dirs as databases and second-level dirs as measurements (`internal/iceberg/source.go:62-91`), and `FilesAndLocal` lists `{db}/{meas}/` recursively. Nothing forbids `iceberg.enabled` on a hub (`config.go:1218-1236` checks only backend/tiering/retain).

## Failure scenario

Hub receiving spoke `rocket-01` with database `factory` (measurements `cpu`, `mem`): discovery yields `Measurement{Database:"rocket-01", Measurement:"factory"}`. Every parquet from **all** measurements under `rocket-01/factory/` lands in one file list; `UnionSchema` unions unrelated measurements' schemas. Either it errors on a cross-measurement type collision (e.g. `value` long vs double) and the "measurement" logs ERROR **every pass forever**, or it succeeds and mints catalog namespace `arc_rocket-01`, table `factory` — all measurements' files mixed into one table with a franken-schema. #619's hub compaction then churns those file sets, snapshotting the garbage every cycle.

(`.sync-staging` and `.part` files are safely excluded — hidden-dir skip + `.parquet` suffix check; the problem is purely the un-expanded namespace.)

## Fix shape

Either give `StorageWalkSource` the same registered-spoke expansion the compaction manager got in #619 (spoke namespaces become `{spoke}/{db}` pseudo-databases with their real measurements — probably the desired end state: hubs exporting received data as proper tables), or exclude registered spoke namespaces from Iceberg discovery, or refuse `iceberg.enabled` + edge-sync hub at config load until supported. Whichever is chosen, the hub + iceberg combination must not silently produce garbage tables.

xe-nvdk:
Fixed in #824 (merged as `4681c59`). This issue stayed open only because the commit referenced it without a closing keyword — the fix is on main.

[...]

Issue #633, opened 2026-08-20
high(iceberg): in-place parquet rewrite by row-level DELETE is invisible to the path-only fingerprint — external engines keep serving deleted rows


xe-nvdk:
Found during the 2026-08-20 adversarial audit of the Iceberg export subsystem.

## Problem

The row-level DELETE API's partial-match branch rewrites a parquet file **in place**: filtered data goes to a temp file, then `os.Rename(tempFile, filePath)` over the original path (`internal/api/delete.go:766-770`) — same filename, fewer rows, different size.

The Iceberg reconciler's change detection is a hash of **paths only**: `fingerprint()` uses `FileRef.PhysicalPath` exclusively (`internal/iceberg/scheduler.go:299-308`), and `source.go:126` never populates `FileRef.SizeBytes`. The rewritten file is in both `want` and `have`, so the reconcile diff (`exporter.go:374-385`) is empty — forever.

## Failure scenario

Iceberg-enabled instance, table exported. Operator issues a WHERE-based delete that partially matches a file. The Iceberg manifest entry keeps the registration-time `record_count`, `file_size_in_bytes`, and column bounds. External engines answer `COUNT(*)` from manifest record counts without touching data → **deleted rows are still counted**; engines that trust `file_size_in_bytes` for split planning/footer location (Trino, Java Iceberg) can fail reading the shorter file. Nothing ever refreshes the entry until compaction happens to replace the file. No non-default config needed — `iceberg.enabled=true` + one `POST /delete` with a partial WHERE.

(The full-file-match branch deletes the file outright and is handled correctly; only the partial-rewrite branch is invisible.)

Note: `delete.go:722-727` already nudges the Raft manifest and claims deletion "does not affect query correctness" — for Iceberg external readers it does.

## Fix shape

Include file size (and/or mtime) in both the fingerprint and the reconcile diff key so a same-path rewrite re-registers the file (remove + re-add in one snapshot), or have the delete handler nudge the exporter the way it nudges the Raft manifest.

PR #824, opened 2026-09-14
fix(iceberg): snapshot-expiry data-file deletion (#632) and spoke-namespace export (#634)

### LINKED ISSUE #632: high(iceberg): ExpireSnapshots misreads successful expiry as failure; PostCommit hook holds delete authority over Arc's data files

xe-nvdk:
Fixes #632. First of the five Iceberg issues milestoned for 26.09.2.

## The bug

iceberg-go runs orphan deletion for an expiry as a **post-commit hook**, and defaults it on. `Commit` lands in the catalog, then the hook `os.Remove`s the expiring snapshots' manifest lists, manifests, and any data files they referenced — joining every failure into the error `Commit` returns.

Arc is the wrong shape for that twice over.

**Files leave an Arc table precisely because Arc already deleted them** — compaction, retention, the delete API. So the hook reported `ENOENT` for a commit that had succeeded, and the error list grew every pass, because each expiring manifest carries entries for every file ever removed. The exporter read that as failure, logged

```
Iceberg ExpireSnapshots commit failed (non-fatal) — snapshot history grows until it recovers
```

(wrong on both counts), and returned the **pre-expire** table. `writeVersionHint` then published metadata still listing snapshots whose manifest-list files the hook had just deleted. Directory readers doing snapshot listing or time travel hit missing files, and a measurement that went quiet kept the dangling hint indefinitely.

**The tail risk is the more important half.** `postCommit=true` handed iceberg-go physical delete authority over Arc's **primary Parquet data files**. Any future listing bug that transiently omitted live files for `retain` generations would have become silent deletion of customer data by the export subsystem. Arc owns the data-file lifecycle; the exporter must never delete data files, and now cannot.

## The fix

`WithPostCommit(false)` — available in the pinned iceberg-go v0.6.0, and upstream's own doc comment describes exactly this case:

> Set to false to defer file deletion to a separate maintenance job, avoiding conflicts with in-flight queries that may still reference those files.

The exporter now expires snapshots in the catalog and leaves file deletion to Arc. The residue is unreferenced metadata files, bounded by `retain`; `pruneOldVersionFiles` already bounds the `v<N>.metadata.json` copies.

## Why the existing test missed it

`TestExpireSnapshotsAndPruneVersions` writes 8 files and **leaves them all on disk**, so the post-commit hook always succeeds. The issue called this out, and it is the whole reason the bug was invisible.

The new test reproduces the production order: register a file as the table's only content, then **delete the superseded file from disk** before the pass that expires the snapshot referencing it. It asserts both symptoms:

1. No false `"ExpireSnapshots commit failed"` in the captured log
2. No published snapshot whose manifest list is missing from disk

Verified to fail on **both** assertions against the unfixed code:

```
exporter_test.go:459: expiry logged a false commit failure (#632)
exporter_test.go:506: published metadata advertises snapshot 2750253405094528940
                      whose manifest list is missing: .../snap-2750253405094528940-0-....avro
```

## Configuration matrix

| Configuration | Reaches new code? | Preconditions |
|---|---|---|
| `iceberg.enabled=false` (default) | **No** — exporter never constructed | N/A |
| `iceberg.enabled=true`, history ≤ `retain` | **No** — `expireSnapshots` returns early | Guarded by the snapshot-count check |
| `iceberg.enabled=true`, history > `retain`, files still on disk | **Yes** — previously worked by luck | Unchanged behaviour |
| `iceberg.enabled=true`, history > `retain`, **files deleted by Arc** | **Yes** — the bug path | Now expires cleanly; new test covers it |
| `iceberg.retain_snapshots=1` (non-default) | **Yes** — expires sooner, so more churn | Same code path, more often |
| Cluster mode | **Yes** — reconciler is writer-gated, single writer | Unchanged; no concurrent expire race |

## Test plan

- [x] New regression test fails on both assertions pre-fix, passes after (revert-run-restore)
- [x] Full `./internal/iceberg/` suite green
- [x] Full `./internal/...` suite green
- [x] `go build`, `go vet` clean; `gofmt` clean on both files I touched (`scheduler_test.go` is flagged but is a pre-existing issue on main, untouched here)

## Remaining Iceberg milestone

#634 (spoke namespaces exported as garbage tables — worth doing before any hub runs Iceberg), #633, #637, #638.


xe-nvdk:
## Added: #634 — spoke namespaces exported as garbage tables

This PR now carries the second Iceberg fix as well, since both touch the same package and the audit flagged #634 as needing to land *before any hub runs Iceberg*.

### The bug

A hub stores received data one level deeper than local data: `{spoke_id}/{db}/{measurement}/{y}/{m}/{d}/{h}/*.parquet`. `StorageWalkSource` read the top two levels as `(database, measurement)`, so a spoke `rocket-01` holding database `factory` was discovered as **one** measurement named `factory`.

Every Parquet from every measurement under that spoke then went into a single file list with their schemas unioned. Either the union failed on a cross-measurement type collision — error every pass, forever — or it succeeded and minted a table mixing all measurements into one franken-schema, which hub compaction then re-snapshotted every cycle.

**Confirmed directly** rather than inferred. Without the expander:

```
CONFIRMED #634: spoke discovered as (database="rocket-01", measurement="factory")
                — cpu and mem would union into one table
```

### The fix

Compaction learned this in #619; the Iceberg source never did. It now takes the same registered-spoke expansion, wired from `main.go` with the same `receivedNamespaces(spokeRegistry)` helper, and nil on a non-hub where the walk is already at the right depth.

**One subtlety worth review.** `tableIdent` now maps the separator to `.`, so the namespace stays a single path token. The SQL catalog names namespace directories `<namespace>.db`, and an unsanitized slash would nest that directory a level deeper than `isWarehouseDir` expects — which would feed the exporter's **own metadata back in as a user database** on the next walk. A real Arc database name cannot contain a `.`, so there is no collision. Same mapping compaction uses for job IDs.

`DropDatabase` already refused slashed names, so its two unsanitized namespace uses are unreachable for pseudo-databases — verified.

**Fail-safe direction:** if the spoke lookup fails, spoke namespaces are *skipped* for the pass rather than exported un-expanded. Exporting them wrong mints catalog tables that have to be cleaned up by hand, so skipping is the cheaper failure.

### Operator note in the release notes

Hubs that already ran `iceberg.enabled` may have garbage tables in the catalog. They are not repaired automatically — drop them and the next pass creates the correct per-measurement tables.

### Tests

- Expansion produces the real measurements under a `{spoke}/{db}` pseudo-database
- A local database is untouched by expansion
- The un-expanded `(rocket-01, factory)` pair must **not** appear
- Expander failure skips spokes without losing ordinary databases

Full `./internal/...` suite green; `gofmt`, `go vet` clean.

PR #830, opened 2026-09-15
feat(cluster): add DELETE /api/v1/cluster/files endpoint (#794)

### LINKED ISSUE #794: cluster: DELETE /api/v1/cluster/files must be built as a destructive, cluster-wide operation (confirm, honest wording, audit target, validation, no test scaffolding)
Spec for the endpoint half of the #790 split (the self-heal half is #759 plus the follow-up issue on the in-flight ordering). #759 says the escape hatch needs `DELETE /api/v1/cluster/files?path=` to exist; #790 built it and was closed by its author to be reopened as two PRs. Everything below was verified on #790's head (7374b2c) with a live three-writer + reader cluster.

## What a manifest delete actually does

The endpoint is described everywhere in #790 (PR body, handler comment, release note) as "removing a manifest entry". The mechanism underneath it is a cluster-wide file deletion:

`DeleteFileFromManifest` → Raft `CommandDeleteFile` → `ClusterFSM.applyDeleteFileStruct` on every node → `onFileDeleted` callback → `coordinator.go` `onDelete` → on every node with local storage, the Phase 4 delete worker unlinks the file after a 500 ms grace.

Live: a file present on all three writers and the reader, one `DELETE /api/v1/cluster/files?path=…` with no other parameter, HTTP 200, and the file was gone from all four nodes within 6 s. On S3/Azure the effect flips: nothing is unlinked and the object becomes an orphan for the Phase 5 reconciler.

That is the right mechanism for the #759 case (no node holds bytes anyway), but on a healthy path it is an unrecoverable, cluster-wide delete in one admin request, so the endpoint has to be built as one.

## Requirements

1. **Confirmation.** Require `?confirm=true` and refuse with 400 otherwise, as `DELETE /api/v1/databases/:name` already does for a destructive admin call.
2. **Honest wording**, in the handler comment, the release note (this is a `## New` entry, not a bug fix) and the docs (enterprise `configuration/clustering.md` per-endpoint section and the API reference overview): removes the manifest entry; on local-storage nodes, deletes the file on every node; on shared backends, leaves an orphan object.
3. **Warn-level log** on the handling node with node id, path and reason.
4. **Audit target.** `internal/audit/middleware.go` records `c.Path()` only and never populates `Detail`, so a query-string target never reaches the audit row: today the row would say `api.delete` + actor + status and not which file. Every existing DELETE route carries its target in the URL (`/cluster/nodes/:id`, `/databases/:name`, `/backups/:id`, …). Either move the path into the URL (`DELETE /api/v1/cluster/files/*`) or put it in the audit event's `Detail`.
5. **Validation.** `clusterraft.ValidateManifestPath` on `path` before the manifest lookup; cap `reason` (256 is plenty), because it is written into the Raft log and into every node's log line.
6. **Retry-later mapping.** `errors.Is(err, raft.ErrManifestApply)` (no leader, leader unreachable, apply timeout) → 503 with `Retry-After`, the way `delete.go` and `compaction.go` do it, instead of 500 with the raw error string.
7. **No test scaffolding in production code.** #790 carried `NewTestCoordinator`, `SetRaftFSM`, three unused `*ForTest` hooks on the puller, and a `raftNode == nil && raftFSM != nil` branch in `DeleteFileFromManifest` that applied the command straight to the FSM with a fabricated `Index: 1`. The FSM and the Raft node are created in the same block of the constructor and a node failure aborts construction, so that branch is unreachable in every supported mode and exists only so the API unit test can run without Raft. It is a no-quorum mutation sitting in the method every retention, compaction and delete caller uses. The handler needs two methods from the coordinator: use the `delete.go` `DeleteCoordinator` pattern (unexported interface, stubbed in `cluster_test.go`). For the real path, `package cluster` already builds coordinators around a bootstrapped Raft node (`filereplication_integration_test.go`, `multi_writer_integration_test.go`); one test through `DeleteFileFromManifest → raftNode → FSM → onDelete` is what proves the feature.
8. A read-only token gets 403 (test).

Auth wiring was correct in #790 (`filesGroup.Use(auth.RequireAdmin)`, constructor receives the auth manager) and should stay that way.

Refs: #759, #790 (review + live runs at https://github.com/Basekick-Labs/arc/pull/790#issuecomment-5668313471).

Thundercloud12:
## Summary

This PR implements the administrative `DELETE /api/v1/cluster/files` endpoint, completing the second part of the split requested in #790 and tracked by #794. The query-gate self-healing groundwork was merged in #800.

The endpoint allows cluster operators to remove an entry from the cluster-wide Raft manifest and triggers physical file deletion across the cluster on nodes using local storage.

### 1. Destructive Guard (`?confirm=true`)

* Mirrors `DELETE /api/v1/databases/:name`.
* Requires the `?confirm=true` query parameter.
* Missing or false values return `400 Bad Request` with:

  > `confirmation required: add ?confirm=true to delete the file from cluster manifest and storage`

### 2. Validation

* Validates the path using `clusterraft.ValidateManifestPath(path)` before lookup.
* Rejects invalid, absolute, scheme-prefixed, traversal, and null-byte paths with `400 Bad Request`.
* Adds a maximum length of 256 characters for the optional `reason` parameter using `maxDeleteReasonLen = 256`.
* Reasons exceeding the limit return `400 Bad Request`.

### 3. Raft Apply Error Mapping

* Raft application failures such as no leader, an unreachable leader, or an apply timeout return `503 Service Unavailable`.
* Errors matching `clusterraft.ErrManifestApply` set:

  * `Retry-After: 1`
  * Response: `cluster manifest update unavailable; retry later`
* Generic `500 Internal Server Error` is reserved for unexpected errors.

### 4. Interface-Based Test Scaffolding

* Defines `clusterFilesCoordinator` with:

  * `GetFileEntry`
  * `DeleteFileFromManifest`
* Wires the interface into `ClusterHandler`.
* Adds an in-memory `stubFilesCoordinator` in `internal/api/cluster_test.go`.
* No production test hooks or fake coordinators are introduced.

### 5. Logging & Documentation

* Emits a `Warn`-level log:

  > `File removed from cluster manifest and deleted cluster-wide`
* Documentation, handler comments, and release notes clarify that:

  * Removing the manifest entry enqueues physical deletion through the worker pool across local-storage nodes.
  * Orphan objects on shared backends such as S3/Azure are left for reconciler sweeping.

### 6. Release Notes

* Added an entry under `## New` at the top of `RELEASE_NOTES_2026.09.2.md`.
* Maintains strict newest-first ordering.
* Includes contributor attribution.

## Test Plan

* [x] `gofmt -l` clean across all touched files.
* [x] `go vet ./internal/api` clean.
* [x] `go test -race -v -run TestClusterHandler_DeleteFile ./internal/api` passes all 10 test cases:

  * [x] `TestClusterHandler_DeleteFile_Disabled` — standalone mode → `200` disabled
  * [x] `TestClusterHandler_DeleteFile_AdminAuthRequired` — `401` unauthenticated, `403` non-admin, `404` admin
  * [x] `TestClusterHandler_DeleteFile_MissingPath` — `400` missing path
  * [x] `TestClusterHandler_DeleteFile_InvalidPath` — `400` traversal, absolute, scheme, and null-byte paths
  * [x] `TestClusterHandler_DeleteFile_RequiresConfirmation` — `400` missing, false, or non-true confirmation
  * [x] `TestClusterHandler_DeleteFile_ReasonLength` — `400` when reason exceeds 256 characters
  * [x] `TestClusterHandler_DeleteFile_NotFound` — `404` when file is not in the manifest
  * [x] `TestClusterHandler_DeleteFile_RaftApplyError` — `503` + `Retry-After: 1` on `ErrManifestApply`
  * [x] `TestClusterHandler_DeleteFile_GenericError` — `500` on unexpected errors
  * [x] `TestClusterHandler_DeleteFile_Success` — `200` and file removed from the stub

### Full Test Suites

* [x] `go test ./internal/api`
* [x] `go test ./internal/cluster/...`

Closes #794


Thundercloud12:
@xe-nvdk as suggested by you, inform any changes if needed!

xe-nvdk (changes requested):
Thanks — this is the endpoint #794 described, built the way it asked: the confirm gate, the reason cap, the 503 + `Retry-After` mapping, the 404 before the Raft write, the interface-and-stub test shape with nothing test-only in production code, and a handler comment and release note that say what the delete really does. Auth is wired (`RequireAdmin` on the group; the 401/403/404 test proves it), `go test -race` for `internal/api` and CI are green. Three things before merge, one of which is a correction to my own spec.

**1. The audit row still cannot say which file was deleted (required, #794 item 4).** `internal/audit/middleware.go` records `c.Path()` (`/api/v1/cluster/files`) and never the query string; `AuditEvent.Detail` exists and is persisted (`audit.go` marshals it as JSON) but nothing populates it. So today the audit log of this destructive call shows `api.delete` + actor + status and no target, and the handler comment's "reason recorded in audit logs" is not true yet. Smallest fix, and a useful one for other handlers: the handler stashes `c.Locals("audit_detail", map[string]string{"path": path, "reason": reason})` before returning, and the middleware, after `c.Next()`, does `if d, ok := c.Locals("audit_detail").(map[string]string); ok { event.Detail = d }`. Put the key in a constant in the audit package, add a middleware test that a handler-supplied detail lands in the stored event, and assert in `TestClusterHandler_DeleteFile_*` (success path) that the Locals value is set. Moving the path into the URL would also work but needs care with escaped slashes; the detail map is the safer route.

**2. Do not validate the path before the lookup — my spec was wrong on this (required).** #794 item 5 said run `ValidateManifestPath` first. But `ValidateManifestPath` rejects a backslash, `..`, a dot-prefixed segment and the other unaddressable shapes, and those are precisely the #747-class entries #759 needs this endpoint for: keys an older Arc registered that no listing can reach and that retention and the reconciliation sweep cannot remove. As written, the endpoint can delete a healthy entry and cannot delete the broken one. Correct shape: cap the length of `path` (the FSM already caps keys at 4096), look it up by exact string, 404 if absent, and delete whatever exists. Deleting an unaddressable key is safe end to end: the FSM delete is by exact key, `OnManifestDelete` is map-only, and the delete worker already maps `storage.ErrInvalidPath` to a logged no-op (`coordinator.go` `runDeleteWorker`). Keep `ValidateManifestPath` only to *classify*: when it fails for an existing entry, add `unaddressable_key=true` to the Warn line and the audit detail, so the operator sees they removed one of those. Replace the `InvalidPath → 400` test with: an existing entry whose key fails validation is deleted (200) and flagged; a non-existent invalid key is 404, never 400.

**3. `main` moved under you, and #787 changed a signature you call.** `DeleteFileFromManifest` now takes a context (`DeleteFileFromManifest(ctx, path, reason)`, merged today as #787), so the branch will not compile once rebased: add `ctx context.Context` to `clusterFilesCoordinator`, pass `context.WithTimeout(c.Context(), 30*time.Second)` from the handler (the handler is synchronous, so `c.Context()` is fine here), and update the stub. The release-notes entry also needs the rebase (four entries landed above yours) and its Contributed-by link should point at this PR, #830, not the issue.

Smaller, please fold in:
- Add the node id to the Warn line (`h.coordinator.GetLocalNode().ID`), so a log aggregator can tell which node accepted the delete; #794 asked for node, path and reason.
- Standalone mode answers `200 {"enabled": false}` to a DELETE. It matches `handleRemoveNode`, so it is consistent, but for a destructive call a script that checks only the status code will believe it succeeded. `success: false` in that body, or a 409, would be kinder; your call.
- Docs: the enterprise clustering page (per-endpoint curl section) and the API reference overview on docs.basekick.net need the endpoint and its warning; a docs PR alongside this one, same as the other contributor PRs.

Once those are in I will run it on the live cluster end to end: an entry no peer holds, reader at 503, `DELETE`, reader back at 200 with no restart, which is the whole #759 story with #800 on the other side.

Thundercloud12:
@xe-nvdk made the changes you asked for, can you have a look?

xe-nvdk (changes requested):
Thanks for the rework — every item from the first round is in: exact-key lookup with the length cap and 404 (so #747-class keys can be removed), the `unaddressable_key` classification, audit detail through `audit.DetailLocalsKey` with the middleware and its test, the `ctx` signature after #787, node id in the warning, `success: false` on the standalone response, and the Contributed-by link. The revert-run confirms `TestMiddleware_HandlerSuppliedDetail` fails on main (`detail = ""`), and I ran a 3-writer + 1-reader cluster from this branch: a DELETE sent to a non-leader writer is forwarded, the entry disappears from every node's manifest, the file is unlinked on all four nodes, the query returns no rows afterwards, and the second DELETE of the same path returns 404. Two things left, both small and in `handleDeleteFile`:

**1. Copy `path` and `reason` before putting them in the audit map (must).** Arc runs Fiber with `Immutable=false`, so `c.Query()` returns a string that aliases the request's pooled buffer. The audit middleware only enqueues the event (`Logger.LogEvent` → channel); `writerLoop` marshals `event.Detail` up to a second later, after the handler has returned and the connection buffer has been reused by the next request. The row can then record another request's bytes. The middleware's own fields have the same problem today (filed as #837, not yours to fix), but the two you add should not join them:

```go
import "github.com/gofiber/fiber/v2/utils"

path := utils.CopyString(c.Query("path"))
// ...
reason := utils.CopyString(c.Query("reason"))
```

Everything else in the handler (the coordinator call, the JSON response, the log line) uses them synchronously, so copying once at the top is enough.

**2. Set the audit detail before calling `DeleteFileFromManifest` (should).** Right now the `c.Locals(audit.DetailLocalsKey, …)` block runs only on success, so a refused attempt (404 unknown path, 503 manifest unavailable, 500) is audited as `DELETE /api/v1/cluster/files` with no target. For a destructive endpoint the failed attempts are the rows an auditor wants most. Move the `auditDetail` map construction (with `unaddressable_key`) to just after the `isUnaddressable` line and before the timeout context; the existing tests still pass, and `TestClusterHandler_DeleteFile_NotFound` / `_RaftApplyError` can then also assert the captured detail carries the path.

Optional, not blocking: in `TestForwardApplyToLeader_UsesCallerDeadline` the new wait is a fair fix for the socket-deadline vs. context-timer ordering, but it also lets an early return (a call that failed at 5 ms for an unrelated reason) pass. If you want to keep the "not early" half, add `if elapsed < 90*time.Millisecond { t.Fatalf(...) }` next to the existing upper bound. Fine to leave as is.

I'll write the docs page for the endpoint after merge, no action needed there.

xe-nvdk:
Heads-up: your `TestForwardApplyToLeader_UsesCallerDeadline` tweak was right, and the same race just failed CI on an unrelated PR, so I landed it on main as #843 (your diagnosis, credited; plus a lower bound on `elapsed` to keep the "not early" half). When you rebase for the two remaining items, drop your copy of that hunk and keep main's; the rest of the branch is unaffected.

Thundercloud12:
@xe-nvdk done, have a look whenever free!

xe-nvdk (changes requested):
Thanks for the rework. I rebased the branch onto current `main` and reran the focused race suites plus a live 3-writer + 1-reader local-storage cluster. The core path is solid: a DELETE sent to a non-leader forwarded successfully, all four manifests converged, all four local files were unlinked, a second DELETE returned 404, and quorum loss returned 503 with `Retry-After: 1` while preserving the manifest entry.

Two issues remain before I would merge:

1. **Audit targets are still absent for requests rejected before the handler reaches the `auditDetail` block.** In the live run, a missing-confirmation request was stored as `api.delete`, status 400, `detail: null`; a valid non-admin token was stored as `auth.failed`, status 403, `detail: null`. Successful, 404, and 503 requests do carry the copied path and reason correctly. The confirmation check at `internal/api/cluster.go:467` happens before detail is set, and the group-level `RequireAdmin` at `:94-96` can return before `handleDeleteFile` runs at all. Please capture the bounded/copied target before those rejection paths (route middleware before `RequireAdmin` is one workable shape for the 403 case) and add persisted-audit assertions for 400/403.

2. **The destructive endpoint returns HTTP 200 when clustering is disabled.** `handleDeleteFile` delegates to `respondNotEnabled`, whose status remains 200. A client that checks only status will report the deletion as successful although nothing happened. Current `main` already handles the analogous writer-demotion operation with 409 specifically to avoid this ambiguity. Please return a non-2xx status directly from the DELETE path and update `TestClusterHandler_DeleteFile_Disabled`.

Validation completed on the rebased branch:

- `go test -race -count=1 ./internal/api ./internal/audit`
- `go test -count=1 ./internal/cluster/...`
- `go vet ./internal/api ./internal/audit`
- Live local-storage cluster: auth guard, confirmation guard, non-leader forwarding, physical unlink on every node, repeated-delete 404, quorum-loss 503/retry, recovery, warning fields, and persisted success/503 audit detail.

Thundercloud12:
@xe-nvdk hey so the changes are applied please have a look

xe-nvdk:
Awesome work! Thank you!

PR #831, opened 2026-09-15
fix(query): keep DECIMAL JSON values numeric

### LINKED ISSUE #818: JSON query endpoint returns DECIMAL columns as strings, unlike the arrow and msgpack endpoints
## Summary

`POST /api/v1/query` returns DECIMAL columns as JSON **strings**, while `/api/v1/query/arrow` and `/api/v1/query/msgpack` return them as numbers for the same SQL. Since DuckDB types `SUM(integer)` as `decimal(38,0)` and `AVG(...)` as `decimal(x,y)`, this affects most aggregate queries.

Found while building dashboarding in Launchpad, which consumes the JSON endpoint through its instance proxy.

## Cause

`internal/api/query_arrow.go:599` and the msgpack path both call `normalizeDecimalSchema(schema)` before encoding. `internal/api/query_arrow_json.go` does not — `grep -n "Decimal" internal/api/query_arrow_json.go` returns nothing.

So a decimal column reaches `writeArrowValue`, matches no case, and falls through:

```go
default:
    // Fallback: use ValueStr for any unhandled Arrow types
    writeJSONString(w, scratch, col.ValueStr(row))
```

The msgpack path already carries a comment describing this exact failure, and `query_msgpack_decimal_test.go` guards against a regression there:

> DuckDB returns SUM(integer) as decimal(38,0) and AVG as decimal(x,y). Without this the msgpack encoder has no `*array.Decimal128` case, so decimal columns fall through to `encodeFallbackColumn` and go out as msgpack *strings* while the `types` array advertises decimal(38, 0).

The JSON writer has the same hole and no equivalent test.

## Reproduce

```sql
SELECT avg(cpu) AS avg_cpu, sum(requests) AS total FROM metrics;
```

- `POST /api/v1/query`        → `{"data": [["3.14", "42"]]}`   ← strings
- `POST /api/v1/query/arrow`  → numbers
- `POST /api/v1/query/msgpack`→ numbers

Three endpoints, same SQL, two different types for the same column.

## Impact

A consumer that infers column types from JSON values sees an aggregate as a string column. For a charting client that means a query as ordinary as

```sql
SELECT $__timeGroup(time, '5m') AS time, avg(cpu) FROM metrics GROUP BY 1
```

yields no numeric column at all, and renders nothing — with no error, because nothing is malformed. Every workaround is a client-side heuristic ("this string looks like a number") that then has to decide what to do about HUGEINT values beyond 2^53.

The non-Arrow JSON writer looks affected too, differently: `mapColumnType` in `query_json_writer.go` maps DECIMAL to `colFloat64`, but the scanned Go value is `duckdb.Decimal{Width, Scale, Value *big.Int}`, which matches none of the `float64/float32/int64/sql.NullFloat64` cases and reaches `writeFallbackJSON` → `json.Marshal` → `{"Width":38,"Scale":0,"Value":42}`, an object in a row cell. Worth checking as part of the same fix.

## Suggested fix

Call `normalizeDecimalSchema` in `executeQueryArrowJSON` where the schema is read (`query_arrow_json.go`, at `schema := reader.Schema()`), mirroring `query_msgpack.go`. `castInfo` is nil when there are no decimal columns, so the common path is unaffected.

A test mirroring `query_msgpack_decimal_test.go` for the JSON writer would stop this recurring in the third writer.

## Notes

- Timestamps are consistent across all three writers and are not affected — the JSON writers emit `RFC3339Nano` in UTC.
- `rows_capped` / `truncated` are emitted correctly by the JSON writers; this is only about the decimal column type.

TayfurYldz:
## Summary

Fix DECIMAL wire-type parity for `POST /api/v1/query` across both JSON execution paths.

The Arrow-backed JSON path now reuses the existing decimal normalization/cast machinery already used by Arrow IPC and msgpack, so DuckDB decimals are emitted as JSON numbers rather than falling through to `ValueStr` strings. The database/sql JSON fallback now handles `duckdb.Decimal` directly and writes its exact base-10 representation as a JSON number instead of marshaling the driver struct into an object cell.

Closes #818.

## Tests

Regression proof before the production fix:
- Arrow JSON regression returned a string for DECIMAL and failed the numeric assertion.
- database/sql JSON regression returned the `duckdb.Decimal` struct as an object cell and failed the numeric assertion.

Validation after the fix:
- `go test -tags=duckdb_arrow ./internal/api -run 'TestStreamArrowJSON_DecimalsAreNumbers|TestStreamTypedJSON_DecimalsAreNumbers' -count=1`
- `go test -tags=duckdb_arrow ./internal/api -count=1`
- `go vet -tags=duckdb_arrow ./internal/api`
- `go test -tags=duckdb_arrow -race ./internal/api -count=1`
- `git diff --check`

All of the above pass.

I also ran `go test ./...`; the changed package passed, but the full run hit `internal/ingest/TestBufferMetrics_FlushCountersMove` once (`buffer_flushes_total did not advance`). An isolated rerun of that test then passed on both this branch and an untouched `upstream/main` worktree, so I did not change unrelated ingest code. CI remains authoritative for the full repository matrix.


TayfurYldz:

xe-nvdk (approved):
Welcome, and thank you: this is a first PR done the way this repo asks, down to the "regression proof before the production fix" section, which is exactly the evidence a reviewer wants. I went deep rather than take the description on faith:

- **The claim holds.** With `main`'s `query_arrow_json.go` and `query_json_writer.go` swapped in, both new tests fail with the issue's exact symptoms (`data[0][0] has type string (3)` and `map[Scale:0 Value:42 Width:38]`); on this branch they pass. The Arrow test covers both cast targets (scale 0 → int64, scale 2 → double).
- **Live binary.** Built with the arrow tag, two rows written over line protocol: `SELECT sum(v), avg(v), avg(w), count(*)` over `POST /api/v1/query` now comes back as `[7, 3.5, 3.5, 2]`, all JSON numbers.
- **The Arrow path is the same code msgpack and Arrow IPC use**, and it is used correctly: `castDecimalBatch` returns an owned record (`NewRecord` retains the columns, the helper releases its own casted references), and you release it once per batch on every exit — normal, row cap, flush failure. The `goto done` → `capped`/`stop` rewrite is equivalent on every path I traced (`reader.Err()` is still skipped on a cap, still consulted on a clean end, `rows_capped` is computed from the count as before). `castInfo` is nil without decimal columns, so the common path costs nothing.
- **The `database/sql` fallback** emits `duckdb.Decimal.String()`, which I checked against the driver source: `0` for zero, sign handled, trailing zeros trimmed, `0.05`-style padding, so every output is a valid JSON number token with the exact digits. That path is only reachable when the Arrow interface cannot be created, so the unit test is the right coverage.
- Build under both tag sets, gofmt, vet, and `go test -race` for `internal/api` are clean. CLA passes.

One consequence worth stating, and I added it to your release-notes entry while resolving the positional conflict (seven entries landed above yours since your base): the Arrow-path cast is a safe cast, so a `DECIMAL(38,0)` result outside the int64 range now ends the response with `truncated: true` and `decimal cast failed … integer value out of bounds` instead of a quoted string. I reproduced that on the binary with `sum(CAST(v AS DECIMAL(38,0)) * 4611686018427387904)`; it is the contract msgpack already has, so parity is the right call, but readers of the notes should know. The merge commit I pushed to your branch changes nothing else; CI is re-running on it.

The `TestBufferMetrics_FlushCountersMove` flake you saw in the full run is a known one, fixed on `main` yesterday; your branch predated the fix.

Issue #750, opened 2026-09-12
More lossy encodings that become paths, keys or authorization decisions


xe-nvdk:
Found while auditing #744. Filed together because they share one shape: a lossy transform whose output is then used as an identity. Ordered by severity.

## 1. Daily compaction flattens source keys into one temp directory

`internal/compaction/job.go:596`: `localPath := filepath.Join(tempDir, filepath.Base(fileKey))`, then `os.Create`, which truncates with no collision check.

The daily tier's candidate spans all 24 hour subdirectories (`daily.go:150-218`), and file basenames are `{measurement}_{YYYYMMDD}_{HHMMSS}_{nanos}.parquet` derived from wall-clock only (`ingest/arrow_writer.go:3668`, no node id, PID or random component). Two source files from different hours can share a basename. The result: one file's rows are duplicated into the output, the other's are dropped, and **both originals are marked compacted and deleted**.

Same shape as the already-closed #357.

## 2. RBAC dedup key case-folds while the permission check does not

`internal/api/query.go:1222` and `:1271` build the dedup key with `table := strings.ToLower(tableName)`, but the `TableReference` that reaches `CheckPermissionsBatch` keeps the original case, and `matchPattern` (`internal/auth/rbac_manager.go:1968`) compares case-sensitively.

So `SELECT * FROM cpu WHERE x IN (SELECT y FROM CPU)` checks the permission for `cpu` only, and reads both. Object keys are case-sensitive on S3, so those are two measurements with two separate grants.

Distinct from #740, which covers spoke identifiers rather than this.

## 3. `sanitizeDBForName`'s doc comment asserts an invariant that is false

`internal/compaction/manager.go:394-396` claims "no legal Arc database name may contain `.`". Its inputs are pseudo-databases from `expandNamespaces` (`manager.go:378`), and `validateSpokeID` (`internal/edgesync/receive.go:620-665`) rejects only a leading dot and `..`. So spoke `rocket.01` + database `telemetry` and spoke `rocket` + database `01.telemetry` both sanitize to `rocket.01.telemetry`.

Masked today because `jobID` also embeds the folded partition path, which differs. The comment should not be trusted as written.

## 4. `NewJob`'s fallback jobID uses the raw database name

[...]

PR #832, opened 2026-09-15
fix(rbac): preserve case in table reference deduplication (#750)

Thundercloud12:
## Summary

Fixes an RBAC authorization bypass split out of #750 (item 2), where table references differing only by case were deduplicated before permission checking.

## Problem

During query permission analysis in `internal/api/query.go`, `extractTableReferences` deduplicated simple table references (`FROM table`) and JOIN table references (`JOIN table`) using a lowercased key:

```go
tableName := resolve(sql[matchIdx[2]:matchIdx[3])
table := strings.ToLower(tableName)
...
key := "default." + table
```

However, Arc's RBAC pattern matching (`matchPattern` in `internal/auth/rbac_manager.go`) is **case-sensitive**, matching the case-sensitive semantics of object storage backends such as S3 and Linux filesystems, where `cpu` and `CPU` can represent distinct measurement paths.

When a query referenced both spellings, for example:

```sql
SELECT * FROM cpu WHERE x IN (SELECT y FROM CPU)
```

both references were folded into `default.cpu` during deduplication.

As a result, only one reference was passed to `CheckPermissionsBatch`. If a principal had permission for `default.cpu` but lacked permission for `default.CPU`, the query could be authorized even though DuckDB subsequently executed against both tables.

This allowed access to a measurement for which the principal had no permission.

## Fix

Preserve the original identifier case in the deduplication key:

```go
key := "default." + tableName
```

This change was applied to both extraction paths:

* `internal/api/query.go:1257` — simple `FROM` table extraction
* `internal/api/query.go:1299` — `JOIN` table extraction

Database-qualified references already preserved case and therefore required no change.

With this fix, `cpu` and `CPU` remain distinct references and are independently evaluated against the principal's permission grants.

### Pre-fix reproduction

* Verified that `TestExtractTableReferences_PreservesCaseInDedup` fails before the fix.
* Expected 2 table references, but only 1 was returned:

```text
expected 2 table references, got 1: [{Database:default Measurement:cpu}]
```

### Unit tests

Added/updated:

* `TestExtractTableReferences_PreservesCaseInDedup`

  * Verifies that:

```sql
SELECT * FROM cpu WHERE x IN (SELECT y FROM CPU)
```

preserves both `default.cpu` and `default.CPU`.

* Verifies that:

```sql
SELECT * FROM cpu JOIN CPU ON cpu.id = CPU.id
```

also preserves both references.

* `TestQueryRBAC_CaseSensitiveTableDedup`

  * Verifies that a principal granted only `default.cpu` receives `403 Forbidden` when a query also references `CPU`.
  * Covers both subquery and JOIN cases.

### Test suite & race detection

```bash
go test -race -run 'Test.*RBAC.*' ./internal/api
```

Result:

```text
ok github.com/basekick-labs/arc/internal/api 1.219s
```

### Code health

```bash
gofmt -l
go vet ./internal/api
```

Both completed cleanly.

### Release notes

Added the fix under:

```text
## Security fixes
```

in `RELEASE_NOTES_2026.09.2.md`, following the existing newest-first ordering.

## Related

Split out of #750.


Thundercloud12:
@xe-nvdk hey so this was one small issue took the recommended approach!

xe-nvdk (approved):
Correct, minimal, and proven. I traced the one question a two-line security fix has to answer, whether the permission check now matches what the query actually reads:

- Execution keeps identifier case end to end: the rewriter resolves the identifier as written (`resolveIdent` → `getStoragePath(ctx, "default", resolved)`; the lowercased copies feed only the CTE and system-table checks), `storage.GetStoragePath` concatenates `database + "/" + measurement` verbatim, ingest never folds measurement names, and `matchPattern` compares with plain prefix/suffix/equality. So `cpu` and `CPU` are two measurements on disk and the dedup key was the only place that pretended otherwise. Preserving case in the key is the fix that makes extraction agree with execution, which is the invariant this codebase learned the hard way on #487.
- Revert-run: with `main`'s `query.go`, `TestExtractTableReferences_PreservesCaseInDedup` fails with `expected 2 table references, got 1: [{default cpu}]` and `TestQueryRBAC_CaseSensitiveTableDedup` no longer gets its 403 for either the subquery or the JOIN shape; both pass on the branch. The full `internal/api` suite under `-race` is green, build and vet clean.
- Note under `## Security fixes` reads right. I pushed a one-line commit to your branch changing its Contributed-by link from the issue to this PR, nothing else.

Thanks for taking a slice of #750 cleanly rather than the whole list.

Issue #835, opened 2026-09-15
iceberg: expired snapshots' manifests are never reclaimed, and the live manifest list grows by one DELETED-only manifest per removal pass


xe-nvdk:
Found while fixing #633 (validator probes against iceberg-go v0.6.0).

## Two related growth problems in the Iceberg export metadata

**1. Orphaned manifests are never deleted.** `expireSnapshots` runs with `WithPostCommit(false)` (deliberately, #632: iceberg-go's post-commit hook would also delete Arc's data files). Nothing else deletes the manifest lists and manifests of expired snapshots, so every commit's `.avro` files stay in `<warehouse>/<ns>.db/<table>/metadata/` forever and are copied into every backup. Today that is one manifest list plus one small manifest per commit. After #633, a pass that re-registers a rewritten or restored path leaves behind one merged manifest the size of the table's live file list (~150-400 B per entry). Measured on a 319-file table: ~6.6 KB per ordinary commit, ~15 KB per merged commit, and the merged figure scales with the file count.

**2. The live manifest list grows with the removal history.** iceberg-go v0.6.0's snapshot producers inherit the parent's manifests verbatim, so a manifest that only holds `DELETED` entries stays in every later snapshot's manifest list. A table that adds and removes files (compaction, retention) accumulates one DELETED-only manifest per removal pass; readers open every one of them on every plan. Probe: 6 adds interleaved with 4 removes leaves 6 manifests in the current list, 4 of them DELETED-only. The #633 merge commit collapses the list when it runs, but ordinary passes do not.

## Fix shape

- A metadata-only orphan sweep in the exporter: compute the reachable set (manifest lists and manifests referenced by the snapshots of every metadata version still retained, plus the `v<N>.metadata.json` copies) and delete other `*.avro` files under the table's metadata directory older than a grace period. Data files must never be touched.
- For the manifest-list growth, either run the merge-append (see `manifestMergeOn` in `internal/iceberg/exporter.go`) periodically when the DELETED-only manifest count exceeds a threshold, or upstream a fix to iceberg-go so producers drop manifests with no live entries (Java Iceberg's `ManifestFilterManager` behaviour).

Not a correctness issue for readers; it is unbounded disk and backup growth plus slower planning on busy tables.

xe-nvdk:
[...]

Issue #834, opened 2026-09-15
delete API: WHERE keyword scan runs on raw text, so a string literal containing drop/delete/update is refused


xe-nvdk:
Found while running the #633 end-to-end check.

## Problem

`validateWhereClause` (`internal/api/delete.go`) applies `dangerousKeywordPattern.FindString(where)` to the raw WHERE text, before any string-literal masking. A literal that happens to contain a forbidden word is rejected as if it were SQL:

```
POST /api/v1/delete
{"database":"icedb","measurement":"m","where":"host = 'drop'","confirm":true}

{"success":false,"error":"WHERE clause contains forbidden keyword: DROP"}
```

Any tag or field value such as `status = 'delete-pending'`, `action = 'update'`, `kind = 'alter'` is undeletable through the API. The value is data, not SQL.

## Fix shape

Mask string literals with the shared `sqlutil` masker (the same normalization the query path uses before its own gating) and run the keyword and prefix scans on the masked form. Keep the unmatched-quote check on the raw text. Add a regression test with a literal containing each forbidden keyword, plus one proving `'; DROP TABLE x --` outside a literal is still refused.

Not a security issue: the scan errs on the side of refusing. It is a false-positive usability bug in a feature that is off by default (`delete.enabled=false`).

xe-nvdk:
## Implementation notes (read before starting)

This validator is an injection boundary, so the fix must reuse the existing hardened masker rather than add a new regex:

- Use `sqlutil.MaskStringLiterals(where, sqlutil.HasQuotes(where))` from `internal/sql/mask.go`. It already masks plain `'…'`, escape-string `E'…'`, and dollar-quoted `$tag$…$tag$` literals (see the SECURITY comments in that file, added for #640). **Do not write a local quote-matching regex** — a `'…'`-only matcher misses the other two forms and turns this usability bug into a bypass.
- Run the keyword scan (`dangerousKeywordPattern`), the prefix scan, and the punctuation scan on the **masked** text. The raw `where` stays what gets interpolated into the DuckDB statement; the masked form is for validation only. That is the same pattern the query path uses.
- Keep the unmatched-quote and unmatched-parentheses checks on the raw text.

Tests to add in `internal/api/delete_test.go` (or the nearest existing delete test file):

1. A literal containing each forbidden keyword is accepted: `host = 'drop'`, `status = 'delete-pending'`, `action = 'update'`, and so on for the full list in `dangerousKeywordPattern`.
[...]

PR #836, opened 2026-09-15
fix(iceberg): re-register files the delete API rewrote in place (#633)

### LINKED ISSUE #633: high(iceberg): in-place parquet rewrite by row-level DELETE is invisible to the path-only fingerprint — external engines keep serving deleted rows

xe-nvdk:
## Summary

- The row-level delete API's partial-WHERE branch rewrites a Parquet file in place (same path, fewer rows, smaller size). The Iceberg reconciler keyed both its change detection and its diff on paths alone, so the rewritten file was never re-registered and the manifest kept the original `record_count` and `file_size_in_bytes` forever: engines answering `COUNT(*)` from manifest statistics kept counting deleted rows, and engines that trust `file_size_in_bytes` for split planning could fail on the shorter file.
- `FileRef.SizeBytes` is now populated by the storage-walk source (one `os.Stat` per listed file; a file that vanished between the listing and the stat is dropped from the pass), the scheduler fingerprint hashes path + size, and the exporter diff keys on (path, size). A path present on both sides with a different size is dropped and re-registered inside one transaction, so readers never see it absent.
- Re-adding a path is something iceberg-go v0.6.0 refuses outright: it carries every removed file's `DELETED` manifest entry in all later snapshots and checks new adds against those entries too ("cannot add files that are already referenced by table"), and forcing the add with the duplicate check off wedges every later removal of that path. For the add that hits that refusal, and only for it, the exporter enables Iceberg manifest merging (`commit.manifest-merge.enabled`, min-count 2, 1 GiB target) in the same transaction, retries with the duplicate check off, and switches merging back off before the commit. The merge-append rewrites the table's manifests into one and keeps `DELETED` entries only of its own snapshot, so the stale history is gone once the commit lands. This also covers a backup restore of a key that had left the table, which failed the same way.
- Ordinary passes changed shape slightly: removes and adds are now applied as an overwrite snapshot followed by a fast-append snapshot in the same commit, instead of one overwrite snapshot; cost per pass is unchanged. Merging is never left on the table because with it every append would rewrite the full manifest set and Arc never reclaims superseded manifests (#835 filed with numbers).
- Known limit: a rewrite that lands on exactly the same byte size at the same path is still invisible.

Closes #633.

## Design notes

The obvious fix, remove then re-add the path, was validated adversarially and found impossible on stock iceberg-go v0.6.0 in every variant (same transaction, two commits, DataFile-based removal). The alternative of making the delete rewrite land at a new path was rejected: it turns every delete into a compaction-shaped event for edge-sync (the discoverer would ship the new file to the hub, duplicating the surviving rows there), tiering (`tier_files` is only populated by the ingest writer), the Raft manifest, and the filename-driven daily-compaction cooldown and file-time pruning. Manifest merging is a standard Iceberg table property, and the probes behind this design (single-commit rewrite, later removals, restore of a removed key, upgrade of a table with accumulated `DELETED` history, snapshot expiry) all read correctly in DuckDB.

Cluster note: the exported table reflects the compactor node's copy of the file. The delete handler updates the Raft manifest entry after a rewrite and peers re-pull on size mismatch, so copies converge; that manifest update is warn-only on failure, unchanged by this PR.

## Test plan

- [x] `gofmt -l`, `go vet`, `go test -race ./internal/iceberg/...` green.
- [x] Revert-run: with main's `exporter.go`, `scheduler.go`, `source.go` swapped in, all eight new tests fail for the intended reasons (manifest still describes the old file; pass skipped as unchanged; `SizeBytes = 0`; re-add refused as "already referenced"; merge property absent).
- [x] Live on the branch binary (`iceberg.enabled`, 3 s reconcile interval, `delete.enabled`): 200 rows in one file, partial delete of 50 rows rewrites the file 3153 → 1299 bytes; two intervals later the manifest entry reads `record_count=150, file_size_in_bytes=1299`, one merged manifest, `reregistered=1` in the log; DuckDB `iceberg_scan` and `iceberg_metadata` and PyIceberg (`SqlCatalog` + `inspect.entries()`) all agree. Same script on the main binary keeps `record_count=200` forever.
- [x] Probe harness against iceberg-go v0.6.0 (not in the tree): remove + ignoreDuplicates add in one transaction with merging on leaves a single manifest with no `DELETED` entries; a later removal of that path succeeds; a table created without the properties and carrying six manifests collapses to one on the merged commit; restore of a removed key works; ExpireSnapshots unaffected.

Follow-ups: #835 (orphan manifests and manifest-list growth), #834 (WHERE literal false positive found while running the e2e).


xe-nvdk:
Docs companion: Basekick-Labs/docs.basekick.net#71 (Apache Iceberg page: path-and-size diff, delete API FAQ).

PR #839, opened 2026-09-15
fix(backup): back up an Iceberg warehouse outside the storage root and stop the restore wedge (#637)

### LINKED ISSUE #637: medium(backup): iceberg.warehouse outside the storage root is silently omitted from backups — DR restore wedges every measurement

xe-nvdk:
## Summary

- Backup found Iceberg table metadata only by filtering the data-storage listing, so a warehouse configured outside `storage.local_path` was never copied. The backup still reported success while carrying catalog rows whose `metadata_location`s pointed at files it did not hold.
- The backup manager now receives the warehouse path (a new `iceberg.LocalWarehousePath` turns `file://` URIs and bare paths into an absolute directory and rejects other schemes) and classifies it against the storage root with symlinks resolved and a path-boundary match. A warehouse outside the root is walked on the filesystem, scoped to Arc's own layout (`<prefix>_<db>.db/<table>/metadata/<file>`, dot-prefixed names and symlink entries skipped, any other walk error fatal), copied under `<backup_id>/iceberg/<rel>`, and recorded in the manifest as `iceberg_warehouse` with the source path, file count and bytes.
- Restore writes those objects back into this node's configured `iceberg.warehouse` whenever `restore_data` or `restore_metadata` is set, through a temp file and rename with 0600/0700 modes, refusing any key that is not a clean relative path. A node with no outside-root warehouse skips them, reports `iceberg_warehouse_files_skipped` on the status endpoint, and logs that the catalog stores absolute paths so the target's warehouse has to be the backup's path.
- The SQLite snapshot now precedes both Iceberg metadata copies, so a reconcile commit during a long backup cannot leave a row pointing at a file the backup lacks.
- `EnsureTable` no longer falls through to `CreateTable` when a catalog row exists but cannot be loaded. On a restored host that used to write a fresh metadata file into the warehouse and then fail on the catalog's primary key, on every pass, forever. A missing metadata file now fails the pass with an error naming the table, the cause, and the two ways out.

Closes #637.

## Test plan

- [x] `gofmt -l`, `go vet`, `go test -race ./internal/backup/... ./internal/iceberg/...` green.
- [x] Revert-run: with the wiring disabled and main's `EnsureTable`, the four regression tests fail for the intended reasons (no `iceberg/` objects and no manifest field; symlinked warehouse backs up nothing; restore skip counter stays 0; `EnsureTable` falls through to the UNIQUE-constraint error and writes an orphan metadata file). The under-root and path-helper tests are guards.
- [x] Live DR on the branch binary with an outside-root warehouse, in three spellings (`file://` URI, bare path, symlinked directory): write, wait for the table, `POST /api/v1/backup` → backup holds `iceberg/arc_icedb.db/m/metadata/*` and the manifest field; wipe data, warehouse and SQLite; fresh start, `POST /api/v1/backup/restore`, restart → warehouse files back, the reconciler runs without errors and writes no orphan metadata, PyIceberg reads 20 rows through the restored catalog. Same script on the main binary: no `iceberg/` group, empty warehouse after restore, the `UNIQUE constraint failed` loop every 3 s with a new metadata file per pass, and a catalog reader fails with `FileNotFoundError`.

Docs: companion PR on docs.basekick.net (backup structure, restore behaviour, absolute-path and cluster caveats).


xe-nvdk:
Docs companion: Basekick-Labs/docs.basekick.net#72.

PR #840, opened 2026-09-15
fix(mqtt): validate database, mapping targets and measurement names before they reach the buffer (#300)

### LINKED ISSUE #300: high(mqtt): No database name validation on MQTT topic mapping

xe-nvdk:
## Summary

- `ValidateCreateRequest` built the temporary subscription it validates without `topic_mapping`, so `POST /api/v1/mqtt/subscriptions` accepted targets such as `../other-db`, `a/b` or an empty string that `PUT` already refused. The maintainer's re-verification on the issue describes the class: the storage layer's containment check holds, but the value becomes a storage-key segment. Traced on main, the shapes differ: a slash-containing target is written under a mis-partitioned path no query can address; a `..` or empty target is accepted into the buffer and refused at flush, which latches the buffer's flush-failure state (WAL retained, non-clean shutdown).
- The review turned up the publisher-controlled sibling: a payload measurement such as `{"m":"../x"}` reached the buffer the same way. `mapToRecord` now refuses a measurement that fails the HTTP write path's rule as a decode error.
- Create and update now apply the same rule to mapping targets; `startSubscriber` refuses a persisted row whose targets fail it (edited in SQLite, or written before the rule), leaving it in the `error` status with the reason; `onMessage` re-checks the resolved database per message as the last line and drops offenders before decoding, counted in `messages_failed` and the MQTT failed-messages metric, logged once per value per subscriber run.

Closes #300.

## Test plan

- [x] `gofmt -l`, `go vet`, `go test -race ./internal/mqtt/...` green.
- [x] Revert-run against main's code: `TestValidateCreateRequest_TopicMappingValidated` fails (six bad targets accepted); `TestOnMessage_DropsMessageForInvalidDatabase` reaches the nil buffer (guard absent); `TestMapToRecord_RejectsInvalidMeasurement` and `TestManager_StartSubscriber_RejectsInvalidPersistedTargets` fail (accepted / started).
- [x] Live with a local mosquitto on a scratch port: create with `topic_mapping: {"t/x": "../other-db"}` → 400 with the validator's message (main: 201). A valid subscription ingests one message; two messages with `m` set to `../x` and `a/b` are counted failed with no flush failure. After editing the persisted mapping in SQLite to `../other-db` and restarting, the start attempt is refused and the row shows the `error` status with the reason; nothing is written outside the expected databases.

Docs: companion PR on docs.basekick.net (MQTT page, topic-mapping and measurement rules).


xe-nvdk:
Docs companion: Basekick-Labs/docs.basekick.net#73.

PR #841, opened 2026-09-15
fix: ignore SQL tokens inside delete string literals

### LINKED ISSUE #834: delete API: WHERE keyword scan runs on raw text, so a string literal containing drop/delete/update is refused

hizlidepoo:
Closes #834.

## Summary
- mask SQL string literals before DELETE WHERE keyword, prefix, and punctuation scans
- keep raw WHERE text for unmatched quote/parenthesis checks and execution
- cover all forbidden keywords plus E-string, dollar-quoted, punctuation, and outside-literal rejection cases
- add the requested 2026.09.2 bug-fix release note

## Validation
- reproduced on current main 6233e7c6a9965ed4da4db75bcf4233e092ad3f64 with the focused regression test
- go test -tags=duckdb_arrow ./internal/api -run TestDeleteWhereIgnoresDangerousTokensInsideStringLiterals -count=1 — PASS
- go test -tags=duckdb_arrow -race ./internal/api/... — PASS
- go vet -tags=duckdb_arrow ./internal/api/... — PASS
- git diff --check — PASS


xe-nvdk (changes requested):
Thanks for taking #834; the shape matches the implementation notes (shared masker, keyword and punctuation scans on the masked form, raw text for the quote and parenthesis counts), and the regression test covers every forbidden keyword.

Requesting changes for three things:

1. **Hold for a masker fix.** While verifying this I found an edge case in the shared `sqlutil.MaskStringLiterals` where its view of a literal can differ from DuckDB's parse. On main that does not matter for the delete path, because the keyword scan runs on the raw text; once the scan runs on the masked form, the edge case becomes reachable. I am landing a fix for the masker separately (shared code, and it needs its own review). Once it merges, please rebase onto it and add the regression case I will point you to as a must-reject in your test.
2. **Release note format.** Entries in `RELEASE_NOTES_2026.09.2.md` are `### <title> ([#834](...))` headings followed by a paragraph, not a bare paragraph under `## Bug fixes`; please match the neighbouring entries and end with `Contributed by @hizlidepoo.` (see the other contributor entries for the wording).
3. **CLA.** The CLA check is failing; follow the bot's comment to sign.

The `DIRTY` state is a release-notes conflict with main; it resolves on the rebase.

xe-nvdk:
The masker fix has landed on `main` (68ab3290), so this is unblocked. Three things to do:

1. **Rebase onto `main`.** That also resolves the release-notes conflict showing as `DIRTY`.
2. **Add a must-reject case to your test.** `sqlutil.MaskStringLiterals` now follows DuckDB's rules for each quoted form, and the delete validator inherits that once your patch routes through it. Please add these to the `invalid` list, each of which must still be refused:
   - `host = '\' OR 1=1; DROP TABLE x -- '`
   - `host = 'a' -- '` followed by a newline and `; DROP TABLE x`
   - `host = 'a' AS t$$$ ; DROP TABLE x`
   The first is the one that matters most for your change: before the masker fix, a literal ending in a backslash swallowed the rest of the statement, so a scan over the masked form would have accepted it.
3. **Release-note format.** Entries in `RELEASE_NOTES_2026.09.2.md` are a `### <title> ([#834](...))` heading followed by a paragraph, not a bare line under `## Bug fixes`. Match the neighbouring entries and end with `Contributed by @hizlidepoo.`

The CLA check is still failing; follow the bot's comment to sign, otherwise this cannot merge.

The approach in your patch was right, and the masker weakness it surfaced was pre-existing rather than anything you introduced. Thanks for the clean reproduction steps in the description; they made it quick to verify.

xe-nvdk (approved):
Reviewed the updated implementation and found no blocking issues. String literals are masked before the DELETE WHERE keyword, prefix, and punctuation scans, while the raw validation checks remain intact. The requested must-reject regression cases and release-note attribution are included. The full race-enabled test suite and an HTTP-level delete test passed.

Approved from my side. @hizlidepoo, please sign the CLA so the remaining check can pass and we can complete the merge process. Thank you for the contribution!

PR #842, opened 2026-09-15
fix(query): track Arrow IPC queries in the query registry and slow-query log (#309)

### LINKED ISSUE #309: high(query): Arrow IPC endpoint bypasses enterprise features
query_arrow.go:118-153 — no registry, governance, slow query logging. Fix: add integrations.

xe-nvdk:
## Summary

- The March audit listed the Arrow endpoint as bypassing enterprise features. Since then it gained RBAC, query governance, the row-cap trailer, the catch-up gate and request metrics; what was still missing was the query registry and slow-query logging. An Arrow query never appeared in `GET /api/v1/queries/active` or `/history`, could not be cancelled through `DELETE /api/v1/queries/:id`, carried no `X-Arc-Query-ID` header, and never counted as slow.
- `executeQueryArrow` now registers every query (after the arcx hook declines, so that untracked experimental path stays as it is), returns `X-Arc-Query-ID`, derives its execution context from the registry so a cancel reaches DuckDB, and records the same dispositions as the JSON endpoint: completed with the row count, failed with the sanitized cause, timed out, or cancelled, with the governance row cap recorded first so a capped-then-truncated result is visible in history. The panic recovery closure fails the entry too, so a panicking query does not stay listed as running.
- Because go-duckdb materializes an Arrow result inside `QueryContext`, an operator cancel during execution ends the request with `500 Query cancelled` rather than a truncated stream; a cancel during the short streaming phase ends the stream with the truncation trailer. A deadline that fires mid-stream now counts in `arc_query_timeouts_total` (only the pre-stream timeout did before), and `query.slow_query_threshold_ms` covers Arrow queries.
- The `ArrowQueryContext` error branch reads the context error before releasing the context. The deep review found the JSON endpoint's two error branches have had that ordering wrong all along: `cancelTimeout()` / `cancel()` ran first, so with the default timeout every plain execution failure read as `context canceled`, was filed as "already cancelled", and stayed listed as running in the registry forever. Both branches now read the cause first. No new test pins the JSON case: a failing partitioned execution is not reproducible in the handler harness, so the change is covered by the trace in the review and by the Arrow test of the same shape.

Closes #309.

## Test plan

- [x] `gofmt -l`, `go vet -tags=duckdb_arrow`, `go test -tags=duckdb_arrow -race ./internal/api/` green.
- [x] Revert-run against main's `query_arrow.go`: the four regression tests fail as intended (header missing; history empty after a failure; a registry cancel never reaches DuckDB and the query ends as timed out after the 15 s backstop; slow-query counter stays at 0). The no-registry guard passes on both. The cancel test runs with and without a configured timeout, so the cancel-only context arm is covered.
- [x] Live on the branch binary with the smoke license and `query_management.enabled=true`, `query.slow_query_threshold_ms=1`: an Arrow query returns `X-Arc-Query-ID` and shows up in history as completed with its row count; the slow-query log line fires for it; a long Arrow query is listed under `/queries/active`, `DELETE /queries/:id` ends it within a second with `500 Query cancelled`, and history shows it cancelled.

Docs: companion PR on docs.basekick.net (query management page, API reference on both product docs, monitoring callout).


xe-nvdk:
Docs companion: Basekick-Labs/docs.basekick.net#74.

PR #844, opened 2026-09-15
fix(iceberg): keep a compacted file out of the table until its compaction has committed (#638)

### LINKED ISSUE #638: medium(iceberg): compaction's upload-then-delete window commits snapshots with duplicate data (~2x rows transiently)

xe-nvdk:
## Summary

- Compaction uploads the compacted file into the partition before deleting its sources, the right order for Arc's crash safety. A reconcile pass that listed the partition in that window registered the compacted file and every source it replaced in one snapshot; external readers saw roughly twice the rows for that hour or day until a later pass removed the sources, and with compaction running across many measurements this happened in steady state.
- The reconciler now consults compaction's crash-recovery manifests (`_compaction_state/`) per measurement, between the storage listing and the stat loop, fresh from storage, and leaves out any compacted output whose manifest exists and whose source files still exist (`ManifestManager.PendingOutputsUnder`). The output lands on the pass after the sources are gone, in one snapshot with their removal.
- The read order is what makes it race-free: every path (job, hub finalize, recovery) deletes the manifest only after every source is deleted. A manifest absent at the read means the sources vanish from the same pass; a manifest present means the output is held back. An output whose inputs are all gone counts as committed even if its manifest lingers, so a manifest that could not be deleted, or hub receipts not yet marked, never hide data. A manifest listed but gone by the read is absent; any other read failure fails that measurement's pass and is retried.
- The lookup is wired through its own manifest manager even when compaction is disabled (a manifest from an earlier run can still be in storage), and `_compaction_state` is no longer enumerated as a database.
- A manifest that reads but does not parse (a zero-length file left by a crash) is skipped and reported once per path rather than failing every measurement on every pass: it names nothing, and recovery deletes it. A storage read failure still fails the measurement's pass.
- Remaining under-count, bounded by the compaction schedule: a compaction that deleted some sources and failed on the rest keeps its output hidden while the surviving sources stay exported, until the next compaction cycle's recovery finishes the job (no recovery runs while compaction is disabled). A commit that lands in the milliseconds between the manifest read and the file stats leaves that partition out for one pass.

Closes #638.

## Test plan

- [x] `gofmt -l`, `go vet`, `go test -race ./internal/compaction/... ./internal/iceberg/...` green.
- [x] Revert-run with the exclusion disabled: `TestFilesAndLocal_ExcludesPendingCompactionOutputs` and `TestScheduler_CompactionWindowNeverDuplicates` fail (pending output not excluded; the mid-compaction pass registers all three files). The two ordering probes pin the read order on both sides: one commits the compaction inside the lookup and expects only the output, the other uploads inside the lookup and expects only the raws. `TestPendingOutputsUnder` covers the manifest side: in-flight vs committed by inputs, prefix scoping, a manifest deleted between listing and read, and unparseable manifests skipped. A test pins the `_compaction_state` literal against `compaction.ManifestBasePath`.
- [x] Live on the branch binary: Iceberg export at a 1 s interval with hourly compaction at two files; three flushes (file names aged so the partition is eligible), a compaction cycle, then every Iceberg snapshot inspected through PyIceberg: none holds a source and the compacted file together, and the final snapshot holds the compacted file only. The non-default `compaction.enabled=false` row also ran: the lookup is wired, a write is exported, no errors.

Docs: companion PR on docs.basekick.net (Iceberg FAQ).


xe-nvdk:
Docs companion: Basekick-Labs/docs.basekick.net#75.

PR #871, opened 2026-09-16
fix(audit): copy Fiber request data before asynchronous logging

### LINKED ISSUE #837: audit: middleware stores zero-copy Fiber strings in events that a background writer serializes later
Found while reviewing #830.

## Problem

`internal/audit/middleware.go` builds an `AuditEvent` from `c.Path()`, `c.IP()`, `c.Get("User-Agent")`, `c.Get("x-arc-database")`, `c.Params(...)` and `c.Query("db")`, then hands the event to `Logger.LogEvent`, which only enqueues the pointer on a channel. The batch writer (`writerLoop`) serializes the fields up to a second later.

Arc runs Fiber with `Immutable=false` (see the note in `internal/api/msgpack.go`), so every one of those accessors returns a string that aliases a pooled buffer: `c.Query` is `app.getString(QueryArgs().Peek(key))`, `c.Path()` is `app.getString(c.pathBuffer)` on the pooled `Ctx`, and `c.Get` reads the request header buffer. Once the handler returns, the connection's `RequestCtx` and the `Ctx` go back to their pools and the next request overwrites those bytes. An audit row can therefore record a later request's path, database or user agent, or garbage. #830 adds `path` and `reason` from `c.Query` to the same event through `audit.DetailLocalsKey`, which the review asks to clone in the handler.

This is the same class as the retained `c.Get`/`c.Query` aliasing found in the edge-sync receive path (#776/#783 reviews).

## Fix shape

Copy every request-derived string when the event is built: `utils.CopyString` from `github.com/gofiber/fiber/v2/utils` on `path`, `c.IP()`, the user agent (after truncation), database, measurement, and any map supplied through the locals key (copy keys and values into a fresh map). `tokenInfo.Name` comes from the auth cache and needs no copy. Add a test that sends two different requests over one connection with `app.Test` disabled in favor of a real listener, or unit-test the copy by mutating the source buffer after the event is built and asserting the event is unchanged.

Small, self-contained, and a good first issue once the recipe above is followed exactly.

0utsights:
Fixes #837.

The audit middleware copies request-backed strings before enqueueing each event, including the HTTP method, and takes an owned copy of handler detail maps and their string keys and values. Header/parameter/query precedence, the User-Agent limit, and actor/classification behavior are preserved.

The regression runs the real middleware, then mutates the original borrowed buffers after enqueue. It covers method, path, proxy IP, database, measurement, User-Agent, and detail keys/values, as well as later map mutation, empty metadata, and skipped requests. Without the method copy, all four lifetime cases change `event.Method` from `POST` to `xxxx`; they pass with the fix.

Rebased onto `6d611cd838ce7487420ba173b8f7125f73c275ad`, retaining the detail-map support now on `main`. Release notes preserve the upstream file with LF endings and add only this contribution's entry.

Validation on this revision with Go 1.26.6 on Windows (`CGO_ENABLED=0`):

- `go test -tags=duckdb_arrow ./internal/audit/... -run '^TestMiddleware(OwnsRequestDataBeforeEnqueue|SkipsExcludedAndReadPaths)$' -count=1` passed.
- `go vet -tags=duckdb_arrow ./internal/audit/...` passed.
- Independent review reran both checks and verified the method-buffer regression, upstream preservation, and LF-only release notes.
- `gofmt` and `git diff --check` passed.

Linux build, race, and object-storage checks could not be rerun for this rebased revision because the local Docker daemon failed to start. The existing SQLite audit tests require cgo. Full Linux validation remains for CI; the earlier PR version's Linux results are not claimed for this revision.


xe-nvdk (changes requested):
Thanks for the focused fix. The displayed diff is currently much larger than the actual change because `RELEASE_NOTES_2026.09.2.md` was converted from LF to mostly CRLF. This produces more than 4,000 noisy changed lines and conflicts with updates already on `main`. Please rebase onto the latest `main`, restore that file from `main`, and reapply only this PR’s release-note entry while preserving LF endings. The resulting semantic diff should be approximately +147/-2.

There is also one request-backed value still being retained without an owned copy: `method := c.Method()`. Fiber returns this from the reusable request buffer just like the other accessors. A regression that mutates the method buffer after enqueue currently changes `event.Method` from `"POST"` to `"xxxx"`. Please copy it with `utils.CopyString` and extend the lifetime test to cover it.

Finally, please sign the CLA so its check can pass. Once these items are addressed, the PR will be small and ready for full review.

0utsights:

Issue #876, opened 2026-09-16
cluster: the compactor lease never moves to a dedicated compactor, and there is no way to move it by hand


xe-nvdk:
Found while reviewing the chart fix for #870.

The compactor lease is deliberately never preempted. `CompactorFailoverManager`'s own doc comment states it:

> - RoleCompactor nodes are preferred for the initial claim
> - On failover, a RoleWriter can hold the lease without restarting
> - When the original compactor returns, the current holder keeps it (no preemption — the lease is stable once assigned)

`checkCompactorHealth` only reassigns when the current holder is **unhealthy**. A healthy writer holding the lease keeps it forever, even once a healthy `RoleCompactor` node is present.

Stability is a reasonable default. The problem is that there is no way to move the lease deliberately.

**How a cluster ends up in this state.** Two routes, neither exotic:

1. The compactor pod is slower to join than the leader's first assignment tick, so `tryInitialAssignment` finds no `RoleCompactor` node and falls back to a healthy writer. The compactor joins a moment later and never gets the lease.
2. Fixing #870. Every existing chart cluster has a writer holding the lease precisely because no compactor was ever visible. After that upgrade the compactor pod joins for the first time and the lease stays where it is.

In both cases a node provisioned for compaction, with its own CPU and memory budget and its own PVC, sits idle while a writer does the work on top of ingest.

**There is no operator lever.** `CompactorFailoverManager.TriggerManualFailover` exists and is exercised by its own unit test, but nothing routes to it: there is no API endpoint, no CLI verb, and no config. The only way to move the lease is to make the holder unhealthy, i.e. restart the writer, which is a blunt instrument on a node that is also taking ingest.

**Fix shape, in order of preference:**

- **Preempt toward a configured compactor.** When the lease holder is not `RoleCompactor` and a healthy `RoleCompactor` node exists, hand the lease over once, on a cooldown. This is what an operator who deployed a dedicated compactor already expects. It needs a guard so a flapping compactor cannot ping-pong the lease.
- **Expose the manual trigger.** An admin endpoint that reassigns the lease, either to a named node or to the best candidate. Smaller change, and useful regardless of whether preemption lands, because it also covers "move compaction off this node right now".
- At minimum, surface the holder and whether it is a dedicated compactor in the cluster status endpoint, so the state is visible rather than inferred from a log line at assignment time.

Related: #870 (the chart bug that guarantees every existing cluster is in this state after upgrading).

Issue #856, opened 2026-09-16
cluster: readers are not promotion candidates, so the documented 1-writer + N-reader Pattern 1 cannot fail over


xe-nvdk:
Found while fixing #850 (no initial primary is elected). Separate defect, same feature.

The clustering docs describe Pattern 1 as "single-writer + multi-reader", say the readers are the failover pool ("Deploy 1 writer + 2+ readers... the readers are the failover pool"), and promise that on writer loss "the Raft leader selects the most caught-up reader (by replication LSN)" and promotes it. The Enterprise Helm chart's Pattern 1 example is one writer and three readers with `failover.enabled: true`.

The code cannot do that. `WriterFailoverManager.selectNewPrimary` only considers `Registry.GetStandbyWriters()` and `Registry.GetWriters()`, and both filter on `Role == RoleWriter`. A `RoleReader` node is never a candidate, and nothing anywhere changes a node's role. So in the documented topology, when the single writer dies there is nothing to promote and the cluster has no writer at all. Nor is any LSN comparison implemented; selection is first-match over the writer list.

With #850 fixed, that single writer is now elected primary at startup so retention and CQ run, but its loss is still terminal for ingest until an operator intervenes.

Fix shape (needs design): decide whether Pattern 1 HA means promoting a reader to writer at runtime, which requires a role change through Raft plus a defined WAL catch-up and cutover, or whether Pattern 1 should require two or more writer-role nodes and the docs and chart example should say so. The second is much cheaper and matches what the code already supports. Either way the docs and the chart must stop promising reader promotion until it exists.

PR #877, opened 2026-09-16
fix(helm): the compactor pod never joined the cluster (#870)

### LINKED ISSUE #870: helm: the compactor pod has no Raft address, so it never joins the cluster
Found while reviewing the chart changes for #856.

The Enterprise chart's compactor pod never joins the cluster. `compactor-statefulset.yaml` sets `ARC_CLUSTER_ADVERTISE_ADDR` (the coordinator address, port 9100) but sets no Raft address at all:

```yaml
# compactor-statefulset.yaml, env block
- name: ARC_CLUSTER_ROLE
  value: "compactor"
- name: ARC_CLUSTER_ADVERTISE_ADDR
  value: "$(POD_NAME).<compactor-headless>.<ns>.svc.cluster.local:9100"
- name: ARC_CLUSTER_SEEDS
  value: ...
```

The writer template includes `arc-enterprise.writerRaftEnv` plus its own `ARC_CLUSTER_RAFT_ADVERTISE_ADDR`, and the reader template sets `ARC_CLUSTER_RAFT_ADVERTISE_ADDR` directly. The compactor gets neither.

With `cluster.raft_advertise_addr` empty, Arc falls back to `cluster.raft_bind_addr`, which defaults to `:9200`. That resolves to a nil IP, and the Raft transport refuses to stand up on an address it cannot advertise:

```
internal/cluster/security/raft_plain.go:28
errNotAdvertisable = errors.New("local bind address is not advertisable")
```

`raft.NewNode` then fails, `NewCoordinator` returns an error, and `cmd/arc/main.go` logs "Failed to initialize cluster coordinator - running in standalone mode" and carries on. The compactor pod runs as an isolated OSS node: not a Raft voter, not in any registry, invisible to every other node.

This is not a new failure mode. The reader template carries a comment documenting exactly this, added when the same gap was fixed for readers:

> Readers join the Raft cluster as voters. The Raft transport requires an advertisable bind address even on readers; without this the node fails to start with "local bind address is not advertisable" and falls back to OSS mode (no cluster state, no replicated auth, no peer file replication).

The compactor was never given the same treatment.

Knock-on effects on a default chart install:
- The compactor is not in the registry, so `checkCompactorElected` sees zero `RoleCompactor` nodes and logs "No compactor elected" once a minute, forever.
- Phase 5 compactor failover has no compactor to lease to.
- The compactor writes compacted files but cannot register them in the Raft manifest, because it has no Raft.
- Any guidance to size a cluster by counting pods is off by one, since this pod is not a voter.

Fix: give the compactor template the same Raft env the reader has — `ARC_CLUSTER_RAFT_ADVERTISE_ADDR` pointing at its own pod DNS name on 9200 — and a `raft` containerPort for parity with the writer. Worth a rendered-manifest test that asserts every role's pod carries a Raft advertise address, so the next role added does not repeat this.

xe-nvdk:
Closes #870. Found while reviewing the chart changes for #856.

Every Arc node joins Raft, whatever its role. The Enterprise chart gave the compactor pod a coordinator address but no Raft address, so `Coordinator.Start()` failed on an address it could not advertise and `main.go` logged "Failed to start cluster coordinator - running in standalone mode" and carried on. The pod passed its health checks throughout.

**The damage was not an idle compactor.** A node outside the cluster falls back to its configured role to decide whether to compact, and its role says yes. So it compacted the shared bucket on its own schedule while the node holding the compactor lease compacted it too. Two compactors on one bucket is the duplicate-output hazard the single-compactor design exists to prevent.

## Summary

- **The compactor advertises its own pod DNS name on the Raft port**, the same way the writer and reader already do.
- **The Raft helper now applies to every role.** It was named `writerRaftEnv` and its comment said Raft was writer-only, which is the belief that produced this bug twice. Renamed and included by all three roles, so each binds its Raft port explicitly instead of relying on Arc's own default happening to equal the writer's explicit value. That is the `iceberg.warehouse` shape from #534: two expressions equal only because a default was never changed.
- **The reader and compactor headless services publish the Raft port.** SRV-record consistency only; per-pod A records never depended on it.
- **`values-local-storage.yaml` asked for one writer**, so following the Pattern 1 guide silently overrode the default of three back down to one. Raised to three, matching the default values and the shared-storage preset. This is a gap in what #873 shipped.
- **A rendered-template check in CI** asserts every clustered role carries a Raft address and every headless service publishes 9200, against the default values and both presets.

## Upgrade consequence, stated plainly

Applying the chart restarts the compactor pod, after which it joins the cluster and stops compacting behind its back. **It will not start compacting on the cluster's behalf.** The compactor lease is deliberately never preempted, and on an existing cluster a writer holds it, chosen precisely because no compactor was visible. The dedicated pod idles until the lease moves, and the only way to move it is to restart the writer holding it. Filed as #876, which also notes that `TriggerManualFailover` exists but has no route.

## Configuration matrix

|---|---|---|
| Default install (`storage.mode=shared`) | **Yes.** `$(POD_NAME)` expands because `commonClusterEnv` declares it earlier in the same container's env list, and Kubernetes expands `$(VAR)` against earlier entries. Same shape as the coordinator address on that pod. | `serviceName` matches the headless service whose DNS name is advertised; the selector matches the pod labels. `publishNotReadyAddresses: true` is already set on all three headless services, and it matters because the join precedes `/ready`. |
| `compactor.enabled=false` (**non-default**) | **No.** Both compactor templates are wrapped in that condition. The CI check takes the expected role set as an argument, so this renders and passes with two roles. | N/A |
| `storage.mode=local` preset (**non-default**) | **Yes**, all three roles. Now covered by its own CI step. | Same helper, unaffected by storage mode. |
| `storage.mode=shared` preset (**non-default**) | **Yes**, all three roles. Own CI step. | As above. |
| `reader.replicas=0` (**non-default**) | **Unreachable.** The chart's JSON schema sets a minimum of 1 and rejects the render. | N/A |
| Existing install upgrading | **Yes.** The compactor and reader pod templates change, so those pods restart; writers are untouched. Quorum survives a single restart. The lease does not move; see above. | N/A |
| Raft bootstrap | **Unaffected.** `raft_bootstrap` is set only by `writerEntrypoint`, only on writer-0, so the compactor never stands up a rival single-node Raft. | `cluster.raft_bootstrap` defaults to false. |

## Review

One deep reviewer plus the matrix. No blockers, four High findings, all confirmed against the code before acting:

1. **My regression guard false-passed, two ways**, and the reviewer demonstrated both. Deleting the env var while leaving its name in the template's own explanatory comment satisfied a substring check. Changing `value: "compactor"` to the equally valid `value: compactor` made the role regex miss, silently dropping the role, and the script only failed when it had found no roles at all. Both attacks now fail, and I replayed them to confirm.
2. **Two release-note claims were false in the chart's default mode.** Manifest registration and the "No compactor elected" warning are both gated on peer replication, which shared storage does not use. The note now says so explicitly, because their absence is otherwise misread as evidence.
3. **The lease does not move after the upgrade.** The note claimed the opposite. Rewritten, and filed as #876.
4. **The wrong log line in three places.** The failure is in `Coordinator.Start()`, not the constructor, so an operator greps for "Failed to start" and not "Failed to initialize".

Also fixed: the helm action is pinned, since a manifest-parsing gate should not change behaviour with whatever shipped that morning; the script no longer hardcodes three roles; and I caught one more false claim in my own note during verification, that the cluster status endpoint reports the lease holder. It does not. `CompactorFailoverManager.Stats()` exists but nothing routes to it.

## Test plan

- [x] `helm lint` clean; the chart renders at default values and both presets.
- [x] The new check **fails on the pre-fix chart**, naming all three defects, and passes after.
- [x] Both of the reviewer's false-pass attacks replayed against the new check: both now fail. A third, removing the headless port, also fails.
- [x] `compactor.enabled=false` renders and passes with the two-role expectation.
- [x] Rendered compactor manifest inspected directly: the advertise address, the container port, and the matching `serviceName`.
- [ ] Not run on a live Kubernetes cluster. The check is a rendered-manifest assertion, and the runtime behaviour it guards is the same one the reader template already relies on in production.


xe-nvdk:
CI caught a bug in the guard itself, which is worth recording because the guard is the point of this PR.

`shift 2` fails outright when fewer than two arguments were given. Swallowing that failure left the arguments in place, so `"$@"` passed the chart path to helm a second time as a positional argument:

```
Error: expected at most two arguments, unexpected arguments: helm/arc-enterprise
```

CI calls the script with exactly one argument. I had run it locally with none and with four, so both arities I tested worked and the one that shipped did not. It now shifts one at a time, and I verified it at zero, one, two, three and four arguments, against the same helm 3.16.4 the workflow pins rather than the 4.x on my machine.

The Enterprise chart job is green on the fix.

Issue #875, opened 2026-09-16
ci: gofmt the eight dirty files on main and turn the gate on


xe-nvdk:
`ci.yml` carries this note:

> gofmt is deliberately NOT gated yet: a handful of pre-existing files (compaction/reconciliation/scheduler) are gofmt-dirty on main. Gate it once [they are clean].

Eight files are still dirty on `main` (verified at f3b5a1c0 with go1.26.6):

```
internal/compaction/scheduler.go
internal/compaction/watcher.go
internal/reconciliation/diff_test.go
internal/reconciliation/reconciler.go
internal/reconciliation/reconciler_test.go
internal/reconciliation/scheduler.go
internal/scheduler/cq_scheduler_test.go
internal/scheduler/retention_scheduler.go
```

Every diff is struct-field alignment, e.g. in `retention_scheduler.go` a field added to an aligned block was never re-aligned:

```diff
 	cron             *cron.Cron
-	running    bool
-	runningJob bool
-	mu         sync.Mutex
+	running          bool
+	runningJob       bool
+	mu               sync.Mutex
```

There is no semantic change in any of them.

Two reasons to close this out rather than leave the note standing. `.claude/CLAUDE.md` tells contributors `gofmt -l ./internal ./cmd` "must return empty", which is currently false, so the instruction trains people to ignore the output. And every review pass in a session reports these eight files as noise that has to be re-triaged as pre-existing each time.

Suggested: one commit running `gofmt -w` on exactly those eight files, then drop the note and add a gofmt step to `ci.yml`. Worth timing for a moment when nothing is in flight in `internal/compaction`, `internal/reconciliation` or `internal/scheduler`, since a whitespace-only change across a package conflicts with any open branch that touches it.

Issue #848, opened 2026-09-15
cluster: handleJoinRequest stores the joiner's role unvalidated; an unknown role parses as standalone, which can ingest


xe-nvdk:
Follow-up from the #807 review.

`handleJoinRequest` writes `req.Role` into the FSM node table as sent. `ParseRole` maps an empty or unknown string to `RoleStandalone`, whose capabilities include `CanIngest`. So a joiner that presents a bogus role (it already holds the cluster shared secret, so this is not an escalation across a trust boundary) ends up with a registry and FSM entry that passes the manifest-command role gate in `handleForwardApply`, both before and after #807, and shows up in listings with a role no operator configured.

Fix shape: validate the role in `handleJoinRequest` against the known roles (writer, reader, compactor) and reject the join otherwise; consider making `ParseRole` return an error or a dedicated invalid role instead of defaulting to standalone.

PR #879, opened 2026-09-16
feat(cluster): only nodes that can ingest vote in Raft elections (#862)

### LINKED ISSUE #862: cluster: a reader or compactor can win Raft leadership, which stops all singleton work in shared-storage mode
Found while answering how Pattern 2 is meant to fit together. Related to #850, but a distinct defect and, unlike #850, it affects the shared-storage pattern.

**The gate.** In shared-storage mode `Coordinator.IsPrimaryWriter()` is "I am the Raft leader AND my role is writer". The role half is deliberate and correct: a reader must never run retention, continuous queries or deletes against shared storage.

**The hole.** `handleJoinRequest` calls `AddVoter` for every node that joins, whatever its role, and the code comment at the gate acknowledges it ("every joining node becomes a Raft voter regardless of role"). So a reader or the compactor can be elected Raft leader. When that happens the leader fails the role half and every writer fails the leader half, so **no node in the cluster passes the gate**: retention, continuous queries and the non-dry-run retention, CQ and delete endpoints all stop until leadership happens to return to a writer. Nothing forces that to happen; there is no leadership transfer anywhere in the tree.

Compaction is unaffected, because it is gated on the compactor lease rather than on writer state.

**Fix: only nodes that take ingest should vote.** `hashicorp/raft` supports this directly through `AddNonvoter`, which we never use. A non-voter replicates the log and serves reads but never campaigns and cannot be elected. Join should dispatch on role: writers (and `standalone`, which ingests) join as voters; readers and compactors join as non-voters.

**Consequence to accept deliberately.** Quorum is then computed among writers only. Three writers survive losing one, which is the documented Pattern 2 shape. A cluster with a single writer makes that writer the only voter: it is always leader, but its loss takes Raft's quorum with it, so token and permission writes and cluster reconfiguration stop too, where today the readers would keep Raft alive. That is the correct trade — a cluster that cannot ingest should not be pretending to be healthy — but it means the documented Pattern 1 topology of one writer plus readers is not HA, which is the same conclusion #856 reaches from the promotion side. The guidance for both patterns should become three writer-role nodes.

**Work involved.**
- Wrap `AddNonvoter` on the raft node alongside `AddVoter`, and dispatch on role in `handleJoinRequest`.
- Refuse to bootstrap on a non-ingest role, or the first node of a cluster is an ineligible leader by construction.
- Decide the migration for existing clusters: a node already added as a voter stays one until it is removed and re-added, so either accept that this applies to new joins only, or demote on the next join.
- Tests: a reader joins as a non-voter and never appears in the voter configuration; a cluster of writers plus readers always elects a writer; a single-writer cluster reports the quorum consequence clearly rather than silently.
- Docs and the Helm chart: three writers, in both patterns.

xe-nvdk:
Closes #862.

On shared storage, retention, continuous queries and deletes run on the node that is both the Raft leader and a writer. The role half is deliberate: a reader must never run them against the shared bucket. But `handleJoinRequest` called `AddVoter` for every joining node regardless of role, so a reader or the compactor could be elected leader. When that happened the leader failed the role half and every writer failed the leader half, so **no node in the cluster passed the gate** and all of that work stopped, with nothing to force leadership back to a writer.

## Summary

- **Suffrage follows the role.** `NodeRole.VotesInElections` reads `CanIngest`, so writers and standalone nodes vote and readers and compactors join as non-voters. They replicate the log and see all cluster state; they just never campaign.
- **`CanIngest`, not `CanCoordinate`**, which sounds like the right field. `CanCoordinate` is false for `standalone`, which is the *default* `cluster.role`, so that predicate leaves a default-configured cluster with zero voters, which Raft rejects outright.
- **The predicate runs on a `ParseRole`-normalised role.** The capabilities table's default branch returns the zero value, so a raw unknown role would not vote, while `ParseRole` maps it to standalone, which does. They disagree on exactly the input a migration meets, and the safe answer is to keep the vote. Since #848 an unrecognised role cannot get this far anyway.
- **Bootstrap on a non-voting role is refused at startup.** That node would be the only server and an ineligible leader by construction, so nothing would ever elect. Placed after #848's validity guard, so a typo exits before this check can `ParseRole` it into standalone.
- **Three Raft wrappers**, for this and for the membership reconcile that follows. `DemoteVoter` refuses to demote this node *in the wrapper*, not by convention in the caller: `DefaultConfig` sets `ShutdownOnRemove` and a demotion trips the same `stepDown` path as a removal, so a leader that demotes itself shuts its own Raft down while Arc's wrapper keeps `running=true` and `n.raft` non-nil. A silent zombie only a process restart fixes. `AddNonvoter` carries the matching warning that it does **not** demote an existing voter.
- **Two accessors that raced `Stop()`** are fixed: `Node.Barrier` and `Node.ConfigurationServerIDs` read `n.raft` without the mutex, unlike every other accessor.

## Scope: this corrects new clusters

A node already recorded as a voter stays one until something changes it, so an existing cluster keeps its current voters until every node has re-joined on this version. **Converging existing membership is deliberately a separate change.** The adversarial validation of the plan found that an auto-promoting reconcile is the one path in that design which can leave a cluster permanently leaderless — adding a voter raises the quorum requirement, and if the promoted node is down the entry never commits, no further membership change is possible, and the sole leader steps down. Arc has no `RecoverCluster` or `peers.json` path.

Also worth stating plainly: `IsPrimaryWriter` only consults the Raft leader in shared-storage mode, so the headline symptom is shared-storage only. Local-storage clusters get the voter-set change without a behaviour fix.

## The trade, taken deliberately

Quorum becomes a property of the ingest-capable nodes. Three writers survive losing one, which is the documented shape for both patterns and what the charts have asked for since the writer default moved to three. A cluster with a single writer makes that writer the only voter: it is always leader, but losing it takes Raft's quorum with it, where previously the readers would have kept Raft alive. A cluster that cannot ingest should not report itself as healthy, and it is the same conclusion #856 reached from the promotion side.

## Prose moves with the code

Two PRs ago I corrected the charts and docs to say the writer count does **not** govern quorum, because every node was a voter. That is now wrong in the other direction, so this PR carries it: the chart README, `values.yaml`, `values-shared-storage.yaml`, and the validation message. The #856 two-writer warnings say it too.

`TestWriterRedundancy_MessageDoesNotClaimQuorumLoss` is **inverted rather than deleted**. It now asserts the two-writer messages *do* explain the quorum loss, and additionally that no role votes independently of whether it ingests — so a future change that re-enfranchises a non-ingesting role fails the test instead of quietly rotting the wording. Note it would otherwise have kept passing: nothing in CI would have prompted the prose update.

The README's quick-start blurbs also still claimed both presets deploy one writer. Both deploy three.

## Configuration matrix

|---|---|---|
| OSS standalone (`cluster.enabled=false`) | **No.** Both guards sit inside the cluster-enabled and clustering-licence blocks; no coordinator is built. | N/A |
| Cluster, default role (`standalone`), single node | **Yes**, and it votes. This is the row that rules out a `CanCoordinate` predicate: it would yield zero voters and Raft would reject the configuration. | Covered by a table test and an end-to-end join test. |
| Writer joins | **Yes**, as a Voter. Asserted against a real single-node Raft leader. | The joining node has a real Raft transport in the test, so `AddVoter`'s future actually commits. |
| Reader or compactor joins | **Yes**, as a Nonvoter. Same rig. | As above. |
| Join with an unrecognised role | **Unreachable.** #848 refuses it before this code. | Verified: the dispatch reads `node.Role`, already `ParseRole`d. |
| `raft_bootstrap=true` with a non-voting role | **Yes**, exits at startup. | Verified live for reader and compactor. |
| `raft_bootstrap=true` with a typo | **Yes**, exits through #848's guard first, before this one can normalise it. | Ordering verified live. |
| Existing cluster, mixed versions during upgrade | **Yes, and convergent rather than oscillating.** An old-code leader re-adds a demoted reader as a voter; a new-code leader would demote it again. No demotion ships in this PR, so the only effect is that nodes joining a new-code leader get correct suffrage. Upgrade writers first. | N/A |
| `cluster.raft_data_dir=""` (no Raft) | **No.** The dispatch is inside `if c.raftNode != nil`. | N/A |

## Test plan

- [x] `go test -race ./internal/cluster/ ./internal/cluster/raft/` green.
- [x] Revert-run, five behaviours: always `AddVoter` (the shipped bug) fails the reader and compactor tests; always `AddNonvoter` fails the writer and standalone tests; the predicate reading `CanCoordinate` fails the standalone tests; re-enfranchising readers fails the inverted quorum guard.
- [x] Join suffrage asserted against a **real single-node Raft leader**, reading `Suffrage` out of the live configuration — not against a nil-Raft harness, which is the gap the #848 review found in the existing join tests.
- [x] `gofmt -l` clean on the touched files (the eight it lists are pre-existing, #875), `go vet` clean, build clean.
- [x] **Live binary.** A reader joining a real two-node cluster is recorded `voter=false`, and the leader holds its term with no re-election. Bootstrap on reader and compactor exits 1; on writer it starts and becomes leader; a typo exits through #848's guard.


xe-nvdk:
Correcting a claim in the description before anyone acts on it.

I wrote that an existing cluster keeps its current voters "until every node has re-joined on this version". That is wrong, and I verified it in the hashicorp/raft v1.6.0 source rather than assuming:

```go
// configuration.go, nextConfiguration, case AddNonvoter
if server.ID == change.serverID {
    if server.Suffrage != Nonvoter {
        configuration.Servers[i].Address = change.serverAddress   // address only
    } else {
        configuration.Servers[i] = newServer
    }
```

Adding a server that is already a Voter as a non-voter updates its address and leaves `Suffrage` untouched. So a reader recorded as a voter before this change **stays a voter across restarts and re-joins**. Nothing in this PR demotes it; only `DemoteVoter` does, and nothing drives that yet.

The practical effect is unchanged for new clusters and strictly narrower for existing ones: they keep exactly the voters they have until the membership reconcile lands. The release note and a code comment now say that. It also means the prose asserting "writers are the Raft voters" is true of a cluster formed on this version, and not yet of one upgraded into it — worth deciding whether the chart README should carry that qualifier or wait for the reconcile.

xe-nvdk:
Review addressed. Six High findings, each verified before acting.

**The migration story was wrong twice, in opposite directions.** My first version said an existing cluster converges once every node re-joins. I corrected that to "it never converges by re-joining". Both were wrong. A graceful shutdown broadcasts a leave and the leader answers with `RemoveServer`, so the node leaves the configuration entirely and its next join gets correct suffrage — a rolling upgrade converges most of a cluster as a side effect. What stays behind is an ungraceful termination, a leave that arrived with no leader to process it, and the departing leader itself. Those keep their vote, because `AddNonvoter` on an existing voter only updates the address. There is now a test pinning exactly that, so the claim is checked rather than reasoned about, and the reconcile is filed as #880 and linked from the notes.

**The warning asserted something false on a cluster with standalone nodes.** They accept writes, so they vote, but `checkWriterRedundancy` counts only `RoleWriter`. A two-writer cluster with a standalone node has three voters and a real margin. The quorum sentence is now appended only when the ingest-capable count is actually thin, and that count is in the log fields.

**The join log said `voter=false` for a node that stayed a voter** — the opposite of the truth for exactly the operator auditing an upgraded cluster. It now reports the suffrage read back from the configuration.

**The bootstrap guard's justification was wrong.** The bootstrap configuration omits `Suffrage` and `ServerSuffrage`'s zero value is `Voter`, so a bootstrapping reader is elected. The harm is this issue's own harm, not a cluster that never elects. It is also gated on `raft_data_dir` now, and the fact that a previously harmless `raft_bootstrap=true` on a reader now exits has an upgrade note.

**Two comments still described the old behaviour**, including the one #862 quotes as the acknowledgement of the bug, and the reader StatefulSet's.

**The inverted guard test was true by construction** — it compared `VotesInElections` against its own definition and stayed green when readers were re-enfranchised through the capabilities table. It now checks both directions of the conditional sentence and names the roles. `DemoteVoter`'s self-refusal and `AddNonvoter`'s non-demotion had no tests; both now have behavioural ones against a real Raft node.

PR #882, opened 2026-09-16
chore(ci): gofmt the eight dirty files and turn the gate on (#875)

### LINKED ISSUE #875: ci: gofmt the eight dirty files on main and turn the gate on

xe-nvdk:
Closes #875.

`ci.yml` carried this note:

> gofmt is deliberately NOT gated yet: a handful of pre-existing files (compaction/reconciliation/scheduler) are gofmt-dirty on main. Gate it once that debt is cleared.

The debt was never cleared, while the project instructions told contributors `gofmt -l ./internal ./cmd` "must return empty". So the instruction was false, and the output was noise people learned to scroll past. Every review in a long session re-triaged the same eight files as pre-existing.

## The formatting change is semantics-free

Alignment, plus one blank comment line that gofmt adds when it reflows a doc-comment code block in `watcher.go`, plus one trailing blank line. Verified by stripping all whitespace from both sides of the diff and comparing what remained, which leaves only the added comment marker.

```
internal/compaction/scheduler.go           |  8 ++++----
internal/compaction/watcher.go             | 13 +++++++------
internal/reconciliation/diff_test.go       |  2 +-
internal/reconciliation/reconciler.go      | 14 +++++++-------
internal/reconciliation/reconciler_test.go | 16 ++++++++--------
internal/reconciliation/scheduler.go       | 20 ++++++++++----------
internal/scheduler/cq_scheduler_test.go    |  1 -
internal/scheduler/retention_scheduler.go  |  8 ++++----
```

## The gate

Added to the `test` job before the build, so a formatting-only failure reports in seconds rather than after the race suite, and it names the offending files rather than just exiting non-zero.

## Test plan

- [x] `go build ./...`, `go vet` on all three packages, and `go test -race ./internal/...` entire, all green.
- [x] The gate passes on this tree.
- [x] The gate **fails** on a deliberately misformatted probe file, naming it. A gate that cannot fail is not a gate.

## Timing

The issue asked for this to land when nothing was in flight in `internal/compaction`, `internal/reconciliation` or `internal/scheduler`, since a whitespace-only change across a package conflicts with any open branch that touches it. Nothing of mine is open in those packages now.


xe-nvdk:
Review came back clean, and it verified the formatting more rigorously than I had. It re-ran `gofmt -w` on the eight files as they stand on `main` and compared byte-for-byte against this branch: identical in all eight, so the diff is exactly gofmt's output with no hand-authored byte in it. It also compared token streams and position-stripped ASTs, and confirmed the `watcher.go` comment reflow is the Go 1.19+ doc-comment rule rather than an edit.

Took its one substantive suggestion: the gate now covers the whole repository instead of just `./internal ./cmd`. `pkg/` and `scripts/` are Go too and were unchecked. They are clean today, so covering them costs nothing now and closes the hole permanently, and it matches what `make fmt` already does.

The suggested fix in the failure message widened with it. A fix command narrower than the check would be this same trap in miniature: a contributor runs what the message says, sees it pass, and CI still fails.

Verified that a misformatted file in `pkg/models`, outside the old scope, now fails the gate.

Two things from the review worth recording for whoever merges this. Open branches need no rebase: the Format step runs on the PR merge ref, so they inherit this fix and go green on their own. And none of the other remote branches touches any of the eight files, so there is no textual conflict to manage.

PR #890, opened 2026-09-16
fix(compaction): append the implicit "time" sort key like ingest does

### LINKED ISSUE #792: fix(compaction): sort keys drop the implicit "time" column that ingest appends
## Summary

Ingest appends `"time"` to a measurement's sort keys; compaction does **not**. With any custom sort key configured, compaction re-sorts the data by the configured columns *only*, dropping the time ordering that ingest established within each group.

The comment directly above the compaction call site claims the opposite:

```go
// Build ORDER BY clause from sort keys
// This ensures compacted files maintain the same sort order as ingested files
orderByClause := buildOrderByClause(j.SortKeys)
```
`internal/compaction/job.go:744-746`

## Root cause

`config.ParseSortKeys` deliberately returns only the *additional* columns — `time` is documented as being appended "at the usage site":

```go
// Users only configure ADDITIONAL sort columns - the "time" column is always
// appended automatically at the usage site (getSortKeys in arrow_writer.go).
```
`internal/config/sort_keys.go:9-13`

There are two usage sites, and only one does the append.

**Ingest appends it** — `internal/ingest/arrow_writer.go:1003-1013`:
```go
// Always ensure "time" is the last sort key
for _, k := range keys {
    if k == "time" {
        return keys
    }
}
return append(keys, "time")
```

**Compaction does not** — `internal/compaction/manager.go:1194-1201`:
```go
func (m *Manager) GetSortKeys(measurement string) []string {
    if keys, exists := m.SortKeysConfig[measurement]; exists {
        return keys
    }
    return m.DefaultSortKeys
}
```

Both are fed the same un-appended parse output from `cmd/arc/main.go:1224` → `:1254-1255`.

## Repro

Config:
```toml
[ingest]
sort_keys = ["cpu:host"]
```

- Ingest sorts each flush by `(host, time)` — `arrow_writer.go:2600` → `sortTypedColumnBatchByKeys`.
- Compaction emits `ORDER BY "host"` — `buildOrderByClause`, `internal/compaction/job.go:83-99`, injected at `internal/compaction/dedup.go:158` / `:226`.

Rows within each `host` group land in whatever order the multi-file `read_parquet(..., union_by_name=true)` scan produces, which is not time-ordered across source files.

## Impact

- **Time-ordering within each sort group is lost after compaction.** Compacted files are the long-lived ones, so this is the state most data ends up in.
- **Row-group min/max stats on `time` widen**, so DuckDB skips fewer row groups on time-filtered queries — the exact pruning benefit sorting is supposed to buy. Arc writes statistics by default (`ingest.write_statistics`, `internal/config/config.go:1498`).
- Silent: no error, no warning, and only reachable when an operator sets a non-default `ingest.sort_keys` / `ingest.default_sort_keys`.

The default path (`default_sort_keys = "time"`, `config.go:1504`) is unaffected — `["time"]` is the same list either way. This is another instance of the "correct only at the default" class from CLAUDE.md.

## Suggested fix

Append `time` in `Manager.GetSortKeys` the same way ingest does, or hoist the append into `config.ParseSortKeys` and drop it from both usage sites (preferable — one authority instead of two that must agree).

Note `buildOrderByClause` emits plain `ORDER BY col` (ASC). If the fix lands, worth checking the interaction with #155 (default sort order DESC) and #153 (ORDER BY time DESC ~2x slower).

## Test plan

- [ ] Unit test asserting `Manager.GetSortKeys("cpu")` ends with `"time"` for config `"cpu:host"`
- [ ] Unit test asserting `buildOrderByClause` output for that config is `ORDER BY "host", "time"`
- [ ] Regression test must FAIL pre-fix (revert-run-restore)
- [ ] Integration: ingest with `sort_keys = ["cpu:host"]`, force hourly compaction, assert output is ordered by `(host, time)`

pujitha24:
Motivation:
Ingest sorts each flush by the configured sort keys with "time" appended
last (getSortKeys in internal/ingest/arrow_writer.go), so rows land in
(configured-keys..., time) order within a file. Compaction rebuilds the
long-lived files from many such inputs and is meant to preserve that
order -- the comment directly above the ORDER BY call site in
internal/compaction/job.go says so explicitly -- but
Manager.GetSortKeys in internal/compaction/manager.go returned the
configured keys unchanged, without appending "time". With any custom
ingest.sort_keys or ingest.default_sort_keys configured, compacted
files were re-sorted by the configured columns only, and rows within
each group landed in whatever order the multi-file
read_parquet(..., union_by_name=true) scan produced, not time order.

The default configuration (default_sort_keys = "time") was unaffected,
since ["time"] is the same list with or without the append -- this bug
only manifests once an operator sets a non-default sort key.

Impact is on query pruning, not correctness: with intra-group time
ordering lost, the "time" column's row-group min/max statistics widen,
so DuckDB skips fewer row groups on time-filtered queries against
compacted data. No rows are dropped, duplicated, or misordered across
files; compaction's own dedup ordering (ORDER BY "time" DESC) is a
separate mechanism and is unaffected.

Approach:
GetSortKeys now appends "time" when it is not already present, mirroring
the exact loop-then-append pattern arrow_writer.go's getSortKeys already
uses on the ingest side, so the two paths agree again.

Validation:
- go build ./... and go build -tags=duckdb_arrow ./... both succeed.
- go vet -tags=duckdb_arrow ./... is clean.
- gofmt -l on the changed files shows nothing (one pre-existing,
  unrelated gofmt issue in internal/iceberg/scheduler_test.go was not
  touched by this change).
- go test -tags=duckdb_arrow -race ./internal/... ./cmd/... passes in
  full.
- Added TestManager_GetSortKeys_AppendsTime and
  TestManager_GetSortKeys_TimeAlreadyPresent to
  internal/compaction/manager_test.go. Confirmed both new subtests of
  the former FAIL against the pre-fix code (git stash of manager.go
  alone) and PASS after the fix, which is the regression proof for
  this class of bug.

Report: https://github.com/Basekick-Labs/arc/issues/792

Fixes #792


xe-nvdk (approved):
Reviewed against #792. The compaction sort-key fix correctly mirrors ingest by appending the implicit `time` key, and the regression tests cover measurement-specific, default, and already-present cases. I resolved the release-note conflict with current `main`, preserved both entries, and corrected the contribution link to #890. The affected race-enabled suites pass. Approved.

PR #898, opened 2026-09-17
fix(compaction): prevent input filename collisions

### LINKED ISSUE #826: fix(compaction): daily tier flattens source keys by basename, so a collision duplicates one file's rows and silently drops another's
## Summary

Daily compaction flattens every source key into one temp directory by basename, with no collision check. Two source files that share a basename collide: one file's rows are duplicated into the output, the other's are dropped, and **both originals are marked compacted and deleted**.

Split out of #750 (item 1) so it can be milestoned and closed on its own. Same shape as the already-closed #357.

## The code

`internal/compaction/job.go:591`:

```go
localPath := filepath.Join(tempDir, filepath.Base(fileKey))

// Create local file for streaming
file, err := os.Create(localPath)
```

`os.Create` truncates an existing file. There is no check that `localPath` was not already written by another source key in the same job.

## Why basenames collide

The daily tier's candidate set spans all 24 hour subdirectories (`internal/compaction/daily.go:150-218`), so one job downloads files from 24 different prefixes into one flat temp dir.

Basenames are built from wall-clock only — `internal/ingest/arrow_writer.go:3753-3760`:

```go
now := time.Now().UTC()
timestamp := now.Format("20060102_150405")
nanos := now.UnixNano() % 1_000_000_000

return fmt.Sprintf("%s/%s/%s/%s/%s/%s/%s_%s_%09d.parquet",
    database, measurement, year, month, day, hour, measurement, timestamp, nanos)
```

The basename is `{measurement}_{YYYYMMDD_HHMMSS}_{nanos}.parquet`. It carries **no node ID, no PID, and no random component** — only the flush wall-clock. The partition hour that distinguishes the two keys is in the *directory* portion, which `filepath.Base` discards.

Note `nanos` is `UnixNano() % 1_000_000_000`, the sub-second remainder, so it repeats every second rather than being globally unique.

## Impact

Silent, and it destroys data: the dropped file's rows are gone, the duplicated file's rows are double-counted, and both source files are deleted as successfully compacted. Nothing errors.

## Suggested fix

Do not derive the temp filename from `filepath.Base`. The full key is already unambiguous — use a name derived from the whole `fileKey` (hash it, or mirror the key's directory structure under `tempDir`), or add an index to the local name.

Whichever is chosen, add a collision assertion: if the target local path already exists, fail the job rather than truncating.

## Test plan

- [ ] Unit test: two source keys differing only in the hour segment, asserting both land at distinct local paths
- [ ] Regression test must FAIL pre-fix (revert-run-restore)
- [ ] Integration: daily compaction over two hours holding same-basename files, asserting output row count equals the sum of inputs

Split from #750. Related: #357 (same shape, closed), #749 (temp-dir cleanup, same package).

efegokdemir:
## Summary

Fixes #826.

Daily compaction previously downloaded source files into a shared temporary directory using only their basenames. Different hour partitions could contain files with identical names, causing one download to overwrite another and potentially duplicate or lose rows.

This change:
- Prefixes temporary input filenames with their download index.
- Uses exclusive file creation to prevent overwriting an existing temporary file.
- Preserves the streaming download behavior.

## Regression tests

The basename-collision test failed against the original implementation and passed after the fix.

Additional coverage verifies:
- Different inputs with the same basename retain their individual contents.
- Existing temporary files are not overwritten.
- Real Parquet files from two different hours preserve both rows through daily compaction.

The integration test checks the row count, distinct hosts and value sum.

## Validation

Passed locally:
- `go test ./...`
- `go vet ./...`
- `go build ./...`
- Targeted compaction and integration tests
- Formatting and diff checks

## AI assistance

ChatGPT assisted with the implementation and regression tests.


efegokdemir:
The CI race-test job failed in TestReceiverKeepsItsMarkWhenWriterIsAhead (expected 100, got 674).

The failing test file is unchanged by this PR. I ran the test 50 times with the race detector on the base commit and 50 times on the PR commit; all 100 runs passed locally. This doesn't rule out an intermittent failure or establish its cause.

I tried to rerun the failed CI jobs, but GitHub requires repository admin rights. Could a maintainer please rerun the failed job when convenient? I'd prefer not to add an unrelated commit just to trigger CI.

xe-nvdk:
Thank you for the fix and the focused regression coverage, @efegokdemir! We completed a deeper review and confirmed the data-loss scenario: with 24 same-basename inputs, the original code produced 24 rows with only one distinct host and deleted all source files. Your fix preserved all 24 distinct hosts and the correct aggregate values.

The full race suite, vet, and build passed locally with current main's replication-test fix included. Additional lifecycle probes covered upload, source deletion, completion manifests, corrupt/missing inputs, read failures, and filename-length limits. No blocking code findings.

We also reproduced the unrelated flaky replication test on the base code and fixed it in #900. I merged main into this branch and resolved the release-note conflict while preserving both entries; fresh CI is now running. Thanks again for the contribution and for flagging the CI issue clearly!

PR #899, opened 2026-09-17
fix(cluster): enforce overall peer fetch timeout

### LINKED ISSUE #796: cluster: replication_fetch_timeout_ms is not honored past the fetch ack header; a stalled peer blocks a pull worker without bound
Found while reviewing the #759/#795 fix; pre-existing and independent of it.

`cluster.replication_fetch_timeout_ms` (default 60 s) is documented as the puller-side per-fetch overall timeout, and `Puller.pullOnce` does derive `fetchCtx` from it. But `FetchClient.Fetch` (`internal/cluster/filereplication/fetch_client.go`) only honors that deadline for the dial and the request send:

1. It sets `conn.SetDeadline(ctxDeadline)` right after dialing, then calls `protocol.ReceiveMessage(conn, f.ResponseHeaderTimeout)` for the ack header. `ReceiveMessage` (`internal/cluster/protocol/codec.go`) does `conn.SetReadDeadline(time.Now().Add(timeout))`, which **replaces** the ctx-derived deadline with the 15 s default, and `startFilePullerLocked` only sets `DialTimeout`, so `ResponseHeaderTimeout` is not tunable from cluster config. A peer that accepts and never answers costs 15 s per attempt regardless of `replication_fetch_timeout_ms`.
2. `ReceiveMessage` then runs `defer conn.SetReadDeadline(time.Time{})`, which **clears** the read deadline entirely, and the body transfer that follows is a bare `io.CopyN(mw, conn, ack.SizeBytes)` with no context and no deadline. A peer that acks and then stalls mid-body blocks the worker with no bound at all. With the default 4 workers, four such peers wedge the puller.

Live measurement while reproducing #795: three attempts against a paused peer took ~15 s in total (5 s handshake each), not the 60 s the config suggests; the stalled-body case is unbounded.

Fix shape: keep the ctx deadline authoritative (`ReceiveMessage` should not extend past it: use the earlier of the two, and not clear the deadline on return when the caller set one), and bound the body copy with the same ctx (a deadline-aware reader, or `SetReadDeadline` before `CopyN` and a goroutine that closes the conn on `ctx.Done()`). Expose `ResponseHeaderTimeout` from the coordinator or derive it from the fetch timeout.

efegokdemir:
## Summary

This PR fixes the file replication timeout handling described in #796.

The fetch client now keeps the per-fetch deadline effective across dialing, sending the request, reading the acknowledgement and streaming the file body. It restores the read deadline after the framed acknowledgement and closes the connection when the context is cancelled, so a stalled peer cannot keep the network read blocked.

The coordinator also derives the acknowledgement timeout from `cluster.replication_fetch_timeout_ms`, using the puller's default when the setting is omitted.

Regression tests cover stalled acknowledgement headers, partial bodies, overall deadlines and cancellation without a deadline.

The shutdown regression test has been updated for prompt fetch cancellation. It now synchronizes a file deletion after shutdown begins but before Raft stops. This preserves a shutdown/deletion overlap, although it no longer specifically tests deletion during the puller join.

## Testing

- Fetch timeout and cancellation regressions: 20 repeated runs with the race detector.
- Shutdown regression: 10 repeated runs with the race detector.
- `go vet ./...`
- `go build ./...`
- Formatting and `git diff --check` passed.

Fixes #796


xe-nvdk:
Thank you, @efegokdemir, for this fix and the regression tests! Merged after all CI checks passed.

The deeper review confirmed that stalled acknowledgement and body reads now respect the fetch budget, cancellation unblocks established connections, and an interrupted transfer stays staged and resumes correctly. The full race suite, vet, and build also passed locally against current main.

We tracked the separate, pre-existing cancellation gap during TCP/TLS connection setup in #901 so it can be addressed without holding up this improvement. Thanks again for both this contribution and the compaction data-loss fix in #898!

PR #925, opened 2026-09-19
fix(query): reject backslashes during tier path extraction (#750)

efegokdemir:
## Summary

Addresses **item 5 only** of #750.

`extractDBMeasurementFromPath` previously converted backslashes into
forward slashes. That disagreed with the storage key contract, which
rejects backslashes, and could cause a malformed path to be interpreted
as a different database/measurement during tier selection.

The extractor now refuses to derive a database or measurement from
paths containing backslashes. Valid forward-slash paths retain their
existing behaviour.

This PR also adds a regression test and release-note entry.

## RED → GREEN evidence

Before the fix, the regression failed for Azure, S3 and relative paths.
Each malformed path was incorrectly interpreted as
`database="private", measurement="cpu"`.

After the fix, the regression passed.

## Validation

Passed locally:

- Focused regression using `duckdb_arrow`.
- API and storage package test suites.
- API and storage race suites.
- Full build with `duckdb_arrow`.
- Relevant `go vet`.
- Formatting and `git diff --check`.

Only three intended files changed.

## Limits

This verifies the path-extraction defect and relevant package
regressions. A live Azure/tiering integration test was not performed.

This is a Draft PR pending review of that remaining validation.

Other findings in #750 are outside this PR's scope.

ChatGPT assisted with the implementation and regression test.
Validation was executed locally.


efegokdemir:
Closing as a duplicate of #918. That earlier PR covers the same #750 item 5 and additionally preserves valid native separators in a trusted local storage root. No branch is being deleted.

